PR #87, opened 2026-04-08
Codex permission reply: async xdotool flow + Linux-only actionable bubble

aaronWool:
This PR contains only the permission-reply track (separate from detection-side changes
  already being merged to main).

  ## What changed

  1. Async permission reply execution

  - Replaced blocking execFileSync with async execFile for Codex reply actions.
  - Prevents Electron UI stalls when xdotool hangs or is slow.

  2. Removed /proc/<pid>/fd/0 write path

  - Dropped cross-process stdin write fallback entirely.
  - Avoids PID-recycling / TOCTOU risk.
  - Fallback behavior is now safe: focus terminal only.

  3. Platform-gated Codex bubble actions

  - Linux: shows Allow, Deny, Go to Terminal.
  - macOS/Windows: Codex bubble is dismiss-only (Got it).
  - process.platform is passed to bubble renderer through permission-show IPC payload.

  4. Graceful missing-xdotool handling

  - Handles ENOENT in async callback.
  - Silently falls back to focusTerminalForSession(...) (no crash, no blocking).

  5. Main-flow alignment

  - Keeps permission state as notification.
  - Keeps Codex notify cleanup on any non-permission event (as requested).

  ## Files

  - src/permission.js
  - src/bubble.html
  - src/main.js

  ## Validation

  - node --check src/main.js
  - node --check src/permission.js
  - node --test test/codex-log-monitor.test.js

  ## Notes

  - This PR intentionally does not include detection-side/caching/event-map changes.
  - Behavior on Linux when auto-reply fails remains safe: user is focused to terminal to answer
    directly.

### DIFF
--- a/agents/codex-log-monitor.js
+++ b/agents/codex-log-monitor.js
@@ -26,6 +26,7 @@ class CodexLogMonitor {
     this._recentDayDirsCache = [];
     this._recentDayDirsCacheAt = 0;
     this._recentDayDirsDateKey = "";
+    this._startedAtMs = Date.now();
   }
 
   _resolveBaseDir() {
@@ -38,6 +39,7 @@ class CodexLogMonitor {
 
   start() {
     if (this._interval) return;
+    this._startedAtMs = Date.now();
     // Initial scan
     this._poll();
     this._interval = setInterval(
@@ -247,6 +249,14 @@ class CodexLogMonitor {
       obj = JSON.parse(line);
     } catch {
       return; // corrupted line, skip
+    }
+
+    // Skip historical timestamped events that predate monitor start.
+    // This prevents replay storms on app restart from driving stale state
+    // transitions (e.g., old permission -> old clear) in live UI.
+    if (obj && typeof obj.timestamp === "string") {
+      const ts = Date.parse(obj.timestamp);
+      if (Number.isFinite(ts) && ts < this._startedAtMs - 1500) return;
     }
 
     const type = obj.type;
--- a/agents/codex-log-monitor.js
+++ b/agents/codex-log-monitor.js
@@ -295,10 +295,11 @@ class CodexLogMonitor {
       tracked.hadToolUse = false;
       tracked.lastState = resolved;
       tracked.lastEventTime = Date.now();
+      const agentPid = this._resolveTrackedAgentPid(tracked);
       this._onStateChange(tracked.sessionId, resolved, key, {
         cwd: tracked.cwd,
-        sourcePid: null,
-        agentPid: null,
+        sourcePid: agentPid,
+        agentPid,
       });
       return;
     }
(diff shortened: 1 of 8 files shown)


rullerzhou-afk (commented):
Hey @aaronWool, thanks for the PR! Solid error handling and platform gating throughout — appreciate the care.

A few things to flag before we move forward:

### 1. Overlap with main

Most of the changes in `codex-log-monitor.js`, `launch.js`, `test/codex-log-monitor.test.js`, and `test/registry.test.js` are already on `main` (day-dir caching, timestamp filtering, PID tracking via `/proc`, `exec_command` support, Linux sandbox flags, etc.). The branch would need a rebase — expect conflicts in ~5 of the 8 files. The genuinely new parts are the xdotool reply flow, the platform-aware bubble UI, and the `agent_message` mapping change.

### 2. xdotool PID mismatch (core issue)

`replyCodexPermission()` calls `xdotool search --pid <codexPid>`, where `codexPid` comes from `_findCodexWriterPid()` — this finds the **Codex CLI process** (the Rust binary that has the JSONL file open). But `xdotool search --pid` looks for X11 windows whose `_NET_WM_PID` matches. Codex is a CLI app running *inside* a terminal emulator; it doesn't own any X11 windows. The terminal emulator (gnome-terminal / kitty / alacritty) owns the window under its own PID.

```
terminal-emulator (PID 1000)  ← owns the X11 window
  └─ bash (PID 1001)
       └─ codex (PID 1002)    ← _findCodexWriterPid returns this
```

`xdotool search --pid 1002` → 0 results → falls through to `focusTerminalForSession()`, which also uses the same Codex PID on Linux (`wmctrl -lp` / `xdotool search --pid`) → also 0 results. So Allow/Deny on Linux would silently no-op in practice.

To fix this, `_findCodexWriterPid` (or a wrapper) would need to walk *up* the process tree from the Codex PID to find the terminal emulator — similar to how `createPidResolver()` in `hooks/shared-process.js` walks from `process.ppid` through parent PIDs to find the terminal. On Linux you can read `/proc/<pid>/stat` to get the ppid and climb the chain.

### 3. `agent_message: null → "working"`

The current `null` mapping was intentional — `agent_message` is Codex's text output (reasoning/explanations), not tool execution. Changing it to `"working"` would make the pet show the typing/building animation during plain text responses, losing the thinking-vs-working distinction.

The PR comment mentions "newer Codex builds emit work progress primarily as event_msg records" — could you share a sample JSONL snippet showing this? If newer builds genuinely use `agent_message` for tool progress rather than text output, we can adjust. Otherwise I'd lean toward keeping `null` (or mapping to `"thinking"` if we want some visual feedback during text generation).

### What I'd like to keep

The non-xdotool parts are well-structured and I'd happily take them:
- `platform` injection into `permission-show` payload
- `codexPid` plumbing + validation (`Number.isFinite && > 1`)
- `handleDecide` action routing framework for Codex notify bubbles
- Platform-aware bubble UI in `bubble.html` (Allow/Deny/Go to Terminal on Linux, Got it elsewhere)
- `permLog` additions in `showCodexNotifyBubble` / `clearCodexNotifyBubbles`

### Suggested path forward

1. Rebase onto current `main` to drop the already-merged parts
2. Fix the PID resolution — walk from Codex PID up to terminal PID before calling xdotool
3. Clarify the `agent_message` change with evidence, or drop it for now

Happy to help with the ppid-walk implementation if you'd like to discuss the approach. Thanks again for the contribution!

PR #106, opened 2026-04-14
refactor: able to run on daemon

plumLee211:
## Summary

  This PR adds a detached development launcher so Clawd can keep running without holding the current terminal open.

## Changes

  - add npm run start:daemon for background startup in development mode
  - add scripts/start-detached.js to launch Electron as a detached GUI process
  - update README.md and README.zh-CN.md to document foreground vs background startup
  - document the recommended packaged .app flow for users who do not want to launch from a terminal

## Notes

  - this is intended for local development from a cloned repo
  - npm start remains the foreground/dev-friendly option
  - packaged app usage and Open at Login remain the better long-term path for end users

+) I don't speak Chinese. So I used AI translation. 
     If anything is mistranslated, please let me know.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,6 +55,7 @@ scripts/*
 !scripts/run-clawdbot-stress.sh
 !scripts/monitor-clawd-resources.sh
 !scripts/validate-theme.js
+!scripts/start-detached.js
 
 # oh-my-cursor state
 .omc/
--- a/README.md
+++ b/README.md
@@ -97,9 +97,16 @@ cd clawd-on-desk
 # Install dependencies
 npm install
 
-# Start Clawd (auto-registers Claude Code hooks on launch)
+# Start Clawd in foreground (auto-registers Claude Code hooks on launch)
 npm start
+
+# Start Clawd in background and release the terminal
+npm run start:daemon
 ```
+
+If you do not want a terminal window involved at all, build the macOS app with
+`npm run build:mac` and launch the generated `.app` normally. The tray menu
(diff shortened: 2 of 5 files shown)


rullerzhou-afk:
Hey, thanks for putting this together and for the clear writeup.

I'm going to pass on this one, though — a few reasons:

1. **Shell backgrounding already covers this.** `npm start &` on macOS/Linux or `start /B npm start` on Windows gets you the same "terminal-free" behavior without any new code.

2. **The better path for background use already exists.** As you noted in your own PR description, packaged `.app` / `.exe` builds + the tray's "Open at Login" are the recommended flow for anyone who wants Clawd running persistently. I'd rather steer users there than add a third startup path.

3. **README churn risks confusing new users.** Putting `start:daemon` in Quick Start makes people stop and wonder which one they should pick, when really 99% of them just want `npm start` (devs) or the packaged app (everyone else).

4. **Maintenance surface.** `launch.js` and `start-detached.js` would duplicate env-cleanup / Linux sandbox logic, and I'd have to remember to keep both in sync.

No hard feelings — it's a reasonable itch to scratch, just not one I want to carry in the main repo. Feel free to keep it in your fork if it's useful for your own workflow.

Thanks again for the contribution and the thoughtful notes.

plumLee211:
Thanks for the explanation and for reviewing this PR

I opened the PR because I wanted to avoid having the terminal occupied while running Clawd.
But I didn’t realize there was already a simple way to do that with the existing functionality and without changing the codebase.

Thanks for pointing that out and I appreciate you taking the time to review it.

smile_cat

PR #118, opened 2026-04-16
fix: restore window position when disconnected display is reconnected

wbluke:
## Problem

When an external monitor is disconnected, the pet window is clamped to a surviving display (expected). However, when the monitor is reconnected, the pet stays on the primary display instead of returning to its original position.

This is especially noticeable with multi-monitor setups (e.g., primary + 2 external monitors) — disconnecting both and reconnecting may place the pet on the wrong monitor.

## Root cause

- `display-removed` handler saves position *after* clamping, so if two displays are removed sequentially, the second event overwrites the backup with the already-clamped (primary) position.
- `display-added` handler had no position restoration logic at all.

## Fix

- **`display-removed`**: stash the pre-clamp bounds only when `clampToScreen()` actually moves the window. This naturally prevents overwriting a valid backup when a second, unrelated display is removed.
- **`display-added`**: check whether the stashed position's center point falls within any currently available display's work area. If so, restore the pet there and persist the position.

## Testing

Tested manually on macOS with primary + 2 external monitors:
1. Place pet on external monitor 2
2. Disconnect both external monitors → pet moves to primary ✓
3. Reconnect both monitors → pet returns to original position on monitor 2 ✓

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -193,6 +193,12 @@ function flushRuntimeStateToPrefs() {
   });
 }
 
+// ── Display-disconnect position backup ──
+// When a display is removed the window gets clamped to a surviving monitor.
+// We stash the pre-clamp position so that when the same display comes back
+// (display-added) we can restore the pet to where it was.
+let _preDisconnectBounds = null;   // { x, y, width, height } or null
+
 let _codexMonitor = null;          // Codex CLI JSONL log polling instance
 let _geminiMonitor = null;         // Gemini CLI session JSON polling instance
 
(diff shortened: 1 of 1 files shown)


rullerzhou-afk (changes requested):
The overall direction makes sense, but I do not think the current state handling is safe enough to merge yet.

Main issues:

1. `_preDisconnectBounds` still behaves like a rolling backup, so cascading disconnects are not actually fixed.
If the pet starts on display C, then C is removed and the window clamps onto B, and then B is also removed and the window clamps onto A, the second `display-removed` can still overwrite the original backup with the already-clamped B position. That means reconnecting will restore to B instead of the original C position.

2. There is no invalidation path when the user explicitly repositions the pet after a disconnect.
If the user drags the pet somewhere else after the clamp, `display-added` can still snap it back to the stale pre-disconnect location. That would override an intentional user action.

3. Restoring raw `{ x, y, width, height }` is risky in proportional size mode.
After a disconnect, the saved width/height may already reflect the temporary surviving display rather than the display we actually want to restore to. On reconnect, the window should be resized for the target display before restoring position.

Suggested approach:

1. Treat `_preDisconnectBounds` as a single pending restore target, not a rolling backup.
Only save it on the first `display-removed` event that actually moves the window, and do not overwrite it on later removals.

2. Invalidate the pending restore target on explicit user-driven repositioning.
At minimum, dragging the pet or using a manual reposition action such as `sendToDisplay` should clear it.

3. Do not restore raw saved bounds.
Instead, save the original logical target position, recompute the correct size for the display being restored to, and then run the normal clamp path with that size.

4. Prefer reusing the existing clamp semantics for restore instead of a separate center-point-in-workArea check.
That will be more consistent with the rest of the window placement logic and should avoid edge cases around partially off-screen but still valid positions.

I have only reviewed this statically and have not been able to hardware-test it on a multi-monitor setup. Please verify the final behavior on a real multi-monitor configuration before merge, especially for:
- cascading disconnects (`C -> B -> A`)
- disconnect, then user drags the pet somewhere else, then reconnect
- proportional size mode across displays with different resolutions/scales

PR #130, opened 2026-04-18
Fix Windows Settings window taskbar icon

YOIMIYA66:
## Summary

Fix the Windows Settings window taskbar icon so it uses a project icon instead of inheriting the default Electron icon.

## Problem

After the new Settings panel was added, the Settings window title bar used the Clawd icon, but on Windows the taskbar thumbnail / grouped taskbar icon could still show the default Electron icon.

This was reproduced in source-mode on Windows with the Settings window open from the desktop pet.

## Root cause

The Settings window icon path logic was still centered around `assets/icon.ico`, and the app was not explicitly setting a Windows App User Model ID for the process.

That combination is enough for the window chrome to look correct while the taskbar grouping identity still falls back to Electron defaults in some Windows cases.

## What changed

- Added `src/settings-window-icon.js` to centralize Windows Settings-window icon selection and AppUserModelID setup.
- On Windows, the Settings window now prefers the project `assets/icons/256x256.png` icon, with `assets/icon.ico` as a fallback.
- Applied the project Windows App User Model ID at startup so the Settings window is grouped under the app identity instead of Electron's default identity.
- Updated packaging config so `assets/icons/**/*` is included in packaged builds.
- Added regression tests for the icon selection logic and packaging config.

## Files changed

- `package.json`
- `src/main.js`
- `src/settings-window-icon.js`
- `test/package-build-config.test.js`
- `test/settings-window-icon.test.js`

## Validation

Automated:
- `node --check src/main.js`
- `node --test test/settings-window-icon.test.js test/package-build-config.test.js`
- `npm test`

Latest full run:
- `781` passing
- `0` failing
- `1` skipped

Manual:
- Verified on Windows that the Settings window taskbar entry now uses the project icon rather than the default Electron icon.
- The reported Windows scenario was reproduced from the new Settings panel UI and rechecked after the fix.

## Platform status

- Windows: manually validated
- macOS: not manually tested for this change
- Linux: not manually tested for this change

## Notes

- This PR is intentionally narrow: it only changes the Settings window icon identity path on Windows and the packaging needed to ship the selected icon asset.
- I did not attach the local screenshot file to the PR body, but the Windows validation above was performed against that exact reported taskbar case.

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,6 +86,7 @@
     },
     "files": [
       "src/**/*",
+      "assets/icons/**/*",
       "assets/icons/agents/**/*",
       "assets/svg/**/*",
       "assets/sounds/**/*",
--- a/src/main.js
+++ b/src/main.js
@@ -3,6 +3,10 @@ const path = require("path");
 const fs = require("fs");
 const { pathToFileURL } = require("url");
 const { applyStationaryCollectionBehavior } = require("./mac-window");
+const {
+  applyWindowsAppUserModelId,
+  getSettingsWindowIconPath,
+} = require("./settings-window-icon");
 const hitGeometry = require("./hit-geometry");
 const animationCycle = require("./animation-cycle");
 const { findNearestWorkArea, computeLooseClamp, SYNTHETIC_WORK_AREA } = require("./work-area");
@@ -28,6 +32,8 @@ const isMac = process.platform === "darwin";
 const isLinux = process.platform === "linux";
 const isWin = process.platform === "win32";
 const LINUX_WINDOW_TYPE = "toolbar";
+
+applyWindowsAppUserModelId(app, process.platform);
 
 
 // ── Windows: AllowSetForegroundWindow via FFI ──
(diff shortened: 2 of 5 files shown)


YOIMIYA66:
0fdfcbf57342a10fd384735002b85f11

rullerzhou-afk:
Thanks for the fix, @YOIMIYA66! 🦀

Pulled it locally and verified on Windows:
- ✅ All 781 unit tests pass (including the new `settings-window-icon` + `package-build-config` suites)
- ✅ Settings window taskbar icon now shows the Clawd crab — was the default Electron atom before
- ✅ Taskbar hover reads "Clawd Settings" and groups under "Clawd on Desk"

Really appreciate the clean design — the pure-function split with injected `existsSync` makes the logic trivially testable, and setting `AppUserModelID` is the right root cause rather than a bandaid. Nicely scoped PR with tests, exactly the kind of contribution that's a joy to review.

Merged into `main` 🎉 You'll be credited in the next release notes and the README contributors grid. Thanks again!

PR #131, opened 2026-04-19
[codex] Fix Windows Settings taskbar shell icon

YOIMIYA66:
## Summary

Fix the Windows Settings window taskbar button still showing the default Electron shell identity in source runs.

## Root cause

The Settings window already had a custom `BrowserWindow` icon and an application-wide `AppUserModelID`, but the Windows taskbar right-click menu and shell surfaces still fell back to the underlying `electron.exe` identity in unpackaged runs.

## What changed

- add a Windows-specific helper that builds taskbar-shell metadata for the Settings window
- set `BrowserWindow.setAppDetails(...)` for the Settings window on Windows
- use the project `.ico` asset for shell-facing surfaces while keeping the existing window icon selection logic
- add regression tests for the shell icon path and unpackaged relaunch metadata

## Validation

- `node --check src/main.js`
- `node --test test/settings-window-icon.test.js`
- `npm test`

## Platform notes

- Windows: manually reproduced and verified against the taskbar right-click menu case shown by the reporter
- macOS: not manually tested for this change
- Linux: not manually tested for this change

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -6,6 +6,8 @@ const { applyStationaryCollectionBehavior } = require("./mac-window");
 const {
   applyWindowsAppUserModelId,
   getSettingsWindowIconPath,
+  getSettingsWindowTaskbarDetails: getSettingsWindowTaskbarDetailsHelper,
+  SETTINGS_WINDOW_TITLE,
 } = require("./settings-window-icon");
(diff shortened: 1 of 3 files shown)


YOIMIYA66:
download

PR #148, opened 2026-04-22
feat(settings): add per-theme sound effect overrides

Rladmsrl:
## Summary

Lets users replace a theme's sound effects from the settings panel, alongside the existing animation overrides. Addresses the long-standing UX gap where sounds could only be replaced by editing `theme.json` by hand.

- Files are picked from **any path** via a system file dialog and **copied** into a per-theme folder `userData/sound-overrides/{themeId}/{soundName}.{ext}`, so the override survives even if the source file is later moved or deleted.
- Everything rides the existing `themeOverrides` prefs schema, so sound overrides are automatically preserved across theme reloads and carried through the existing **export / import** channel — no new persistence surface.
- The overrides tab is renamed to **"Animation & Sound Overrides"** / **"动画 / 音效替换"** / **"애니메이션 · 음향 덮어쓰기"** so the scope is no longer ambiguous.

## What's shown

Under the overrides tab, a new **Sound effects** section lists every sound name the current theme declares (default `complete` / `confirm` + any custom names the theme author added). Each row has:

- **Play** — preview the currently-active file (theme default or override)
- **Choose file…** — system file picker, mp3 / wav / ogg / m4a / aac / flac
- **Reset to default** — clear the override (enabled only when overridden)

Plus an **"Open overrides folder"** button as an escape hatch for manual cleanup.

## Design notes

- **Why a new `sound-overrides/` folder instead of writing into the theme's own `sounds/` dir**: built-in themes live inside `app.asar` and aren't writable; external theme dirs are writable but polluting them would confuse the "export theme" and "remove theme" flows. A sibling `userData/sound-overrides/{themeId}/` keeps override data cleanly separated from the theme itself.
- **File naming `{soundName}.{ext}`**: each sound slot has at most one override on disk; changing the extension (e.g. mp3 → wav) deletes the sibling with the old extension so the folder stays tidy.
- **Preview cache-busting**: the renderer caches `new Audio(url)` by URL to avoid reloading on every play. A preview without reload would play the cached stale Audio object — so `settings:preview-sound` appends `?_t=<timestamp>` to force a fresh load. Regular playback doesn't need this because `setSoundOverride` triggers `activateTheme`, which reloads the renderer and clears the cache naturally.
- **Sound override entries only store `{ file }`** — no transition/duration/disabled, unlike animation overrides. The `normalizeSlotOverride` helper is reused for basename safety, but ancillary fields are stripped.
- **Scope intentionally narrowed**: the UI only lists sound names the theme actually declares under `theme.sounds`. Overriding a name `state.js` never triggers would produce a silent, never-played override — declaring new sound names is a theme-author concern, not a user setting.

## Files touched

- `src/prefs.js` — `normalizeSoundOverridesMap` + `themeOverrides[themeId].sounds` branch
- `src/theme-loader.js` — `getSoundOverridesDir`, `_resolveSoundOverrideFiles`, `getSoundUrl` prefers override paths
- `src/settings-actions.js` — `setSoundOverride` command; `cloneSoundOverrides` threaded through every other themeOverrides mutation so edits to one kind don't drop the others
- `src/main.js` — three new IPC handlers (`pick-sound-file`, `preview-sound`, `open-sound-overrides-dir`); `_buildSoundOverrideSlots` feeds the UI under `data.sounds`
- `src/preload-settings.js` — expose the three new APIs
- `src/settings-tab-anim-overrides.js` — new Sound effects section renderer
- `src/settings.html` — minimal CSS for the sound row layout
- `src/settings-i18n.js` — new strings for the sound section; tab label rename in en / zh / ko

## Test plan

Automated (new):
- [x] `test/prefs.test.js` — sounds map round-trips through save/load; invalid entries dropped; ancillary fields stripped (duration / transition / sourceThemeId)
- [x] `test/settings-actions.test.js` — `setSoundOverride` accepts/rejects payloads, clears on `file: null`, preserves other override subtrees, triggers `activateTheme` only for the active theme, errors when dep is missing

Automated (regression):
- [x] `npm test` — relevant suites (prefs, theme-override, settings-actions, theme-loader, i18n) all green. Pre-existing unrelated failures on macOS (Windows-only Claude CLI discovery tests, calico SVG layout snapshots, updater download harness) are untouched — same 7 failures on `main`.

Manual (macOS):
- [x] Pick → copy → preview → reset → open folder, both `complete` and `confirm`
- [x] Extension change (mp3 → wav) removes the old sibling file on disk
- [x] Reset clears the prefs entry and the active theme falls back to the default sound audibly
- [x] Tab label rename visible in zh / en switching

## Out of scope

- Doesn't touch `state.js`'s hardcoded `playSound("complete" | "confirm")` trigger points. Theme authors can declare additional sound names but only `complete` and `confirm` actually fire today.
- The global "Sound effects on/off" mute toggle in General tab is unchanged.


rullerzhou-afk:
Thanks for the thorough work here — the architecture (riding entirely on `themeOverrides`, keeping export/import and legacy migration working) is exactly right, and the test coverage on `setSoundOverride` and `normalizeSoundOverridesMap` is solid. A few issues surfaced on a second read that I'd like sorted before merge. Happy to take any of these myself (repo has maintainer-edit enabled) — just let me know which you'd prefer to own.

## Blockers

### 1. `soundName` is used as a path segment without sanitization (high)

In `src/main.js` (`settings:pick-sound-file`):

```js
const destFilename = `${soundName}${ext}`;
const destPath = path.join(overridesDir, destFilename);
fs.copyFileSync(sourcePath, destPath);
```

`soundName` is validated as "declared by the current theme", but theme authors control `theme.sounds` keys. A third-party theme can ship `sounds: { "../../foo": "x.mp3" }`, and when the user picks a replacement the copy lands outside `sound-overrides/<themeId>/`. The extension whitelist blocks writing an executable, but arbitrary audio files in arbitrary paths is still a real write primitive from an untrusted theme.

The same concern applies to `_cleanupSiblingSoundOverrides` (it uses `path.parse(entry).name` vs `soundName` so it's safer, but the underlying trust assumption is the same) and to `normalizeSoundOverridesMap` in `prefs.js` — the PR description mentions `normalizeSlotOverride` gives "string-basename validation for free", but it actually only checks `typeof entry.file === "string" && entry.file`; no basename enforcement.

Suggested fix: validate `soundName` against `/^[a-zA-Z0-9_-]+$/` (or at least assert `path.basename(soundName) === soundName` and no `..`) in the IPC handler, and `_basenameOnly(entry.file)` inside `normalizeSoundOverridesMap` as defence in depth.

### 2. Same-extension replacement leaves `_audioCache` stuck on the old audio (medium)

The PR description says:
> Regular playback doesn't need this because `setSoundOverride` triggers `activateTheme`, which reloads the renderer and clears the cache naturally.

`activateTheme` fires a `theme-changed` event but doesn't reload the renderer window. `_audioCache` in `renderer.js:782` is a module-level `const` and survives theme switches. When a user replaces `complete.mp3` with another `.mp3`:

- On-disk file is overwritten at the same path
- `file:///.../complete.mp3` URL is unchanged
- `_audioCache[url]` still holds the old `Audio` object with the old buffered data
- `playSound("complete")` plays the old audio on every future trigger until the app restarts

Preview works correctly because `?_t=<timestamp>` produces a new URL, but that's exactly why the real-playback path silently breaks.

Suggested fix: after `pick-sound-file` succeeds, send an IPC to the renderer to `delete _audioCache[url]` for the affected override URL (or broadcast a cache-clear on `theme-changed` / override commits). A regression test that exercises `playSound` twice across a same-ext replacement would have caught this — worth adding alongside the fix.

### 3. `settings:preview-sound` bypasses DND (medium)

`playSound()` in `main.js:585` guards with `if (soundMuted || doNotDisturb) return;`. The new `settings:preview-sound` handler sends `play-sound` directly, bypassing both. A user in DND (meeting / focus mode) clicking "Play" in Settings will still hear the sound through the system.

Suggested fix: short-circuit on `doNotDisturb` in the handler (returning `{ status: "skipped" }` is fine — the UI can silently drop it or show a small hint). `soundMuted` is a judgement call — I'd argue preview should still honor it, since the user could be in a shared space, but happy to leave that open.

## Low (optional, can land in a follow-up)

### 4. Preview grows `_audioCache` unboundedly

Every preview click produces a unique URL (timestamp), and every unique URL gets a permanent entry in `_audioCache`. A user previewing 50 times leaves 50 `Audio` objects in memory for the renderer's lifetime. Not severe, but the fix is small: either `delete _audioCache[url]` on the `ended` event for preview URLs, or don't cache URLs containing `_t=` at all.

---

Happy to push any of these to your branch directly if you'd rather focus on something else — just say the word. Again, really nice PR overall: the design call to put overrides under `userData/sound-overrides/` instead of polluting the theme's own `sounds/` dir, and the sibling-extension cleanup, are both the right instincts.

Rladmsrl:
Thanks for the careful second read — all four addressed in 5bcb29b. Walking through each vs. your suggestion:

### 1. `soundName` sanitization (high)

Added `/^[a-zA-Z0-9_-]+$/` validation in `settings:pick-sound-file` (`src/main.js`) so a malicious theme's `sounds: { "../../foo": "x.mp3" }` key is rejected before any path join. As defence in depth, `normalizeSoundOverridesMap` in `src/prefs.js` now enforces the same regex on keys and basename-sanitises `file` (rejecting `.` / `..` too) — so a hand-edited pref file or bad import can't slip past either.

### 2. Stale `_audioCache` on same-ext replacement (medium)

Took your first option — specific URL invalidation over a blanket broadcast. After a successful pick, main computes `themeLoader.getSoundUrl(soundName)` and sends a new `invalidate-sound-cache` IPC; renderer does `delete _audioCache[url]`. Covers all three paths:

- New override / different filename → `setSoundOverride` commits → `activateTheme` reloads renderer → cache already empty, invalidate is a no-op.
- **Same-ext replacement** → `setSoundOverride` noops (pref bytes identical) → no reload → invalidate deletes the stale entry so next `playSound` creates a fresh `Audio`. This is the path your write-up caught.

**Regression test not added** — `_audioCache` is a module-level `const` in the renderer process, and the repo's existing `test/` is Node `--test` with no Electron BrowserWindow harness. Standing up renderer-process testing felt out of scope for a fix PR; happy to follow up in a separate one if you'd like that infra, or refactor `_audioCache` into a mockable module. Let me know which you'd prefer.

### 3. `settings:preview-sound` bypasses DND (medium)

Handler now short-circuits on both `doNotDisturb` and `soundMuted`, returning `{ status: "skipped", reason }`. I went with honoring mute too per your "shared space" reasoning. The settings UI (`runSoundPreview`) treats `skipped` as silent — no error toast, since the user opted into that state deliberately.

### 4. Preview `_audioCache` growth (low)

Took the "don't cache `_t=` URLs at all" option — simpler than hooking `ended`. The preview path now creates a fresh `Audio` every click but never retains it, so memory is bounded regardless of how many previews happen.

---

Test coverage additions:
- `test/prefs.test.js` — new case verifying path-unsafe `soundName` keys (`../../evil`, `foo/bar`, `spaces bad`) are dropped and `file` values get basename-stripped (`../../etc/passwd` → `passwd`, bare `..` dropped).
- `npm test` — 866/873; the 7 failures are the same macOS-only pre-existing ones from the PR description (Windows install probing, calico SVG baselines, updater download harness).

Re: the maintainer-edit offer — appreciate it, all four ended up small enough that I went ahead. Let me know if anything above wants reworking.

rullerzhou-afk:
Quick heads-up — pushed a rebase on your branch (`fb4fd1f`). Your 4 review-response fixes in 5bcb29b are all intact, this was just protocol drift on `main` that wasn't visible from this PR:

**What changed on main while this PR was open:**
- #151 (sound effect volume slider) changed the `play-sound` IPC payload from a bare URL string to `{ url, volume }`
- #152 added `getPreviewSoundUrl()` to theme-loader for the volume-slider's own preview (separate from this PR's per-slot preview)

**What I had to adjust:**
1. `src/renderer.js` `onPlaySound` — merged both versions: unpacks `{ url, volume }` payload, clamps volume, **and** keeps your `isPreview` cache-bypass logic
2. `src/preload.js` — kept main's `payload` parameter name for `onPlaySound`, added your `onInvalidateSoundCache` binding alongside
3. `src/theme-loader.js` — both `getPreviewSoundUrl` (from main) and `getSoundOverridesDir` (from this PR) now in module exports
4. `src/main.js` `settings:preview-sound` — updated the send payload from bare `bustedUrl` to `{ url: bustedUrl, volume: soundVolume }` so preview honors the user's volume setting (would've otherwise always played at 1.0)

No behavior changes to your hardening work (regex validation, cache invalidation broadcast, DND/mute short-circuit, preview cache bypass) — those all went through clean. Tests pass except 2 pre-existing calico SVG baseline failures unrelated to this PR.

Sorry for the drive-by force-push; figured it was lower overhead than pinging you to rebase around protocol changes you couldn't see from the PR view.

rullerzhou-afk (approved):
Rebased onto main and verified independently — all 4 review-round hardening fixes intact, merged cleanly with the new { url, volume } IPC protocol. LGTM.

PR #155, opened 2026-04-23
feat(settings): per-agent toggle for Notification hook alerts

Rladmsrl:
## Motivation

Claude Code fires a `Notification` hook ~60 seconds after every `Stop` to remind me a response is waiting. Single-session this is fine — it's the whole point of the bell. But I've been running three or four CC sessions in parallel lately (different worktrees / repos), and the idle pings from the ones I'm not actively looking at end up firing the red-exclamation bell a few times a minute. What's helpful becomes noise.

Today the only way to silence it is **Settings → Animation Map → Notification**, which disables the animation + sound for *all* paths into the `notification` state — including permission bubbles. That's a real cost: the bell for "Claude wants to run Bash" is exactly when I *want* the attention-grab.

This PR offers a surgical alternative.

## What it does

Adds a per-agent **"Wait-for-input alerts"** sub-switch under the existing Permissions toggle in Settings → Agents. Default on, so no behavior change unless the user explicitly opts out.

When off for an agent, the pet silently drops that agent's self-issued `Notification` hook events. Permission bubbles, Elicitations, completion bells, and everything else stay untouched.

## Design notes — where this diverges from a "CC-specific" hack

Three calls where I tried to optimize for cross-CLI uniformity rather than scratching the CC itch the shortest way:

1. **Gate lives in `state.js updateSession()`, not in `server.js /hook` POST.** CC / CodeBuddy / Kimi CLI events arrive over HTTP and would be caught at either layer. But Gemini uses `log-poll` and never hits `/hook` — an HTTP-layer gate would silently no-op for Gemini while appearing to work for others. State-layer gating is uniform across hook, log-poll, and any future plugin event source.

2. **Capability-gated UI.** Only agents whose `eventMap` actually carries `Notification` render the sub-switch. That's currently CC, CodeBuddy, Kimi. Gemini's `eventMap` has the entry marked `// retained for future use` and doesn't actually fire one — rather than show an inert toggle, I left `notificationHook` off for Gemini. Flip that bit when Gemini starts firing for real.

3. **i18n copy neutralized.** Label is "Wait-for-input alerts" (not "Claude Code idle alerts"); description reads "this agent's Notification hook (e.g. Claude Code fires one after ~60s idle)" so CC is an example, not the definition. Same framing in zh + ko.

Permission bubble safety: `PermissionRequest` is handled by the branch *before* the new gate inside `updateSession`, and CC/opencode's HTTP permission path lives in `/permission` with its own server.js dispatch — neither route passes through the Notification-hook gate. The three state.js tests for `PermissionRequest`, `Elicitation`, and "other events with flag off" cover this.

## Tests

- `test/state-notification-hook-gate.test.js` — drives `updateSession` directly. Drops `Notification` when flag off, lets it through when on, preserves `PermissionRequest` / `Elicitation` / other events, fails open on missing ctx reader, resolves agentId from the session when the event omits it.
- `test/agent-gate.test.js` — new `isAgentNotificationHookEnabled` block mirrors the existing `isAgentPermissionsEnabled` coverage (default-true, independence from master / sibling flags), plus a `setAgentFlag` case confirming the new flag is a pure data flip with no side effects.
- `test/prefs.test.js` — defaults seeded on all 8 known agents, flag preserved through `normalizeAgents`, pre-flag prefs files upgrade to `notificationHookEnabled=true` (so existing users keep hearing alerts unless they opt out).
- `npm test`: 1002/1010. The 8 failures are the same macOS-only Windows-install probing / calico SVG baseline / updater harness cases pre-existing on main.

Manual: verified on macOS with CC — toggled off, idle 60s → silent; toggled on, idle 60s → red bell; permission bubble request → red bell regardless of toggle state.

## This is a suggestion

I'm not confident the tradeoff is universal — the existing Animation Map "Notification" toggle already addresses the noise for users who are fine collateral-silencing permission bells, and adding a second switch with a narrower scope does grow the Agents tab. Posting this for your judgment on both fronts:

- whether the multi-session idle-ping problem is worth a dedicated setting, and
- if yes, whether this state-layer + capability-gated shape is the right implementation.

Happy to rework, narrow, or drop entirely based on your call. Maintainer-edit is on.


rullerzhou-afk:
@Rladmsrl — quick note now that this is in and the follow-up #157 is merged.

Honest admission: I had been hearing those 60s idle bells across my concurrent CC sessions for **weeks** and had convinced myself it was my own fault. I figured my personal hooks setup had grown too bloated and was polluting the hook pool somehow — kept telling myself "one of these days I'll clean up my hooks config". Your PR was the first time I even registered it as an upstream behavior with a real fix. So thanks for not assuming someone else would have noticed — I genuinely wouldn't have.

The follow-up in #157 handles one edge case codex flagged in review: Kimi uses `Notification` as a secondary signal to release its permission-poll hold (via `KIMI_HOLD_CLEAR_EVENTS` in `state.js`), so the gate couldn't sit at `updateSession` entry. You don't run Kimi so this wouldn't have bitten you in practice, and your `notificationHookEnabled` toggle should stay on for Claude Code — that path was working exactly as you designed it.

The design writeup, capability-gated UI, and i18n neutrality all saved me hours. Don't rework anything — just wanted you to see the follow-up context.

PR #168, opened 2026-04-24
Security: Electron sandbox explicitly disabled on Linux

tomaioo:
## Summary

Security: Electron sandbox explicitly disabled on Linux

## Problem

**Severity**: `High` | **File**: `launch.js:L15`

The launcher unconditionally sets `ELECTRON_DISABLE_SANDBOX=1` and starts Electron with `--no-sandbox --disable-setuid-sandbox` on Linux. This removes Chromium/Electron sandbox protections, so any renderer compromise has significantly higher impact (greater access to host resources/process).

## Solution

Avoid disabling the sandbox globally. Prefer running with sandbox enabled and fixing environment-specific issues via targeted packaging/runtime fixes. If a fallback is required, gate it behind an explicit opt-in flag (e.g., env var for troubleshooting) and default to sandbox ON in production.

## Changes

- `launch.js` (modified)

### DIFF
--- a/launch.js
+++ b/launch.js
@@ -14,14 +14,13 @@ const electron = require("electron");
 
 const env = { ...process.env };
 delete env.ELECTRON_RUN_AS_NODE;
-if (process.platform === "linux") {
-  // Some Linux environments still trip Chromium sandbox initialization even
-  // with argv flags; force-disable via env too for reliability.
+const disableSandbox = process.platform === "linux" && process.env.CLAWD_DISABLE_SANDBOX === "1";
+if (disableSandbox) {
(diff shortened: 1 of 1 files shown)


rullerzhou-afk:
Thanks, this looks reasonable to me.

This keeps the Linux source/dev launcher sandboxed by default, while preserving the previous workaround behind `CLAWD_DISABLE_SANDBOX=1` for environments that still need it.

One follow-up on our side: the Linux setup docs still describe the old default `--no-sandbox` behavior, so we will patch those docs separately after merge. I will keep Windows/WSL startup work out of this PR and handle it separately.

PR #187, opened 2026-04-26
Add animated hero gif to README

livlign:
## Summary

Adds an animated hero GIF (1040×500, 1.6 MB, 28s loop) to the top of the README, between the badges and the opening prose.

The GIF mirrors the real product: a Claude Code TUI on the left (banner, input box, `auto mode on` footer — all from the running app), with the pixel crab docked at the bottom-right corner reacting in real time. Five beats: **sleeping** (Z's float above the dozing crab) → **thinking** (head tilt, model reading the codebase) → **typing** (edit & bash tools) → **juggling** (3 parallel subagents) → **notification** (permission bubble pops, ⌃⇧Y to allow) → **happy** (14 files changed, 312 tests passing).

The right side of the header rotates a quiet status readout (`thinking · reading your codebase`, `juggling · 3 subagents in parallel`, etc.) in pace with the pet's state, so a viewer reads the meaning of each animation even if they skim past the terminal text.

## What's in the diff

- `assets/hero.gif` — 1.6 MB, 1040×500, GIF89a
- `README.md` — one centered `<p>` block right after the existing badges, before the "Clawd lives on your desktop…" paragraph. Alt text describes the full animated demo for screen readers.

## Notes

- The GIF uses real Claude Code output shape and the actual crab sprites from `assets/gif/`. Status phrases (`Sashing…`, `Ready`, `auto mode on`) are taken from a screen recording of the running app.
- File size: 1.6 MB, comfortably under GitHub's mobile-render budget.
- No volatile stats (stars/version) are baked into the image — those stay in the dynamic badges above so they don't drift away from reality.

Happy to revise pacing, copy, or placement if anything reads off.


rullerzhou-afk:
Wow — this is so much more than I expected when I opened the PR title. Thank you for actually building the thing rather than just dropping in a static screenshot.

The five-beat narrative (sleeping → thinking → typing → juggling → notification → happy) is exactly the pitch for what Clawd does, and the rotating status readout on the right is a really thoughtful touch — it means people who skim past the TUI text still walk away understanding the pet's role. Using the real sprites + real TUI shape (instead of a mockup), and deliberately keeping volatile stats like stars/version out of the image so it never drifts away from reality — both of those tell me you thought about this as a long-lived asset, not just a one-off contribution.

Already merged and it's now sitting right at the top of the README, exactly where it belongs. This is the kind of thing I never would have made time to build myself, and seeing the project through someone else's eyes for one beat made my morning. Hope to see you around again 🥹

PR #190, opened 2026-04-27
fix(hooks): narrow PermissionRequest matcher to Bash (#46193 mitigation)

godlockin:
## Summary
- Narrow the PermissionRequest HTTP hook matcher from `""` (match-everything) to `"Bash"` to reduce the blast radius of upstream [anthropics/claude-code#46193](https://github.com/anthropics/claude-code/issues/46193) (still open) — when Clawd's HTTP server is offline, CC currently silent-denies any tool call covered by the hook instead of falling through to the native permission flow.
- Edit / Write / mcp__* are already handled by CC's own permission mode (per the existing comment at `hooks/install.js:503`), so they should never have been routed through Clawd's hook in the first place.
- `syncHttpHook` learns an `expectedMatcher` parameter and idempotently rewrites *Clawd-owned* entries that have a stale matcher (e.g. legacy `""`). Only entries whose hooks array is exclusively Clawd-owned get touched — user-authored entries that happen to bundle a Clawd hook are left alone.

## Why this and not the full fix
The complete fix is to convert the PermissionRequest registration from an HTTP hook to a `command` wrapper hook (like `hooks/codex-hook.js`), so the wrapper can stdout `{}` (no-decision) when Clawd is offline and let CC fall through to its native prompt. That's a larger change behind its own ADR; this PR is the immediate-mitigation step that ships independently.

## Test plan
- [x] `npm test` — 4 new tests pass (fresh install writes `matcher:"Bash"`, legacy `""` upgrades to `"Bash"`, install is idempotent, user-authored entries are not touched). Pre-existing test failures are unrelated to this change (Claude version detection on Windows shims, Kiro Windows paths, updater visual flow — same set fails on baseline `main`).
- [x] Restart-free upgrade path: existing installs pick up the new matcher on next Clawd boot via `registerHooks` → `syncHttpHook` reconcile.
- [ ] Manual smoke on a live Claude Code session: `Bash ls` still routes through Clawd bubble; `Edit somefile` no longer routes through Clawd (falls through to CC native).

Refs: anthropics/claude-code#46193

### DIFF
--- a/hooks/install.js
+++ b/hooks/install.js
@@ -468,7 +468,7 @@ function removeMatchingHttpHooks(entries, predicate) {
   return { entries: nextEntries, removed, changed };
 }
 
-function syncHttpHook(entries, expectedUrl) {
+function syncHttpHook(entries, expectedUrl, expectedMatcher) {
   let found = false;
   let changed = false;
   if (!Array.isArray(entries)) return { found, changed };
@@ -482,13 +482,31 @@ function syncHttpHook(entries, expectedUrl) {
       }
     }
     if (!Array.isArray(entry.hooks)) continue;
+    let entryHasClawd = false;
+    let entryAllClawd = entry.hooks.length > 0;
     for (const hook of entry.hooks) {
-      if (!isClawdPermissionHook(hook)) continue;
-      found = true;
-      if (hook.url !== expectedUrl) {
-        hook.url = expectedUrl;
-        changed = true;
+      if (isClawdPermissionHook(hook)) {
+        entryHasClawd = true;
+        found = true;
+        if (hook.url !== expectedUrl) {
+          hook.url = expectedUrl;
+          changed = true;
+        }
+      } else {
+        entryAllClawd = false;
       }
+    }
+    // Upgrade matcher on Clawd-owned entries that have a stale (e.g. legacy "")
+    // matcher. We only touch entries whose hooks are exclusively Clawd's, so we
+    // never modify user-authored entries that happen to bundle a Clawd hook.
+    if (
+      entryHasClawd
+      && entryAllClawd
+      && typeof expectedMatcher === "string"
+      && entry.matcher !== expectedMatcher
+    ) {
+      entry.matcher = expectedMatcher;
+      changed = true;
     }
   }
   return { found, changed };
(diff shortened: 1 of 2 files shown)


rullerzhou-afk:
Thanks for the contribution. I reviewed this carefully against the current Claude Code behavior, the linked upstream issue, and the existing Clawd runtime path.

I am going to close this PR rather than merge it.

The main reason is that the premise for this change appears to be stale. `anthropics/claude-code#46193` is still open, but the original reporter added an update on 2026-04-18 saying the issue could no longer be reproduced on Claude Code 2.1.113 on Windows. With the `PermissionRequest` HTTP hook pointing at a dead local endpoint, `Write` and `Edit` fell through to Claude Code's default behavior instead of being silently denied. That matches the current hooks documentation, which says HTTP connection failures and timeouts are non-blocking.

There is also a product-behavior issue here that is more important than the implementation details. Changing the `PermissionRequest` matcher from `""` to `"Bash"` is not just a small blast-radius mitigation. It narrows Clawd's permission bubble coverage:

- Current behavior: any Claude Code `PermissionRequest` that reaches Clawd can show the Clawd permission bubble, including non-Bash tools when Claude Code decides they need permission.
- Behavior after this PR: only `Bash` permission requests would go through Clawd. `Edit`, `Write`, MCP tools, and any other permission-gated tools would fall back to Claude Code's native prompt instead.

That is a real product decision, not something we should ship as an incidental workaround for an upstream bug that currently appears fixed in newer Claude Code versions.

The existing comment in `hooks/install.js` saying that Edit/Write are handled by Claude Code's own permission mode is historical and easy to misread. It does not mean Clawd should ignore `Edit` or `Write` when Claude Code emits a real `PermissionRequest` for them. The server-side `/permission` handler also intentionally does not filter by tool name.

I also noticed that if we ever did want to make the matcher a strict invariant, this PR would still need more migration coverage. For example, the settings watcher currently checks only whether the expected `/permission` URL exists, not whether the matcher is correct, and flat legacy HTTP hook entries are detected as existing but are not upgraded to the new matcher. Those are secondary because I do not think the narrowing itself is the right direction.

So the decision here is: keep the broad `PermissionRequest` matcher for Clawd, do not add a Clawd-side workaround for this upstream failure mode right now, and close this PR. We should separately clean up the stale docs/comments around #46193 so future contributors do not reasonably arrive at the same conclusion from old context.

Thanks again for taking the time to investigate and add tests. The implementation is careful; the part we are rejecting is the current product direction and the now-stale upstream premise.

PR #204, opened 2026-04-29
fix: arrow key navigation between options and Other textarea in elicitation bubble

### LINKED ISSUE #203: AskUserQuestion bubble: arrow keys stuck when Other textarea is focused
## Bug

When `AskUserQuestion` is shown and the user navigates down to select **Other**, the textarea appears and receives focus. At this point:

- Pressing **↑** does nothing useful (the browser moves the text cursor inside the empty textarea instead of returning focus to the preset options above)
- There is no way to navigate back to preset options with the keyboard — you have to click with the mouse

This is inconsistent with the CLI experience where ↑/↓ always navigate between options freely.

## Steps to reproduce

1. Trigger any `AskUserQuestion` with multiple options
2. Press ↓ to navigate to **Other** — textarea appears and gets focus
3. Press ↑ — expected: focus returns to the last preset radio; actual: nothing happens / cursor stays in textarea

## Expected behaviour

Arrow keys should work like the CLI:
- ↑ from Other textarea → focus returns to the last preset radio option
- ↓ from the Other radio → focus enters the textarea (when visible)

## Environment

- Platform: macOS
- Clawd on Desk v0.6.2

Ziy1-Tan:
Fixes #203

## Problem

When `AskUserQuestion` shows and the user navigates to **Other**, the textarea gets DOM focus. Arrow keys are then captured by the browser for text cursor movement — there's no way to navigate back to preset options with the keyboard, unlike the CLI where ↑/↓ always move between options.

## Fix

Two changes in `src/bubble.html`:

**1. `otherTextarea` keydown handler**
- ↑ when cursor is at position 0 (or textarea is empty) → `focus() + click()` the last preset radio
- ↓ when cursor is at the end (or textarea is empty) → focus the first preset radio
- Multi-line text editing unaffected when cursor is not at boundary

**2. `otherInput` keydown handler (new)**
- ↓ when the textarea is visible → move focus into the textarea
- Completes the bidirectional loop: `radio list ↔ ArrowUp/Down ↔ Other textarea`

## Test

1. Trigger any `AskUserQuestion` with multiple options
2. Press ↓ to reach **Other** — textarea appears, focus enters it
3. Press ↑ — focus returns to last preset radio and it becomes selected ✅
4. Press ↓ from the Other radio — focus moves into the textarea ✅

clawd

### DIFF
--- a/src/bubble.html
+++ b/src/bubble.html
@@ -932,29 +932,25 @@ function createElicitationQuestionCard(question, questionIndex) {
     updateElicitationSubmitState();
   });
   // Enter activates the primary action when it is enabled; Shift+Enter inserts a newline.
-  // ArrowUp when cursor is at position 0 escapes back to the last preset radio option,
-  // matching CLI arrow-key navigation behaviour where Up always moves to the previous item.
+  // ArrowUp from the start of Other returns focus to the last preset option.
   otherTextarea.addEventListener("keydown", (e) => {
     if (e.key === "Enter" && !e.shiftKey && !e.isComposing) {
       e.preventDefault();
       if (!btnAllow.disabled) btnAllow.click();
       return;
     }
-    if ((e.key === "ArrowUp" || e.key === "ArrowDown") && !e.shiftKey && !e.isComposing) {
-      // Only escape when the textarea is empty OR cursor is at the very start (ArrowUp)
-      // or very end (ArrowDown). This avoids hijacking multi-line navigation inside a
-      // filled textarea.
+    if (e.key === "ArrowUp" && !e.shiftKey && !e.isComposing) {
       const atStart = otherTextarea.selectionStart === 0 && otherTextarea.selectionEnd === 0;
-      const atEnd = otherTextarea.selectionStart === otherTextarea.value.length
-        && otherTextarea.selectionEnd === otherTextarea.value.length;
       const isEmpty = otherTextarea.value.length === 0;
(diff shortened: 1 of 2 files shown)


rullerzhou-afk:
Thanks for the quick fix. The stuck-focus path from #203 makes sense, and the IME / Shift / textarea-boundary guards are good details.

I think it would be safer to keep this PR a little narrower before merging, though.

Two edge cases look worth handling:

1. The textarea `ArrowDown` wrap-around to the first preset option goes beyond the original issue. In the empty-textarea case this is easy to hit accidentally: `ArrowDown` enters the textarea, then another `ArrowDown` immediately selects the first preset option, deselects Other, and hides the textarea. The custom text may still be in renderer state, but visually it can feel like the input disappeared.

2. `target.focus(); target.click();` is fine for radio questions, but `AskUserQuestion` also supports `multiSelect`, where preset options are checkboxes. In that case arrow navigation would toggle a checkbox answer, which is surprising for focus movement.

Could we keep this PR to the minimal safe behavior for #203?

- Keep `ArrowDown` on the Other option moving focus into the textarea.
- Keep `ArrowUp` from an empty/start-position Other textarea returning to the last preset option.
- For radio questions, selecting that preset via `click()` is OK.
- For multi-select checkbox questions, only move focus and avoid clicking/toggling.
- Avoid the textarea `ArrowDown` wrap-around for now unless we define the full wrap-navigation behavior separately.

Thanks again. With that small narrowing, this should stay aligned with #203 and avoid changing answers during keyboard navigation.

PR #206, opened 2026-04-29
fix: complete elicitation keyboard UX — window focus, no scroll, auto-select first option

### LINKED ISSUE #205: AskUserQuestion bubble: keyboard not working (arrow keys / Enter have no effect)
## Bug

When `AskUserQuestion` shows an elicitation bubble, **keyboard input doesn't work at all**:

- Arrow keys do nothing (can't navigate between options)
- Enter does nothing (can't submit)
- The only way to interact is with the mouse

## Root cause

All permission bubbles are shown via `bub.showInactive()`, which deliberately avoids stealing keyboard focus from the terminal. This is correct for regular permission prompts — stealing focus would cause Claude Code to detect "User answered in terminal" and produce false denials.

However, elicitation bubbles (`AskUserQuestion`) need keyboard input to be usable. Since `showInactive()` never grants the window keyboard focus, all keydown events are silently swallowed.

The window is already created with `focusable: true` for elicitation (vs `false` for permission prompts), but that only means the window *can* be focused — it still needs an explicit `bub.focus()` call after load.

## Expected behaviour

Arrow keys navigate between options; Enter submits the selected answer — matching the CLI experience.

## Environment

- Platform: macOS
- Clawd on Desk v0.6.2

Ziy1-Tan:
Fixes #205

### 1. `permission.js` — call `bub.focus()` for elicitation bubbles

`showInactive()` never grants keyboard focus, so arrow keys and Enter were silently dropped. Gated on `isElicitation` so regular permission bubbles are completely unaffected.

### 2. `bubble.html` — `applyElicitationViewport()` → no-op

The `overflow-y: auto` scroll container caused Chromium to consume ArrowUp/Down for scrolling before any JS handler could fire. `preventDefault()` / `stopPropagation()` cannot intercept this — Chromium's scroll-on-arrow fires at the default-action phase, not the JS event phase.

Fix: remove the `maxHeight` clamp and `elicitation-scrollable` class. Window height is driven by `reportHeight() → IPC bubble-height → setBounds()`. `permission.js` `clampBubbleHeight()` already caps the window at `workArea.height` as the overflow safety net, so content can never exceed the screen.

### 3. `bubble.html` — auto-focus first radio on `renderElicitationStep()`

Without an explicitly focused radio, the first ArrowDown after the bubble appears moves browser focus into the radio group but doesn't select anything — requiring an extra keypress. `rAF focus() + click()` on the first preset radio on render fixes this. Navigating back to a previously answered question keeps the existing selection.

---

## Safety analysis — "User answered in terminal" regression

The original `bub.focus()` warning was about **regular permission prompts**: stealing terminal focus can cause CC to detect a terminal interaction and fire a false deny via the PostToolUse/Stop hook path (`server.js:694`).

**Elicitation is immune** because:
- It uses a completely separate code path (`server.js:1008`, `isElicitation: true`)
- It is explicitly excluded from shortcut-nav and resolve logic (`permission.js:321, 630`)
- `AskUserQuestion` has no terminal-answer fallback by design; the only exit paths are bubble submit or "Go to Terminal" (explicit deny)

---

## Result

| Before | After |
|--------|-------|
| Arrow keys do nothing | Arrow keys navigate options immediately |
| First option not selected | First option auto-selected on open |
| Scrollbar appears with many options | Window grows to fit content, no scrollbar |
| Must click with mouse | Fully keyboard-navigable |

### DIFF
--- a/src/bubble.html
+++ b/src/bubble.html
@@ -669,15 +669,25 @@ function measureNaturalBubbleHeight() {
 }
 
 function applyElicitationViewport() {
-  if (!elicitationMode || !elicitationForm.classList.contains("visible")) return;
-  const cardRect = card.getBoundingClientRect();
-  const formRect = elicitationForm.getBoundingClientRect();
-  const fixedCardHeight = Math.max(0, cardRect.height - formRect.height);
-  const availableFormHeight = Math.floor(window.innerHeight - BUBBLE_BODY_PADDING_Y - fixedCardHeight);
-  const maxFormHeight = Math.max(MIN_ELICITATION_FORM_HEIGHT, availableFormHeight);
-
-  elicitationForm.style.maxHeight = `${maxFormHeight}px`;
-  card.classList.add("elicitation-scrollable");
+  // Intentionally a no-op.
+  //
+  // Previously this function clamped the elicitation form's maxHeight and added
+  // the `elicitation-scrollable` class (overflow-y: auto). The scroll container
+  // caused arrow keys to scroll the div instead of navigating between radio
+  // options — even with preventDefault()/stopPropagation() on keydown — because
+  // Chromium's scroll-on-arrow default action fires before JS handlers in the
+  // bubble phase, not after.
+  //
+  // The correct approach: let the form grow to its natural height and drive
+  // window size through reportHeight() → IPC bubble-height → setBounds().
+  // permission.js clampBubbleHeight() already caps the window at workArea.height
+  // so content-heavy bubbles will never exceed the screen.
+  //
+  // Safety: "User answered in terminal" cannot be triggered by elicitation
+  // bubbles. That denial path is wired to PostToolUse/Stop hook events matched
+  // by toolUseId (server.js:694) — elicitation uses a completely separate
+  // code path (server.js:1008, isElicitation: true) and is explicitly excluded
+  // from the shortcut-navigation and resolve logic (permission.js:321, 630).
 }
 
 function scheduleBubbleHeightReport() {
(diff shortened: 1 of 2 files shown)


rullerzhou-afk:
Thanks for looking into the broader keyboard-focus issue here. The `permEntry.isElicitation -> bub.focus()` part makes sense to me for #205, since elicitation bubbles do need keyboard focus while regular permission bubbles should keep the existing non-focus behavior.

I had a couple of concerns about the `bubble.html` changes, though. They may be worth narrowing before this merges.

First, making `applyElicitationViewport()` a no-op might bring back reachability problems for content-heavy elicitation prompts. `clampBubbleHeight()` caps the BrowserWindow height, but it does not make the clipped content scrollable. Since the page/card still hide overflow, long multi-question or multi-option prompts could end up with some content or footer actions out of reach. The previous internal scroll clamp was doing useful work there by keeping the form scrollable while leaving Back/Submit outside the scroll area.

Second, `first.focus(); first.click();` changes the answer during initial focus setup. For radio questions this silently selects the first option; for `multiSelect` questions the first preset is a checkbox, so this would automatically toggle an answer. This is the same principle as the #204 keyboard-navigation feedback: focus movement should not mutate answers unless the user explicitly selects.

Could we consider keeping this PR focused on the #205 fix?

- Keep `bub.focus()` for elicitation bubbles.
- Keep the internal overflow/viewport behavior, or replace it with another scroll strategy that keeps long prompts and footer actions reachable.
- Focus the first option without clicking/toggling it.
- Leave the textarea-specific escape behavior to #204.

That should solve the “keyboard input does not work at all” issue without changing default answers or risking clipped elicitation content. Thanks again for digging into this.

rullerzhou-afk:
Thanks for the PR. Since there has not been further follow-up here and the window-focus part unblocks AskUserQuestion keyboard input, I am going to merge this and track the remaining UX concerns separately.

Follow-up issue: #222

The follow-up covers the two risks from the review thread: preserving reachability for long/content-heavy elicitation prompts, and replacing the initial `first.focus(); first.click();` path so focus setup does not mutate the user's answer.

PR #209, opened 2026-04-29
fix: dismiss stale elicitation bubbles when user answers in terminal

joshua-wu:
## Summary

- AskUserQuestion (elicitation) bubbles persist on screen after the user already answered in the CLI terminal
- Root cause: the `/state` PostToolUse cleanup can't match the elicitation entry because the `tool_input_fingerprint` differs (PostToolUse includes the user's answer, but the original `/permission` request only had the question)
- Added a sweep after the exact-match cleanup that resolves any remaining `isElicitation` entries for the same session when forward progress is observed — safe because AskUserQuestion is a blocking tool call

## Test plan

- [x] Added test: stale elicitation bubble is swept when any PostToolUse fires for the same session
- [x] Added test: non-elicitation bubbles and other sessions are not affected
- [x] All 16 tests in `server-permission-state-cleanup.test.js` pass
- [x] Full test suite shows no regressions (same 5 pre-existing failures)

## Files changed

- `src/server.js` — 11-line elicitation sweep after exact-match cleanup (line 923)
- `test/server-permission-state-cleanup.test.js` — 2 new test cases (60 lines)

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -920,6 +920,17 @@ function startHttpServer() {
                 allowSingletonFallback: event === "Stop",
               });
               if (perm) ctx.resolvePermissionEntry(perm, "deny", "User answered in terminal");
+              // Stale elicitation sweep: AskUserQuestion is a blocking tool
+              // call, so any forward progress in the same session means the
+              // user already answered in the terminal.  The exact-match above
+              // may miss the elicitation entry when the /state PostToolUse
+              // carries a different tool_input fingerprint from the original
+              // /permission request, or when tool_use_id is absent.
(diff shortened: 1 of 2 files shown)


rullerzhou-afk:
Thanks @joshua-wu for fixing #209! This resolves the stale AskUserQuestion / elicitation bubble that could remain visible after the user had already answered in the Claude Code terminal. The same-session cleanup on later PostToolUse / PostToolUseFailure / Stop events is nicely scoped, and the added regression tests cover the important edge cases. I also verified the terminal-answer flow manually, and it behaves as expected.

PR #213, opened 2026-04-30
fix: allow resizing pet when pinned to screen edge in mini mode

### LINKED ISSUE #212: bug: resize has no effect when pet is pinned to a screen edge
## Description

When the pet is positioned at or near a screen edge (right, bottom, or top with edge-pinning enabled), resizing via the context menu or settings preview has no visible effect — the pet stays at the edge without repositioning to accommodate the new size.

## Root Cause

The resize path in `menu.js` (`resizeWindow`) reads the current position via `getPetWindowBounds()`, which returns *virtual* coordinates (real y minus `viewportOffsetY`). It then calls `clampToScreenVisual(x, virtualY, newW, newH)`, which internally queries the nearest workArea using the virtual center point — this can produce an incorrect workArea or wrong boundary values when `viewportOffsetY > 0` (i.e. when the pet is pinned to the top edge).

More broadly, when the pet is already clamped to an edge at the *old* size, the `clampToScreenVisual` result for the *new* size can produce the same (x, y) output as the old virtual position. `applyPetWindowBounds` then materializes that to the same real coordinates, so `win.setBounds` changes the size but not the position — the pet silently overflows the edge.

Concrete example: pet pinned to the right edge with width `oldW`. After resize to `newW > oldW`, the right portion of the pet clips off-screen instead of the window shifting left.

## Affected files

- `src/menu.js` — `resizeWindow` (L371-393)
- `src/main.js` — `clampToScreenVisual` (L3489-3511) / `getPetWindowBounds` (L646-655)

## Steps to Reproduce

1. Drag the pet to any screen edge (right, bottom, or top).
2. Open the context menu and change the size (e.g. S → M or M → L).
3. The pet's visual position does not change — it stays clipped at the edge instead of shifting inward to stay fully visible.

## Expected Behavior

After resize, the pet should be re-clamped so it remains fully (or correctly partially) within the screen boundary at the new size, just as it would after a drag ends.

## Platform

Reproducible on macOS; likely also Windows and Linux.

nmsn:
## Summary

- Fixed the pet not resizing when pinned to screen edge (mini mode)
- Fixed `mini.js handleResize` using `ctx.SIZES[sizeKey]` which only has S/M/L, failing for proportional size keys (e.g. `P:6.9`) and falling back to current size unchanged
- Added `getPixelSizeFor` to `_miniCtx` so mini mode can compute proportional pixel sizes
- `clampToScreenVisual` now accepts an `allowEdgePinning` option override, used by `resizeWindow` to keep the pet fully on-screen after resize

## Root Cause

Two bugs prevented resizing when the pet was edge-pinned (mini mode):

1. `handleResize` in `mini.js` used `ctx.SIZES[sizeKey]` to resolve the new size, but `SIZES` only contains fixed `S`/`M`/`L` entries. For proportional keys like `P:6.9`, the lookup failed and fell back to `_getSize()` which returned the **current** size — so the window bounds never changed.

2. `_miniCtx` did not expose `getPixelSizeFor`, so even if the fallback logic tried to use it, the function was unavailable.

## Test Plan

- [x] Drag pet to right edge → enters mini mode
- [x] Drag size slider in settings panel → pet resizes correctly
- [x] Click size tick marks → pet resizes correctly
- [x] Resize when NOT at edge → still works as before
- [x] Context menu S/M/L resize → still works

Fixes #212

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -3557,7 +3557,7 @@ function clampToScreenVisual(x, y, w, h, options = {}) {
   const clampMargins = getRestClampMargins({
     height: h,
     visibleMargins: margins,
-    allowEdgePinning: allowEdgePinningCached,
+    allowEdgePinning: "allowEdgePinning" in options ? options.allowEdgePinning : allowEdgePinningCached,
     bottomInset,
   });
   return {
@@ -3584,6 +3584,7 @@ const _miniCtx = {
   SIZES,
   getCurrentPixelSize,
   getEffectiveCurrentPixelSize,
+  getPixelSizeFor,
   isProportionalMode,
   sendToRenderer,
   sendToHitWin,
--- a/src/menu.js
+++ b/src/menu.js
@@ -381,7 +381,7 @@ module.exports = function initMenu(ctx) {
     if (!ctx.miniHandleResize(sizeKey)) {
       if (ctx.win && !ctx.win.isDestroyed()) {
         const { x, y } = ctx.getPetWindowBounds();
-        const clamped = ctx.clampToScreenVisual(x, y, size.width, size.height);
+        const clamped = ctx.clampToScreenVisual(x, y, size.width, size.height, { allowEdgePinning: false });
         ctx.applyPetWindowBounds({ ...clamped, width: size.width, height: size.height });
       }
     }
--- a/src/menu.js
+++ b/src/menu.js
@@ -381,7 +381,7 @@ module.exports = function initMenu(ctx) {
     if (!ctx.miniHandleResize(sizeKey)) {
       if (ctx.win && !ctx.win.isDestroyed()) {
         const { x, y } = ctx.getPetWindowBounds();
-        const clamped = ctx.clampToScreenVisual(x, y, size.width, size.height, { allowEdgePinning: false });
+        const clamped = ctx.clampToScreenVisual(x, y, size.width, size.height);
         ctx.applyPetWindowBounds({ ...clamped, width: size.width, height: size.height });
       }
     }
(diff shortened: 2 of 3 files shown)


rullerzhou-afk:
I think the direction of this PR is right, and I appreciate that the change stays very small. Before merging, I’d like to tighten two small points so we do not accidentally change the existing resize and multi-display behavior.

1. I’m a bit concerned about the new `{ allowEdgePinning: false }` in `menu.js`. This branch runs when `miniHandleResize()` returns false, so it also affects normal-mode resizing. If the user has Allow pinning enabled and the pet is currently pinned to the top/bottom edge, resizing will now pull it back fully inside the screen, and a committed resize will persist that new position. That may be a valid UX choice, but it feels outside the scope of this PR, which is focused on mini-mode resize. I’d prefer not to override this here and let `clampToScreenVisual()` continue using the current setting. If we want “resize exits edge pinning” behavior later, we can discuss that UX separately.

2. In `mini.js`, `handleResize()` currently picks the workArea using `currentMiniX + newSize.width / 2`. On adjacent multi-monitor setups, if the pet is pinned to the right edge of the left display and the user resizes from a smaller size to a larger one, that computed center can land on the right display, causing the pet to jump monitors after resize. I’d suggest determining the workArea from the current window’s existing bounds first, then using the target size to compute the new mini x. For proportional sizes, we can also pass that same workArea into `getPixelSizeFor(sizeKey, wa)`, so size calculation and placement are based on the same display.

The `typeof ctx.getPixelSizeFor === "function"` fallback is fine to keep; I do not think it is an issue.

Overall this looks very close to merge-ready to me. I’d just like to see these two edge cases tightened first.

PR #216, opened 2026-04-30
feat: add iTerm2 tab-level focus switching

joshua-wu:
## Summary

- **Root cause**: Clicking Session HUD only brought iTerm2's window to front via System Events AppleScript (`set frontmost`), but could not switch to the specific tab — all sessions shared the same iTerm2 main process PID, so the operation was invisible when already in iTerm2.
- **Fix**: Add `scheduleITermTabFocus()` in `src/focus.js` that detects iTerm2 by checking `sourcePid`'s process name, resolves the shell TTY from `pidChain`, then runs iTerm2-native AppleScript to `select` the matching tab.
- Non-iTerm2 terminals are unaffected (auto-skipped). 4 new unit tests added.

## How it works

1. `ps -o comm=` on `sourcePid` → detect if terminal is iTerm2
2. `ps -o tty=` on `pidChain` candidates → find the shell's TTY (e.g. `ttys003`)
3. iTerm2 AppleScript iterates `windows > tabs > sessions`, matches by `tty of s ends with "<ttyName>"`, then `select` that tab + window
4. Fires 400ms after existing System Events window focus to avoid race conditions

## Test plan

- [x] Unit tests pass (`node --test test/focus-iterm-tab.test.js` — 4/4 pass)
- [x] Full test suite — no regressions (same 6 pre-existing failures)
- [x] Manual verification: clicking Session HUD row switches to correct iTerm2 tab
- [ ] Verify non-iTerm2 terminals (Terminal.app, Alacritty, etc.) still work as before

### DIFF
--- a/src/focus.js
+++ b/src/focus.js
@@ -164,6 +164,53 @@ function scheduleTerminalTabFocus(editor, pidChain) {
   }, 800);
 }
 
+function scheduleITermTabFocus(sourcePid, pidChain) {
+  if (!isMac || !sourcePid || !Array.isArray(pidChain) || !pidChain.length) return;
+  execFile("ps", ["-o", "comm=", "-p", String(sourcePid)], { encoding: "utf8", timeout: 500 }, (err, stdout) => {
+    if (err) return;
+    const name = path.basename(stdout.trim()).toLowerCase();
+    if (name !== "iterm2") return;
+
+    // Find the shell PID's TTY to match against iTerm2 sessions.
+    // Walk pidChain from agent (index 0) upward — the first PID with a valid TTY
+    // is typically the shell or login process that owns the iTerm2 session.
+    const candidates = pidChain.filter(p => Number.isFinite(p) && p > 0 && p !== sourcePid);
+    if (!candidates.length) return;
+
+    const pidsArg = candidates.slice(0, 5).join(",");
+    execFile("ps", ["-o", "pid=,tty=", "-p", pidsArg], { encoding: "utf8", timeout: 500 }, (psErr, psOut) => {
+      if (psErr || !psOut) return;
+      let ttyName = null;
+      for (const line of psOut.trim().split("\n")) {
+        const parts = line.trim().split(/\s+/);
+        if (parts.length >= 2 && parts[1] !== "??" && parts[1] !== "?") {
+          ttyName = parts[1];
+          break;
+        }
+      }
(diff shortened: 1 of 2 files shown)


rullerzhou-afk:
The direction makes sense to me: iTerm2 tabs share the same app/main process PID, so `set frontmost` can bring iTerm2 forward but cannot reliably select the specific tab. Matching the iTerm2 session by TTY is the right shape of fix.

That said, I don't think this PR is ready to merge yet. A few issues need fixing first:

1. `test/focus-iterm-tab.test.js` does not mock `process.platform`

`src/focus.js` captures `isMac` at require time via `process.platform === "darwin"`. The new tests describe macOS behavior, but they do not set `process.platform` to `darwin` before requiring `../src/focus`.

On Windows, running:

```bash
node --test test\focus-iterm-tab.test.js
```

fails with:

```text
Expected >= 2 ps calls, got 0
```

because the test actually goes through the Windows path and `scheduleITermTabFocus()` never runs. The test should temporarily mock `process.platform` before requiring `focus.js` and restore it during cleanup, or `focus.js` should accept platform as an injectable dependency.

2. The shared-`sourcePid` root cause is still partially present

The PR is trying to fix the fact that multiple iTerm2 tabs share the same `sourcePid`, but the macOS focus cooldown still dedupes only by `sourcePid`:

```js
if (inCooldown && macFocusLastPid === sourcePid) return;
```

If a user clicks from one iTerm2 session to another within the 1.5s cooldown window, both sessions can have the same `sourcePid`, so the second request is dropped before the new `pidChain` / TTY ever reaches `scheduleITermTabFocus()`.

The cooldown key needs to distinguish the target session, at least by `pidChain` or by the resolved TTY, rather than only by `sourcePid`.

3. TTY selection should not depend on `ps` output order

The comment says we walk `pidChain` from the agent upward and use the first valid TTY. The implementation passes several PIDs to one `ps` call and then uses the first valid line from `ps` output:

```js
ps -o pid=,tty= -p <comma-list>
```

That does not explicitly preserve `pidChain` priority. Safer approach: parse the `ps` output into a `pid -> tty` map, then iterate the original `candidates` array and pick the first valid TTY in that order.

Conclusion: I agree with the approach, but this needs another pass before merging. Please fix the test platform mocking, make the macOS cooldown target-aware, and make TTY selection follow the original `pidChain` order.

Issue #96, opened 2026-04-11
[Feature Request] Support WSL


Mars160:
希望支持WSL

rullerzhou-afk:
Hi @Mars160 👋

Clawd 桌宠本体是 Electron 应用，只能跑在 Windows 上。但如果你的 Claude Code（或 Kiro / Cursor / Gemini / opencode）跑在 WSL 里，**Windows 上的 Clawd 完全可以感知到它** —— Clawd 用 HTTP + hook 架构，天然支持"桌宠在 Windows + agent 在 WSL"这种组合。

## 前置条件

- **WSL1** — 无需网络配置（WSL1 与 Windows 共享网络栈），直接跳到下面"Hook 注册"步骤
- **WSL2** — 需要开启 mirrored networking 模式。**要求 Windows 11 22H2 及以上**（Build 22621+），且 WSL version 2.0+（在 PowerShell 跑 `wsl --version` 确认）。Windows 10 用户暂时不支持，请在本 issue 下留言告诉我你的环境

## WSL2 网络配置

编辑（或创建）`%USERPROFILE%\.wslconfig`，加入：

```ini
[wsl2]
networkingMode=mirrored
```

然后在 PowerShell 里跑 `wsl --shutdown` 重启 WSL。这步让 WSL2 和 Windows 共享 localhost，WSL 里的 `127.0.0.1:23333` 就能直接打到 Windows 上的 Clawd。

## Hook 注册（在 WSL 里）

Clawd 启动时会自动把 hook 注册到 **Windows 家目录** 下的 `~/.claude/settings.json`。但 WSL 有自己独立的家目录（`/home/<user>/`），WSL 里的 agent 读的是那边的配置文件——需要手动在 WSL 里跑一遍 install 脚本：

```bash
git clone https://github.com/rullerzhou-afk/clawd-on-desk.git
cd clawd-on-desk

# Claude Code
node hooks/install.js

# Kiro CLI
node hooks/kiro-install.js

# Cursor Agent
node hooks/cursor-install.js

# Gemini CLI
node hooks/gemini-install.js

# opencode
node hooks/opencode-install.js
```

> 仓库 clone 在 WSL home 目录（`~/clawd-on-desk`）会自动使用 WSL 的 Node.js；如果 clone 在 Windows 盘（`/mnt/c/...`），要确保 WSL 的 PATH 里有 `node`。

---

[...]

PR #218, opened 2026-05-01
feat: add WSL2 terminal focus support

sunnysonx:
## Problem                                                                                                                                                                                                                              
                                                                               
  When Claude Code runs inside WSL2, hook payloads carry Linux PIDs. The Windows-side Electron process tree walk uses `Get-Process -Id <pid>` which only sees Windows PIDs — Linux PIDs are invisible, so the walk always fails and VS    
  Code never gets focused.
                                                                                                                                                                                                                                          
  ## Solution                                                                  
                                               
  - Added `makeEditorFallbackBlock` in `src/focus.js`: when the PID walk and Windows Terminal scan both find nothing, falls back to locating VS Code or Cursor directly by process name (`Code` / `Cursor`). Tries title-matching by cwd  
  folder name first for multi-window accuracy, then takes the first visible window.
  - Threaded `editor` param through `makeFocusCmd` and `focusTerminalWindowLegacy` so the fallback knows which editor to target.                                                                                                          
  - Added `TERM_PROGRAM` → `editor` fallback in `hooks/clawd-hook.js`: VS Code injects `TERM_PROGRAM=vscode` into WSL terminals, so when process tree detection yields no editor this provides the signal needed to trigger the PowerShell
   fallback.                                                                                                                                                                                                                              
                                                                                                                                                                                                                                          
  ## Setup requirement                                                                                                                                                                                                                    
                                                                               
  WSL2 must use mirrored networking so hook HTTP calls reach the Windows-side Clawd server (as per https://github.com/rullerzhou-afk/clawd-on-desk/issues/96#issuecomment-4230348542)

  ```ini
  # %USERPROFILE%\.wslconfig
  [wsl2]
  networkingMode=mirrored                                                                                                                                                                                                                 
   ```

  Tested on                                                                                                                                                                                                                               
```                                                     
  Windows + WSL2 (Ubuntu) + VS Code Remote-WSL 
```


https://github.com/user-attachments/assets/475789d1-1bbd-4a9c-8ef0-c6c2ef077dc0

### DIFF
--- a/hooks/clawd-hook.js
+++ b/hooks/clawd-hook.js
@@ -167,7 +167,10 @@ function buildStateBody(event, payload, resolve) {
   } else {
     const { stablePid, agentPid, detectedEditor, pidChain } = resolve();
     body.source_pid = stablePid;
-    if (detectedEditor) body.editor = detectedEditor;
+    const termProgram = process.env.TERM_PROGRAM;
+    const editorFromEnv = termProgram === "vscode" ? "code" : termProgram === "cursor" ? "cursor" : null;
+    const resolvedEditor = detectedEditor || editorFromEnv;
+    if (resolvedEditor) body.editor = resolvedEditor;
     if (agentPid) {
       body.agent_pid = agentPid;
       body.claude_pid = agentPid; // backward compat with older Clawd versions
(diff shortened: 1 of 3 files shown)


sunnysonx:
Transfering to draft, to do some additional testing.

PR #220, opened 2026-05-02
fix: guard renderer crash logging against EIO errors

YuChenYunn:
## Summary
- Wrap `console.error` in `safeConsoleError` to prevent EIO crashes when logging renderer process failures
- Applied to both `hitWin` and main window crash recovery handlers

## Problem
When the renderer process crashes, `console.error` itself can throw an EIO error (e.g., when stdout/stderr pipe is broken), causing the crash handler to fail silently without recovering.

## Solution
`safeConsoleError` catches EIO errors gracefully, with a fallback to writing logs to `clawd-main.log` in userData.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -404,6 +404,20 @@ function stopMonitorForAgent(agentId) {
   else if (agentId === "gemini-cli" && _geminiMonitor) _geminiMonitor.stop();
 }
 
+function safeConsoleError(...args) {
+  try {
+    console.error(...args);
+  } catch (err) {
+    if (err && (err.code === "EIO" || /EIO/.test(String(err.message || "")))) {
+      return;
+    }
+    try {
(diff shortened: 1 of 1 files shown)


rullerzhou-afk:
Thanks for the fix. The direction looks right to me: guarding crash-recovery logging is useful because a throwing console.error can prevent the renderer reload path from running.\n\nOne thing to adjust before merge: the PR summary says safeConsoleError falls back to clawd-main.log on EIO, but the implementation currently returns immediately for EIO, so that crash log is dropped. Please route EIO through the same file fallback path as other console.error failures, and silently swallow only if the file write also fails.\n\nAfter that small change, I think this is mergeable.

YuChenYunn:
Thanks — I updated `safeConsoleError` so `EIO` no longer gets dropped. It now falls through to the same `clawd-main.log` fallback as other `console.error` failures, and only silently swallows the error if the file write also fails.

PR #221, opened 2026-05-02
fix: spawn Electron directly in auto-start to avoid console window on Windows

jinqKing:
## Problem

On Windows, the current fix in `hooks/auto-start.js` (`cmd.exe /c npm.cmd start` from PR #216) still produces a visible console window because the `npm start` → `launch.js` → `electron` chain leaves stdio attached to a console.

## Fix

Spawn Electron directly instead of going through npm/launch.js — the same pattern already used in `launch.js`:

- `require("electron")` locates the Electron binary
- Delete `ELECTRON_RUN_AS_NODE` from the inherited environment (same as launch.js)
- Pass `windowsHide: true` to suppress any console window

This bypasses the `npm start` → `launch.js` → `spawn(electron, { stdio: "inherit" })` chain entirely, so no console window appears.

## Test plan

- [ ] Windows (source/dev mode): auto-start launches Clawd without a visible console window
- [ ] macOS/Linux (source mode): launch behavior unchanged
- [ ] Packaged Windows (.exe): launch path unchanged
- [ ] Packaged macOS (.app): launch path unchanged
- [ ] Packaged Linux (AppImage/bin): launch path unchanged

### DIFF
--- a/hooks/auto-start.js
+++ b/hooks/auto-start.js
@@ -56,24 +56,19 @@ function launchApp() {
         }
       }
     } else {
-      // Source / development mode
+      // Source / development mode — spawn Electron directly to avoid
+      // a visible console window from the npm → launch.js chain.
       const projectDir = path.resolve(__dirname, "..");
-      if (isWin) {
-        // On Windows, .cmd files cannot be spawned directly with detached:true
-        // (EINVAL). Wrap with cmd.exe /c instead.
-        spawn("cmd.exe", ["/c", "npm.cmd", "start"], {
-          cwd: projectDir,
-          detached: true,
-          stdio: "ignore",
-          windowsHide: true,
-        }).unref();
-      } else {
-        spawn("npm", ["start"], {
-          cwd: projectDir,
-          detached: true,
-          stdio: "ignore",
-        }).unref();
-      }
+      const electron = require("electron");
+      const env = { ...process.env };
+      delete env.ELECTRON_RUN_AS_NODE;
+      spawn(electron, [projectDir], {
+        cwd: projectDir,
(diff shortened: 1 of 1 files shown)


rullerzhou-afk:
Thanks for chasing down the Windows console-window issue. The direct Electron spawn approach makes sense for that dev auto-start path.

One thing I think should be fixed before merging: this also changes the non-Windows source/dev path from `npm start` to direct `electron`, so it bypasses the Linux sandbox fallback in `launch.js`. Today `launch.js` honors `CLAWD_DISABLE_SANDBOX=1` by setting `ELECTRON_DISABLE_SANDBOX`, clearing `CHROME_DEVEL_SANDBOX`, and passing `--no-sandbox --disable-setuid-sandbox`. With this PR, Linux dev environments that depend on that documented workaround would fail to auto-start again.

Could you either mirror that Linux sandbox branch here or factor the Electron spawn arg/env construction into a small shared helper so `launch.js` and `auto-start.js` stay in sync? The Windows fix itself looks reasonable; I just want to avoid the source-mode Linux regression while this launcher path is being changed.

rullerzhou-afk:
Thanks @jinqKing for identifying the Windows source-mode auto-start console-window issue and proposing the direct Electron spawn approach.

I validated the follow-up implementation in #240 on Windows: with Clawd fully quit, Claude Code SessionStart launches Clawd successfully and no console window flash was observed.

I'm going to close this PR as superseded by #240. The follow-up keeps your core fix direction, but shares the Electron launch env/args with launch.js so we preserve the Linux CLAWD_DISABLE_SANDBOX fallback as well.

PR #229, opened 2026-05-03
Add Session HUD display toggles

jhseo-b:
## Summary
- add a Session HUD time display preference that defaults on
- hide elapsed time in the HUD when disabled and shrink the HUD width
- add a setting to hide ended detached sessions from the Session HUD shortly after they become idle
- apply the detached-session cleanup to all agents instead of only Codex

## Tests
- node --test test/prefs.test.js test/settings-actions.test.js test/session-hud.test.js test/session-hud-style.test.js
- node --test test/settings-renderer-browser-env.test.js
- node --test test/prefs.test.js test/settings-actions.test.js test/state.test.js test/settings-renderer-browser-env.test.js


rullerzhou-afk:
Thanks for the PR. The HUD elapsed-time toggle looks straightforward, but I found a couple of blockers around the detached-session cleanup before this should merge.

1. This PR currently does not merge into `main`.

`git merge-tree main HEAD` reports a content conflict in `src/state.js`. This is not just churn: current `main` added `resolvePidReachable()`, Codex exit probes, Kimi session disposal helpers, and `dismissSession()` cleanup paths after this branch's base. The new detached cleanup needs to be rebased into those newer paths rather than applied over the older `cleanStaleSessions()` shape.

2. The “HUD cleanup” setting deletes sessions globally, not just from the HUD.

The new `cleanStaleSessions()` branch does `sessions.delete(id)` for idle non-headless sessions with no `agentPid` after 30s. `sessions` is the shared backing store for Dashboard, menu, and HUD snapshots, so enabling this setting removes the session from all session views, not only the Session HUD. That does not match the setting/summary wording (`Session HUD`, `Hide ended sessions`, “Remove detached sessions from the Session HUD...”). If the intended behavior is HUD-only, this should be implemented in the HUD snapshot/filtering path instead of deleting the state entry.

3. `idle && !agentPid` is not a reliable “ended detached session” signal by itself.

Some sessions can still have a live `sourcePid` while `agentPid` was not resolved, and remote sessions naturally do not have local PIDs. With the current condition, an idle session can disappear after 30s even if it is still resumable/visible in Dashboard or can receive future events. If this should really delete from global state, the condition should at least account for `sourcePid`, `host`, and current `main`'s `pidReachable` logic, and preserve the existing Codex/Kimi cleanup side effects.

Verification I ran on the PR branch:

```bash
node --test test/prefs.test.js test/settings-actions.test.js test/session-hud.test.js test/session-hud-style.test.js test/settings-renderer-browser-env.test.js test/state.test.js
# pass: 353 tests

git diff --check main...HEAD
# pass

git merge-tree main HEAD
# fails: src/state.js content conflict
```

rullerzhou-afk:
Thanks for the follow-up commits. The elapsed-time toggle part looks good to me: default-on, scoped to the HUD renderer, and the compact width behavior is covered by tests.

I think the remaining open question is around the detached-session cleanup behavior.

Right now the PR seems to do two related but slightly different things:

1. `hiddenFromHud` hides some ended detached sessions from the HUD snapshot.
2. `cleanStaleSessions()` removes detached idle sessions from the global `sessions` map after 30s.

The Settings copy frames this as a Session HUD preference, but deleting from `sessions` also affects Dashboard, menu entries, aliases, pet state resolution, and cleanup side effects. Could we make the intended scope explicit?

My suggestion would be:

- If the goal is just to keep the HUD tidy, keep the `hiddenFromHud` filtering and avoid deleting the session from global state.
- If the goal is broader session cleanup, adjust the setting name/copy accordingly and make sure deletion goes through the same cleanup paths current `main` uses for Codex/Kimi/session dismissal.

One more thing: this branch still conflicts with current `main` in `src/state.js` and `test/state.test.js`. When rebasing, please be careful to preserve the newer `main` behavior around Codex exit probes, Codex thread-title lookup, `dismissSession()`, `resolvePidReachable()`, and Kimi cleanup helpers.

A few tests that would make this easier to review after the rebase:

- HUD time toggle still hides elapsed text and uses compact width.
- If this remains HUD-only, hidden sessions are still present in Dashboard/menu.
- Remote/host sessions are not hidden or deleted based on local PID checks.
- Codex cleanup still cancels exit probes.
- Kimi cleanup still clears hold/suspect timers and passive bubbles.

PR #234, opened 2026-05-04
fix: smooth Settings switch animations

YOIMIYA66:
No linked issue.

## Summary
- Smooths Settings switch motion across General, Agent management, and Animation Map so toggle feedback feels consistent instead of abrupt.
- Moves the shared switch knob animation to transform-based motion and reuses the existing animated switch path where tab-specific controls previously bypassed it.
- Preserves the existing Settings controller, prefs schema, command shape, and committed-state broadcast model.

## Problem context
- The General page switch felt smoother than Agent management and Animation Map because some tabs either rebuilt their content after broadcasts or used a direct activation handler instead of the shared optimistic switch behavior.
- In expanded Agent management groups, toggling a switch could be visually interrupted by a full content rebuild, which produced a close/reopen flash.
- Animation Map switches were also rebuilt after theme override broadcasts, so the improved switch CSS could be cut off before the motion was visible.

## What changed
- Shared Settings switch motion
  - Switched the knob movement from `left` updates to `transform: translateX(...)` for more stable animation.
  - Tuned track, knob, hover, active, pending, and reduced-motion behavior around one shared switch style.
- Language selector motion
  - Added a sliding active pill for the language segmented control instead of a direct visual jump.
  - Tracks the previous and next language during Settings broadcasts so the pill can animate between positions.
- Agent management patching
  - Keeps agent switch broadcasts on the mounted DOM instead of rebuilding when Codex permission mode rows are present.
  - Syncs master/sub-switch state and Codex permission-mode button state in place after broadcasts.
- Animation Map patching
  - Routes Animation Map toggles through `attachAnimatedSwitch()` so they get the same optimistic pending behavior as other Settings switches.
  - Patches `themeOverrides` changes in place for the current Animation Map controls instead of rebuilding the tab and truncating the animation.
  - Keeps the reset button disabled state in sync after local patching.
- Regression coverage
  - Added renderer browser-environment tests for transform-based switch CSS, reduced-motion handling, language sliding behavior, Agent management patching, and Animation Map local patching.

## Behavior after this PR
- Settings switches now use one calmer animation path across the affected tabs.
- Agent management groups stay expanded while switch broadcasts are applied, avoiding the previous flash caused by content teardown and rebuild.
- Animation Map switches visibly slide and keep pending feedback until the command result is reflected by the Settings snapshot.
- Failed switch updates still rely on the existing rollback/toast behavior; no new persistence fields or Settings controller behavior are introduced.
- Reduced-motion users get switch and language-selector transitions disabled through the existing CSS media query path.

## Validation
- `node --test test/settings-renderer-browser-env.test.js test/settings-controller.test.js test/settings-actions.test.js`
- `git diff --check`
- `npm test`

## Platform note
- Automated verification ran locally on Windows with the Node built-in test runner.
- A final manual pass in the Electron Settings window is still recommended for visual timing, especially `Settings -> General`, `Settings -> Agent management`, and `Settings -> Animation Map`.

### DIFF
--- a/src/settings-tab-agents.js
+++ b/src/settings-tab-agents.js
@@ -88,19 +88,17 @@
   }
 
   function buildAgentDetailRows(agent) {
-    const masterOn = readers.readAgentFlagValue(agent.id, "enabled");
     const rows = [];
     const caps = agent.capabilities || {};
     if (agent.id === "codex") {
-      rows.push(buildCodexPermissionModeRow(agent, !masterOn));
+      rows.push(buildCodexPermissionModeRow(agent, computeAgentSubSwitchDisabled(agent.id, "permissionMode")));
     }
     if (caps.permissionApproval || caps.interactiveBubble) {
-      const codexNativeMode = agent.id === "codex" && readers.readAgentPermissionMode(agent.id) !== "intercept";
       rows.push(buildAgentSwitchRow({
         agent,
         flag: "permissionsEnabled",
         extraClass: "row-sub",
-        disabled: !masterOn || codexNativeMode,
+        disabled: computeAgentSubSwitchDisabled(agent.id, "permissionsEnabled"),
         buildText: (text) => {
           const label = document.createElement("span");
           label.className = "row-label";
@@ -118,7 +116,7 @@
         agent,
         flag: "notificationHookEnabled",
         extraClass: "row-sub",
-        disabled: !masterOn,
+        disabled: computeAgentSubSwitchDisabled(agent.id, "notificationHookEnabled"),
         buildText: (text) => {
           const label = document.createElement("span");
           label.className = "row-label";
@@ -132,6 +130,16 @@
       }));
     }
     return rows;
+  }
+
+  function computeAgentSubSwitchDisabled(agentId, flag) {
+    if (flag === "enabled") return false;
+    const masterOn = readers.readAgentFlagValue(agentId, "enabled");
+    if (!masterOn) return true;
+    if (agentId === "codex" && flag === "permissionsEnabled") {
+      return readers.readAgentPermissionMode(agentId) !== "intercept";
+    }
+    return false;
   }
 
   function buildCodexPermissionModeRow(agent, disabled) {
@@ -281,7 +289,7 @@
     for (const [id, meta] of state.mountedControls.agentSwitches) {
       state.transientUiState.agentSwitches.delete(id);
       if (meta.flag !== "enabled") {
-        meta.syncDisabledState(!readers.readAgentFlagValue(meta.agentId, "enabled"));
+        meta.syncDisabledState(computeAgentSubSwitchDisabled(meta.agentId, meta.flag));
       }
       helpers.setSwitchVisual(meta.element, readers.readAgentFlagValue(meta.agentId, meta.flag), { pending: false });
     }
--- a/src/settings-tab-agents.js
+++ b/src/settings-tab-agents.js
@@ -169,7 +169,7 @@
       btn.disabled = !!disabled;
       btn.addEventListener("click", (ev) => {
         ev.stopPropagation();
-        if (disabled || btn.classList.contains("active")) return;
+        if (btn.disabled || btn.classList.contains("active")) return;
         window.settingsAPI.command("setAgentPermissionMode", {
           agentId: agent.id,
           mode: mode.id,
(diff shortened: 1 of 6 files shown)


rullerzhou-afk (commented):
Nice motion polish overall — the transform-based knob and the optimistic Animation Map switches both feel like the right direction. One blocker and a few smaller things worth flagging before merge.

### 🔴 Blocking — Codex `permissionsEnabled` switch stays clickable in Native mode after in-place patch

Initial render in `buildAgentDetailRows` disables the Permissions sub-switch when Codex is in Native mode:

```js
// src/settings-tab-agents.js:97-103
const codexNativeMode = agent.id === "codex"
  && readers.readAgentPermissionMode(agent.id) !== "intercept";
disabled: !masterOn || codexNativeMode,
```

But the new `patchInPlace` only re-derives disabled from the master switch:

```js
// src/settings-tab-agents.js:281-287
if (meta.flag !== "enabled") {
  meta.syncDisabledState(!readers.readAgentFlagValue(meta.agentId, "enabled"));
}
```

**Repro:** Settings → Agent management → expand Codex (master ON, mode = Intercept) → flip mode to Native. The mode buttons update correctly via `syncFromSnapshot`, but the Permissions row stays enabled/clickable until a full rerender (e.g. tab switch). Fresh render of the same snapshot would have it disabled — so `patchInPlace` is silently diverging from render truth.

The PR's new test `keeps Agent management switch broadcasts in place even when Codex permission rows are mounted` keeps `permissionMode: "intercept"` across both snapshots, so this path isn't exercised.

**Suggested fix:** extract the disabled predicate so render and patch share it:

```js
function computeAgentSubSwitchDisabled(agentId, flag) {
  if (flag === "enabled") return false;
  const masterOn = readers.readAgentFlagValue(agentId, "enabled");
  if (!masterOn) return true;
  if (flag === "permissionsEnabled" && agentId === "codex") {
    return readers.readAgentPermissionMode(agentId) !== "intercept";
  }
  return false;
}
```

Use it in both `buildAgentDetailRows` and the `patchInPlace` sub-switch loop, and add a parallel test that flips `permissionMode` from `"intercept"` to `"native"` and asserts the Permissions sub-switch ends up disabled.

---

### 🟡 Non-blocking observations (worth a look, not gating merge)

**1. Language pill text color jumps ahead of the slide**
`.segmented button` is `color: var(--text-secondary)`, `.segmented button.active` is `color: var(--text-primary)`. On lang change, the new button gets `.active` synchronously at render time, so its text color flips immediately while the `::before` pill is still in-flight (0.24s). The slide motion this PR adds is real, but the color change still snaps. Consider transitioning button color, or deferring the `.active` class swap until after the pill arrives.

**2. Knob press feedback slowed from 0.14s to 0.28s**
Old `.switch::after` had `transform 0.14s ease`; the consolidated rule now uses `transform 0.28s cubic-bezier(...)` for both the slide and the `:active` `scale(0.94)` press. Press feedback feels softer as a result. If that's intentional fine — but flagging in case it's a side effect of consolidating timings.

**3. `runtime.languageTransition` can linger if user is not on General tab when lang changes**
It gets set in `applyChanges` regardless of active tab, but only consumed in `buildLanguageRow`. If a lang change broadcast lands while the user is on Agents/Animation Map and they later return to General, they'll see a phantom slide for a transition that's already settled. Cheap fix: clear it on tab switch, or only set it when General is active.

**4. `animMap.patchInPlace` uses closure-captured `themeId`**
Currently safe — `theme` and `themeOverrides` are always committed in separate broadcasts in this codebase — but if a future command ever commits both atomically, the patch loop will re-read `themeOverrides` with the old `themeId` and paint the wrong state. A defensive `if ("theme" in changes) return false` would future-proof this.

**5. Missing `keeps Animation Map broadcasts in place` regression test**
The Agents tab has a `getContentRenderCount()`-based test pinning the no-rebuild guarantee. Animation Map only has source-grep coverage, so a regression that silently re-routes through full rerender wouldn't be caught. A parallel test would close that gap.

**6. CSS source-order dependency**
`.language-segmented { display: grid }` overrides `.segmented { display: inline-flex }` purely by source order (both single-class selectors). If someone reshuffles the stylesheet the pill positioning silently breaks. A short comment near the `.language-segmented` block would help.

---

Happy to take another pass once the P1 is resolved. The rest are quality-of-life — your call on which of those to bundle vs defer.

PR #256, opened 2026-05-08
feat: route Go to Terminal to Superset / iTerm2 / Ghostty

tatsuyanakanogaroinc:
## Summary
- Fix the AskUserQuestion bubble's **Go to Terminal** button — it was sending `decide("deny")` instead of `decide("deny-and-focus")`, so the focus path that Plan Review uses never ran. Users could not jump back to the originating terminal session.
- Add a Mac-only specialized focus chain (in `src/focus.js`) that runs before the generic `set frontmost of process` script:
  - **Superset.app** — looks up `workspaces.id` for the current cwd in `~/.superset*/local.db` (sqlite3 readonly) and opens `superset[-name]://workspace/<id>`. The renderer routes to that workspace and `focusMainWindow()` brings the app forward. Custom instances (`.superset-<name>`) are auto-detected.
  - **iTerm2** — picks the tty from the source/agent process tree (`ps -o tty=`) and selects the matching session via AppleScript `tty of session ends with …`.
  - **Ghostty** — no public IPC, so match the window title against the cwd basename via System Events + `AXRaise`. Requires a shell that sets OSC 2 titles.
- Each step short-circuits on success; if none match, falls through to the existing generic `set frontmost of process whose unix id is N` AppleScript — Terminal.app / Cursor / VS Code users see no change.

## Why
The generic frontmost script could only bring the parent app forward, not navigate to the specific session that triggered the permission request. The most painful case was Superset.app, where every worktree shares one Electron window — focus snapped to whichever workspace happened to be on top. iTerm2 and Ghostty have similar problems with multi-tab/window setups.

## Test plan
- [x] Existing test suite passes (`node --test test/bubble-elicitation-stepper.test.js` — 8/8)
- [x] **Superset URL scheme verified at runtime** — `open superset://workspace/<id>` correctly navigates the running Superset instance to the matching workspace; invalid IDs surface the app's 404 page, confirming the deep-link handler is reached
- [ ] iTerm2 tty match — implementation only, **not verified at runtime** (iTerm2 was not running locally)
- [ ] Ghostty title match — implementation only, **not verified at runtime**; depends on the user's shell sending OSC 2 titles
- [ ] Full-app smoke test (start the app, click Go to Terminal in an AskUserQuestion bubble) — **not run**

## Notes
- Generic frontmost fallback is intact, so platforms outside the specialized chain keep their prior behavior.
- LaunchServices may have multiple Superset claims (e.g., a leftover mounted DMG). The implementation calls `/usr/bin/open <url>` without `-a`, relying on the default handler. If a stale DMG outranks the running app, the deep link routes to the inactive copy. Workaround: `hdiutil detach /Volumes/Superset*` to clear duplicate registrations.
- `sqlite3` CLI is used (no native module dependency added). DB is opened read-only with `-readonly` to stay safe with the WAL connection Superset holds.
- Superset deep-link path discovered via static analysis of `app.asar` on 2026-05-08 — internal scheme, no stability guarantee. Pinned in a comment.

### DIFF
--- a/src/bubble.html
+++ b/src/bubble.html
@@ -973,7 +973,7 @@ function renderElicitationTerminalFallback() {
   btn.addEventListener("click", () => {
     btn.textContent = "...";
     disableAll();
-    window.bubbleAPI.decide("deny");
+    window.bubbleAPI.decide("deny-and-focus");
   });
   suggestionsContainer.appendChild(btn);
 }
--- a/src/focus.js
+++ b/src/focus.js
@@ -11,6 +11,229 @@ const isMac = process.platform === "darwin";
 const isWin = process.platform === "win32";
 const isLinux = process.platform === "linux";
 
+// Shared between the closure-scoped focus pipeline below and the
+// module-level helpers further down.
+const MAC_FOCUS_TIMEOUT_MS = 1500;
+
+// ── Mac-only: specialized focus paths ────────────────────────────────────────
+// These cover terminal hosts the generic `set frontmost of process` script
+// cannot reach correctly:
+//   • Superset.app — multi-workspace Electron host. Generic frontmost only
+//     activates the app, not the specific worktree workspace. Resolved via the
+//     reverse-engineered `superset://workspace/<id>` URL scheme (observed
+//     2026-05-08). `open` is given `-b com.superset.desktop` so a stale DMG
+//     copy registered with LaunchServices cannot intercept the deep link.
+//   • iTerm2 — multiple sessions per window. Match by tty so the right tab
+//     is selected even when many shells share one window.
+//   • Ghostty — public AppleScript dictionary; pick the terminal whose
+//     `working directory` matches cwd and call `focus` on it (raises window
+//     and selects the matching terminal in one step).
+//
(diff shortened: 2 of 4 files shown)


tatsuyanakanogaroinc:
Closing in favor of a fresh PR rebased on the latest upstream/main.

Two upstream changes landed after this branch diverged from main and made the diff incompatible:

- #216 already adds iTerm2 tab-level focus switching, which overlaps with `tryFocusITerm` in this PR.
- `9203315` (refactor: extract bubble renderer assets) splits `src/bubble.html` into `src/bubble-renderer.js` + `src/bubble.css`, so the elicitation-bubble Go-to-Terminal fix needs to land in `bubble-renderer.js` instead.

Will reopen as a smaller PR that:
- Applies the elicitation `decide("deny-and-focus")` fix in `bubble-renderer.js` (the new home).
- Adds only the Superset and Ghostty focus paths to `src/focus.js`, leaving the upstream iTerm2 implementation untouched.
- Keeps the unit tests for the Superset / pid-helper functions.

Sorry for the noise — should have rebased before opening.

PR #257, opened 2026-05-08
feat(focus): focus the originating workspace from Go to Terminal

tatsuyanakanogaroinc:
Replaces #256 — rebased on the latest upstream/main so it doesn't conflict with #216 (iTerm2 tab-level focus) or `9203315` (bubble renderer extract).

## Summary

Make the AskUserQuestion bubble's **Go to Terminal** button actually focus the originating workspace. Two complementary changes:

1. **`src/bubble-renderer.js`** — `renderElicitationTerminalFallback` now sends `decide("deny-and-focus")` instead of `decide("deny")`. The focus path Plan Review already uses now runs for elicitation too; previously the click denied the request but never focused anything.

2. **`src/focus.js`** — two specialized Mac focus helpers fire alongside the existing generic frontmost / iTerm2 paths in `executeMacFocusRequest`:
   - **`scheduleSupersetFocus(cwd)`** — looks up `workspaces.id` for the current cwd in `~/.superset*/local.db` (sqlite3 readonly), then opens `superset[-name]://workspace/<id>` via `open -b com.superset.desktop`. Pinning the bundle id ensures a stale Superset DMG that LaunchServices still claims for the scheme cannot intercept the deep link. Custom instances (`.superset-<name>`) are auto-detected.
   - **`scheduleGhosttyFocus(sourcePid, cwd)`** — when the source process command is `ghostty`, asks Ghostty's scripting dictionary to `focus` the terminal whose `working directory` matches cwd (and its realpath, so symlinks resolve). `focus` selects the surface and raises its window in one step.

Both helpers are best-effort fire-and-forget; they silently no-op when their target app isn't running or the cwd isn't tracked, so non-Superset / non-Ghostty users see no change. iTerm2's tab-level focus stays exactly as added in #216 — this PR doesn't touch it.

## Why
Generic `set frontmost of process whose unix id is N` only activates the parent app, never navigates to the specific workspace/terminal that triggered the request. The most painful case was Superset.app, where every worktree shares one Electron window — focus snapped to whichever workspace was visible. iTerm2 was already addressed by #216; Superset and Ghostty were the remaining gaps.

## Test plan
- [x] `node --test test/focus-superset.test.js test/bubble-elicitation-stepper.test.js` passes (17/17). New `test/focus-superset.test.js` covers `findSupersetDataDirs`, `supersetSchemeForDir`, and `querySupersetWorkspaceId` (sqlite path skips when `sqlite3` CLI is unavailable).
- [x] Verified at runtime: `open -b com.superset.desktop superset://workspace/<id>` navigates the running Superset instance to the matching workspace; invalid IDs surface its built-in 404, confirming the deep-link handler is reached.
- [ ] Ghostty `working directory` AppleScript path — implementation only, **not verified at runtime** (depends on Ghostty 1.3.0+ scripting dictionary).
- [ ] Full end-to-end Go-to-Terminal smoke test (start the app, click Go to Terminal in an AskUserQuestion bubble) — **not run**.

## Notes
- Pure helpers used by the Superset lookup live at module scope and are exposed via `module.exports.__test` (separate from the factory's instance-level `__test` namespace).
- Ghostty AppleScript reads `(contents of cwdLiteral)` to dereference the `repeat with x in list` loop variable — without that the `whose` clause can compare a list reference to a string and miss.
- Internal Superset URL scheme reverse-engineered from `app.asar` on 2026-05-08; pinned with a comment in case the scheme drifts.

### DIFF
--- a/src/bubble-renderer.js
+++ b/src/bubble-renderer.js
@@ -591,7 +591,7 @@ function renderElicitationTerminalFallback() {
   btn.addEventListener("click", () => {
     btn.textContent = "...";
     disableAll();
-    window.bubbleAPI.decide("deny");
+    window.bubbleAPI.decide("deny-and-focus");
   });
   suggestionsContainer.appendChild(btn);
 }
--- a/src/bubble-renderer.js
+++ b/src/bubble-renderer.js
@@ -591,7 +591,11 @@ function renderElicitationTerminalFallback() {
   btn.addEventListener("click", () => {
     btn.textContent = "...";
     disableAll();
-    window.bubbleAPI.decide("deny-and-focus");
+    // Use plain "deny" — permission.js's elicitation branch already calls
+    // focusTerminalForSession after sending the Elicitation deny response.
+    // "deny-and-focus" hides the bubble without writing to perm.res, which
+    // would leave the blocking Elicitation HTTP hook open.
+    window.bubbleAPI.decide("deny");
   });
   suggestionsContainer.appendChild(btn);
 }
--- a/src/focus.js
+++ b/src/focus.js
@@ -1,14 +1,74 @@
 // src/focus.js — Terminal focus system (PowerShell persistent process + macOS osascript)
 // Extracted from main.js L1030-1335
 
+const fs = require("fs");
 const http = require("http");
+const os = require("os");
 const crypto = require("crypto");
 const path = require("path");
-const { execFile, spawn } = require("child_process");
+const { execFile, execFileSync, spawn } = require("child_process");
 
 const isMac = process.platform === "darwin";
 const isWin = process.platform === "win32";
 const isLinux = process.platform === "linux";
+
+// ── Mac-only: Superset workspace deep-link helpers ──────────────────────────
+// Superset.app is a multi-workspace Electron host. The generic
+// `set frontmost of process whose unix id is N` script only activates the app
+// — it can't switch the visible workspace to the worktree the request came
+// from. Use the reverse-engineered `superset://workspace/<id>` URL scheme to
+// navigate (observed 2026-05-08; internal scheme, no stability guarantee).
+//
+// `open` is given `-b com.superset.desktop` so a stale Superset DMG that
+// LaunchServices still claims for the scheme cannot intercept the deep link.
+//
+// These helpers are at module scope so the unit tests can exercise them
+// without standing up the full Electron context.
+
+const SUPERSET_BUNDLE_ID = "com.superset.desktop";
+
+function findSupersetDataDirs(homeDir) {
+  // Superset by default lives in ~/.superset, but custom instances use
+  // `~/.superset-<name>` and the matching URL scheme `superset-<name>://`.
+  const home = homeDir || os.homedir();
+  try {
+    return fs.readdirSync(home, { withFileTypes: true })
+      .filter((e) => e.isDirectory() && e.name.startsWith(".superset"))
+      .map((e) => path.join(home, e.name))
+      .filter((dir) => {
+        try { return fs.existsSync(path.join(dir, "local.db")); }
+        catch { return false; }
+      });
+  } catch { return []; }
+}
+
+function supersetSchemeForDir(dir) {
+  const base = path.basename(dir);
+  if (base === ".superset") return "superset";
+  if (base.startsWith(".superset-")) return `superset-${base.slice(".superset-".length)}`;
+  return null;
+}
+
+function querySupersetWorkspaceId(dbPath, cwd) {
+  if (!cwd) return null;
+  const candidates = [cwd];
+  try {
+    const real = fs.realpathSync(cwd);
+    if (real && real !== cwd) candidates.push(real);
+  } catch {}
+  for (const candidate of candidates) {
+    const escaped = candidate.replace(/'/g, "''");
+    const sql = `SELECT ws.id FROM workspaces ws JOIN worktrees w ON w.id = ws.worktree_id WHERE w.path = '${escaped}' ORDER BY COALESCE(ws.last_opened_at, 0) DESC LIMIT 1;`;
+    try {
+      const out = execFileSync("sqlite3", ["-readonly", dbPath, sql], {
+        encoding: "utf8",
+        timeout: 1500,
+      }).trim();
+      if (out) return out;
+    } catch {}
+  }
+  return null;
+}
 
 module.exports = function initFocus(ctx) {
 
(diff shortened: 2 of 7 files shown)


rullerzhou-afk (dismissed):
Thanks for working on this. The Superset/Ghostty focus support looks useful, but I found a couple of issues that I think should be fixed before merge.

## Blocking: AskUserQuestion "Go to Terminal" no longer responds to the Elicitation hook

`src/bubble-renderer.js` changes the elicitation terminal fallback from `decide("deny")` to `decide("deny-and-focus")`.

For elicitation, those two paths are not equivalent. The previous `deny` path reaches `resolvePermissionEntry()`, then the `permEntry.isElicitation` branch in `src/permission.js`, which sends the Elicitation deny response and already focuses the terminal:

```js
sendPermissionResponse(res, "deny", message, "Elicitation");
ctx.focusTerminalForSession(permEntry.sessionId);
```

The new `deny-and-focus` path hides the bubble and focuses the terminal, but does not write to `perm.res`. That leaves the blocking Elicitation HTTP hook open, so Claude Code does not get the deny response it needs in order to fall back to the native terminal prompt.

Could you either revert this button back to `decide("deny")`, or add an elicitation-specific `deny-and-focus` branch that still sends the Elicitation deny response before focusing? I think reverting is probably the smaller and safer fix, since the existing deny path already focuses.

It would also be good to add a regression test for this path: handling the elicitation terminal fallback should send the Elicitation deny response and call `focusTerminalForSession()`.

## Blocking: test suite is currently red

`npm test` currently fails in:

```text
test/focus-iterm-tab.test.js
should skip iTerm2 tab focus when pidChain is empty

AssertionError: Should not attempt iTerm detection with empty pidChain
2 !== 0
```

This seems to be because the new Superset and Ghostty probes both call `ps -o comm=` even when `pidChain` is empty. The iTerm behavior itself may still be fine, but the existing test is using `ps -o comm=` as a proxy for "no iTerm detection", and that proxy is no longer accurate.

Could you update the test to assert the iTerm-specific behavior instead, such as no `tty=` lookup or no iTerm2 AppleScript when `pidChain` is empty?

## Follow-up suggestions

The Superset helper currently uses `execFileSync("sqlite3", ...)` from the Electron main process focus path. Since this is behind the Superset comm gate, the practical impact is limited, but it can still block the UI event loop for Superset users. An async `execFile` path or a short-lived cwd-to-workspace cache would be safer.

The new Superset/Ghostty focus paths could also use direct tests around the comm gate, the `open -b com.superset.desktop ...` call, and the Ghostty AppleScript dispatch. The current Superset test covers the pure helper functions, which is useful, but it does not exercise the actual focus scheduling path.

tatsuyanakanogaroinc:
Thanks for the careful review! Pushed two follow-up commits addressing all of the requested changes.

**Blocking fixes** (commit `ebee286`)

- `src/bubble-renderer.js` — reverted the Go-to-Terminal handler back to `decide("deny")`. As you pointed out, the elicitation branch in `resolvePermissionEntry` already calls `sendPermissionResponse(res, "deny", message, "Elicitation")` and `ctx.focusTerminalForSession(permEntry.sessionId)`, so the previous behavior was correct on both legs and `deny-and-focus` was actively breaking the HTTP hook. Added `test/permission-elicitation-deny-focus.test.js` as a regression guard so this contract can't silently break again.
- `test/focus-iterm-tab.test.js` — switched the empty-`pidChain` assertion from "no `ps -o comm=` calls" to iTerm-specific behavior (no `ps -o tty=` lookup, no iTerm2 AppleScript dispatch). The Superset / Ghostty probes legitimately call `ps -o comm=` for their own host detection now, so the old proxy was no longer accurate.

**Follow-up improvements** (commits `ebee286` + `54eba08`)

- `querySupersetWorkspaceId` is now async (callback form). `scheduleSupersetFocus` chains the per-data-dir attempts via callbacks, so the Electron main event loop stays unblocked even on cold-disk sqlite reads.
- New `test/focus-mac-extras.test.js` directly exercises `scheduleSupersetFocus` (comm gate, sqlite lookup, `open -b com.superset.desktop`, cwd guard) and `scheduleGhosttyFocus` (comm gate, AppleScript dispatch). Mirrors the `loadFocusWithMock` pattern in `focus-iterm-tab.test.js`.
- Made the new Superset tests hermetic: temp `HOME` with a planted `.superset/local.db`, so the suite no longer reads the host's real Superset install (or fails on machines without one).

PR-relevant tests pass 30/30. The unrelated `updater` / `claude-version-detection` / `kiro-install` failures in `npm test` predate this PR and weren't touched here.

Remaining risk: end-to-end Go-to-Terminal smoke tests against running Ghostty / Superset / iTerm2 are still manual.

tatsuyanakanogaroinc:
Thanks for the careful review! Pushed two follow-up commits addressing all of the requested changes.

**Blocking fixes** (commit `ebee286`)

- `src/bubble-renderer.js` — reverted the Go-to-Terminal handler back to `decide("deny")`. As you pointed out, the elicitation branch in `resolvePermissionEntry` already calls `sendPermissionResponse(res, "deny", message, "Elicitation")` and `ctx.focusTerminalForSession(permEntry.sessionId)`, so the previous behavior was correct on both legs and `deny-and-focus` was actively breaking the HTTP hook. Added `test/permission-elicitation-deny-focus.test.js` as a regression guard so this contract can't silently break again.
- `test/focus-iterm-tab.test.js` — switched the empty-`pidChain` assertion from "no `ps -o comm=` calls" to iTerm-specific behavior (no `ps -o tty=` lookup, no iTerm2 AppleScript dispatch). The Superset / Ghostty probes legitimately call `ps -o comm=` for their own host detection now, so the old proxy was no longer accurate.

**Follow-up improvements** (commits `ebee286` + `54eba08`)

- `querySupersetWorkspaceId` is now async (callback form). `scheduleSupersetFocus` chains the per-data-dir attempts via callbacks, so the Electron main event loop stays unblocked even on cold-disk sqlite reads.
- New `test/focus-mac-extras.test.js` directly exercises `scheduleSupersetFocus` (comm gate, sqlite lookup, `open -b com.superset.desktop`, cwd guard) and `scheduleGhosttyFocus` (comm gate, AppleScript dispatch). Mirrors the `loadFocusWithMock` pattern in `focus-iterm-tab.test.js`.
- Made the new Superset tests hermetic: temp `HOME` with a planted `.superset/local.db`, so the suite no longer reads the host's real Superset install (or fails on machines without one).

PR-relevant tests pass 30/30. The unrelated `updater` / `claude-version-detection` / `kiro-install` failures in `npm test` predate this PR and weren't touched here.

Remaining risk: end-to-end Go-to-Terminal smoke tests against running Ghostty / Superset / iTerm2 are still manual.

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
株式会社GARO

　代表取締役社長
　中野 達哉／Tatsuya Nakano
　〒150-0043
　　東京都渋谷区道玄坂1丁目10番8号渋谷道玄坂東急ビル2F−C
　TEL:03-6453-1872
　Mobile:090-9306-0714

　◇web
　　https://www.garoinc.jp
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
2026年5月10日 9:26 +0900、rullerzhou-afk/clawd-on-desk ***@***.***> のメール:

PR #258, opened 2026-05-08
fix: stabilize animation override controls

YOIMIYA66:
## Summary
- Stabilizes `Settings -> 动画 / 音效替换` so timing slider edits update the visible slider, reset affordance, and changed badge without tearing down the whole content pane.
- Makes the wide-hitbox toggle behave as a scoped control instead of a card-wide click target, with pending state, rollback, and theme-default reset semantics.
- Preserves the existing theme override schema and existing animation/sound replacement command shapes; this only tightens how the renderer and theme runtime apply those overrides.

## Problem context
- Timing edits previously caused visible flicker because successful saves and `themeOverrides` broadcasts could rebuild the Animation Overrides tab while the user was still interacting with a mounted slider.
- The first timing edit could leave the “changed” badge and reset button stale because the active card status was only refreshed after a full render.
- Returning a fade timing to the theme default could still look overridden because the command stored an override equal to the base theme value instead of clearing it.
- Wide-hitbox toggles used heavier theme reload behavior and could leave reset affordances visible when the effective value had returned to the theme default.
- During review I also found a tier-card regression: `clawd` working/juggling tier rows could fail while building transition metadata, leaving the page with only the header/actions visible.

## What changed
- Timing slider state
  - Tracks mounted timing controls and pending edits by card/field.
  - Applies matching `themeOverrides` broadcasts in place when they acknowledge the pending timing edit.
  - Updates reset buttons and changed badges from mounted controls so the first edit and return-to-default cases are reflected immediately.
  - Clears transition overrides when the requested timing matches the theme default.

- Wide-hitbox controls
  - Converts the wide-hitbox row into a scoped checkbox plus a dedicated “reset to theme” chip, instead of making the whole card area clickable.
  - Records pending wide-hitbox edits and syncs the mounted checkbox, reset chip, and summary badge without forcing a content rebuild.
  - Sends `enabled: null` when the effective value matches the theme default so stale hitbox override entries are removed.
  - Makes the slot reset path clear hitbox-only overrides as well, so an enabled reset button always actually restores the slot.

- Theme/runtime metadata
  - Exposes base transition and base wide-hitbox metadata from the theme loader so Settings can compare effective values against theme defaults.
  - Adds a lightweight active-theme hitbox refresh path for active theme hitbox changes, avoiding a full theme reload and pet fade cycle for a checkbox-only change.
  - Keeps existing full theme reload behavior for changes that still need it, such as animation file replacement and broader theme/variant changes.

- Loading and blank-page resilience
  - Keeps the previous Animation Overrides data visible while a theme override refresh is pending.
  - Adds visible loading copy for the initial Animation Overrides fetch.
  - Fixes tier-card transition metadata generation so `clawd` tier rows build normally instead of failing the data payload.

## Behavior after this PR
- Dragging or entering fade/return timing values keeps the current row mounted and updates the visible slider/input without flashing back to the previous value.
- The first timing edit immediately enables the reset affordance and changed badge; returning the timing to the theme default clears them.
- Wide-hitbox toggles no longer trigger the pet’s full fade/reload path and no longer rely on clicking the whole row.
- Resetting a slot clears animation/timing changes and hitbox-only changes consistently.
- Animation file replacement, sound replacement, asset picker behavior, theme switching, and prefs schema remain unchanged.

## Validation
- `node --test test/settings-renderer-browser-env.test.js test/settings-actions-theme-overrides.test.js test/settings-actions.test.js test/settings-animation-overrides-main.test.js test/theme-runtime.test.js`
- `node --test --test-name-pattern "clears hitbox-only|Animation Overrides reset affordances|wide hitbox" test/settings-renderer-browser-env.test.js`
- `git diff --check`
- `npm test` passed with `2113` passing, `0` failing, and `2` skipped tests.
- Direct runtime smoke check against the built-in `clawd` theme successfully built `31` animation cards and `2` sound items for the Animation Overrides payload.

## Platform note
- Automated verification ran locally on Windows with Node's built-in test runner.
- Manual Electron QA is still recommended for the final feel of slider dragging and wide-hitbox toggle interaction in the real Settings window.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -207,6 +207,8 @@ const _settingsController = createSettingsController({
     // Theme runtime is wired after theme-loader.init(); keep these closures
     // lazy so settings actions never capture a pre-init runtime reference.
     activateTheme: (id, variantId, overrideMap) => themeRuntime.activateTheme(id, variantId, overrideMap),
+    refreshActiveThemeHitboxOverrides: (id, overrideMap) =>
+      themeRuntime.refreshActiveThemeHitboxOverrides(id, overrideMap),
     getThemeInfo: (id) => themeRuntime.getThemeInfo(id),
     removeThemeDir: (id) => themeRuntime.removeThemeDir(id),
     globalShortcut,
--- a/src/settings-actions-theme-overrides.js
+++ b/src/settings-actions-theme-overrides.js
@@ -130,6 +130,11 @@ function normalizeTransitionPayload(transition) {
   return Object.keys(out).length > 0 ? out : null;
 }
 
+function transitionMatchesThemeDefault(transition, defaultTransition) {
+  if (!transition || !defaultTransition) return false;
+  return transition.in === defaultTransition.in && transition.out === defaultTransition.out;
+}
+
 const _validateThemeOverrideThemeId = requireString("setThemeOverrideDisabled.themeId");
 function setThemeOverrideDisabled(payload, deps) {
   if (!payload || typeof payload !== "object") {
@@ -213,6 +218,13 @@ function setAnimationOverride(payload, deps) {
   if (touchesTransition && payload.transition !== null && !normalizeTransitionPayload(payload.transition)) {
     return { status: "error", message: "setAnimationOverride.transition must contain finite non-negative in/out values" };
   }
+  const normalizedTransition = touchesTransition && payload.transition !== null
+    ? normalizeTransitionPayload(payload.transition)
+    : null;
+  const normalizedTransitionDefault = normalizeTransitionPayload(payload.transitionThemeDefault);
+  const transitionMatchesDefault = !!(touchesTransition
+    && normalizedTransition
+    && transitionMatchesThemeDefault(normalizedTransition, normalizedTransitionDefault));
   if (touchesAutoReturn && payload.autoReturnMs !== null) {
     if (typeof payload.autoReturnMs !== "number" || !Number.isFinite(payload.autoReturnMs)) {
       return { status: "error", message: "setAnimationOverride.autoReturnMs must be null or a finite number" };
@@ -260,8 +272,8 @@ function setAnimationOverride(payload, deps) {
       }
     }
     if (touchesTransition) {
-      if (payload.transition === null) delete nextEntry.transition;
-      else nextEntry.transition = normalizeTransitionPayload(payload.transition);
+      if (payload.transition === null || transitionMatchesDefault) delete nextEntry.transition;
+      else nextEntry.transition = normalizedTransition;
     }
     if (Object.keys(nextEntry).length > 0) nextStates[stateKey] = nextEntry;
     else delete nextStates[stateKey];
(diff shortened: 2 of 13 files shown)


rullerzhou-afk:
Follow-up review on the current head (`224be94f`): I rechecked the Settings override paths and think there are three real follow-ups worth patching before merge.

1. Returning fade timing to the theme default still falls out of the in-place patch path.

`setAnimationOverride` now correctly deletes a transition override when the submitted transition matches `transitionThemeDefault`, but the renderer-side predictor still applies the pending transition as a concrete override in `applyPendingEditToThemeOverrides()`.

Path:
- User changes `{ in: 160, out: 150 }` back to `{ in: 150, out: 150 }`.
- Main action prunes `states.thinking.transition` because it matches the theme default.
- `pendingEditsMatchThemeOverrideBroadcast()` predicts `states.thinking.transition = { in: 150, out: 150 }`.
- Expected snapshot no longer equals the broadcast snapshot, so `patchInPlace()` returns false and `settings-ui-core` falls back to `requestRender({ content: true })`.

That preserves correctness, but it reintroduces the content rebuild/flicker this PR is trying to avoid. The fix should make the pending prediction default-aware. One detail: deleting `entry.transition` after creating the slot entry is not enough by itself, because `pruneThemeOverrideMap()` only prunes top-level groups, not empty slot entries like `states.thinking = {}`. Either avoid creating the entry for default-matching transitions, or also prune the empty slot entry.

2. Slot reset can leave a stale `hitbox.wide[B]` override when the animation file is reset from B back to A.

`resetAnimationOverrideCard()` runs the animation reset first, then fetches the new card and checks `currentCard.wideHitboxOverridden`. If the hitbox override belonged to the replaced file (`B.svg`), the fetched card is already back on the base file (`A.svg`), so `wideHitboxOverridden` is false and the hitbox cleanup is skipped.

Concrete shape:
- `states.thinking = { file: "B.svg" }`
- `hitbox.wide["B.svg"] = true`
- Reset clears the state file override, card current file becomes `A.svg`
- Cleanup checks `A.svg`, does not see the `B.svg` entry, and `hitbox.wide["B.svg"]` remains in prefs

This is also hard for the user to clean from the slot UI afterwards, because the stale entry is keyed by a file no longer shown on the card. The reset path should capture the pre-reset `card.currentFile` and clear `setWideHitboxOverride({ file: capturedFile, enabled: null })` if that file had a wide-hitbox override.

3. Timing + wide-hitbox pending together can still force a fallback rebuild.

`setAnimationOverride` and `setWideHitboxOverride` are separate command locks, and the wide-hitbox checkbox is not disabled while a timing command is pending. If both pending maps have entries and the first settings broadcast only contains one of the two commits, `pendingEditsMatchThemeOverrideBroadcast()` predicts both changes at once, does not match the broadcast snapshot, and falls back to the full content render.

This is lower impact than the first two because it needs a short overlapping interaction window, but it is the same prediction-model fragility. The minimal fix is probably to make the predictor tolerant of partial acknowledgements, or block the second control while a same-card timing edit is pending.

Non-blocking note: `refreshActiveThemeHitboxOverrides()` also bypasses the full reload protocol while `reloadInProgress` may still be true. That looks like a lifecycle invariant hole if a wide-hitbox edit lands during an active fade reload. I would not block this PR on it if we guard the common paths above, but it is worth either falling back to `activateTheme()` during reload or returning a clear no-op/error in that window.

PR #259, opened 2026-05-09
feat: support cmux workspace switching on session focus

### LINKED ISSUE #248: Feature request: Support cmux terminal workspace/session switching
## Feature request

Please add support for using [cmux](https://cmux.com) as a controllable terminal target.

## Current behavior

When cmux is used as the terminal, clicking a session from Clawd does not switch to the corresponding cmux workspace/session.

## Expected behavior

When a Clawd session is clicked, Clawd should switch/focus the matching cmux workspace/session.

## Possible implementation

cmux appears to be controllable via its CLI, so Clawd could detect cmux sessions and use the cmux CLI to switch/focus the target workspace/session.

## Why this is useful

Clawd already tracks multiple agent sessions, and cmux is useful for managing multiple terminal workspaces. Supporting cmux would make session navigation work properly for users who run Claude Code/Codex/etc. inside cmux.

nmsn:
## Summary

- Detect cmux terminal multiplexer sessions via process chain (`ps -o comm=`)
- Match TTY to workspace by reading cmux session file (`~/Library/Application Support/cmux/session-*.json`)
- Focus the correct cmux workspace via AppleScript (`tell application "cmux" to focus terminal`)
- Add 400ms delay before AppleScript to avoid racing with System Events (same pattern as iTerm2)
- Extract shared `findFirstValidTty` helper to deduplicate TTY parsing between iTerm2 and cmux
- Extract shared `loadFocusWithMock` test helper

## How it works

When `executeMacFocusRequest` fires, `scheduleCmuxWorkspaceSwitch` runs alongside the existing fast-paths:

1. `ps -o comm=` on pidChain to check if cmux process exists
2. `ps -o pid=,tty=` to find shell TTY
3. Read cmux session JSON, match TTY → workspace index
4. AppleScript: `tell application "cmux" to focus terminal 1 of tab N`

## Test plan

- [x] `node --test test/focus-cmux.test.js` — 6 new tests pass
- [x] `node --test test/focus-iterm-tab.test.js` — no regression (5 tests)
- [x] `node --test test/focus-windows.test.js` — no regression (8 tests)
- [ ] Manual: open cmux workspace, run claude-code inside, click pet → switches to correct workspace

Closes #248


rullerzhou-afk:
Thanks for taking this on. cmux focus support would be really useful, and this PR is a good start. I think the current targeting logic still needs one more pass before merge, mainly because it can select the wrong terminal in common cmux layouts.

A few concerns:

1. In `src/focus.js:397-404`, the code matches the TTY to a workspace, but then focuses `terminal 1 of tab N`. If the agent is running in a split workspace and belongs to the second or third terminal panel, this will switch to the right workspace but leave keyboard focus in the wrong terminal. Since the cmux session snapshot includes each panel `id`, and cmux exposes terminals by stable AppleScript id, it would be safer to capture the matched panel id and focus it directly:

   ```applescript
   tell application "cmux" to focus terminal id "<panel-id>"
   ```

2. The implementation only reads `windows[0]`, then sends AppleScript to `front window`. With multiple cmux windows, this can miss sessions in another window or switch the matching tab index in the wrong front window. Focusing by stable terminal id should also address this, because cmux's terminal `focus` command brings the owning window forward, selects the workspace, and focuses the panel.

3. The implementation depends on the autosaved `session-*.json`. cmux autosaves periodically, so newly created workspaces/splits/shells may not be present in the file yet. It may be worth using a live source instead, such as `CMUX_WORKSPACE_ID` / `CMUX_SURFACE_ID`, cmux socket/CLI data, or a fallback AppleScript lookup. If snapshot-based matching stays, we should at least make the limitation explicit and handle misses clearly.

Some smaller follow-ups that would help make this robust:
- Use async `fs` APIs instead of `readdirSync` / `readFileSync` in the Electron main process.
- Check the snapshot schema version before trusting the JSON shape.
- Add tests for multi-panel workspaces, multi-window snapshots, version mismatch/malformed JSON, and multiple session files. The multi-panel case is the important one: it should assert that the AppleScript targets the matched panel id, not just `tab N`.

I'd recommend holding merge until the focus command targets the matched terminal id instead of `terminal 1 of tab N of front window`.

rullerzhou-afk:
Added maintainer fix-up commit 24c5770 for the remaining cmux focus blockers: pass --workspace to cmux focus-panel so Clawd does not depend on CMUX_WORKSPACE_ID in its own environment; build the macOS cmux bundle binary path with POSIX separators so darwin-simulated tests pass on Windows; scan all candidate session-*.json files instead of stopping at the first autosave file. Verification: node --test test/focus-cmux.test.js test/focus-iterm-tab.test.js test/focus-mac-extras.test.js; npm test.

rullerzhou-afk:
I did a second pass on the latest head (24c5770) and cross-checked the remaining concerns against cmux docs/source. No further code blockers from my side.

What looks confirmed:
- cmux documents `focus-panel --panel <id|ref|index> [--workspace <id|ref|index>]`, and `--workspace` defaults to `CMUX_WORKSPACE_ID`, so passing the matched workspace explicitly is the right shape for Clawd.
- `select-workspace --workspace <id>` is documented.
- `com.cmuxterm.app` is the current bundle id per cmux changelog.
- cmux session persistence currently stores `windows[].tabManager.workspaces[].panels[]`, with `id` and `ttyName` on each panel, so the snapshot matcher aligns with current schema.

I think this is ready pending one real macOS/cmux smoke test, since this environment cannot verify cmux focus behavior end-to-end. Suggested smoke:
1. Run `/Applications/cmux.app/Contents/Resources/bin/cmux focus-panel --help` and confirm `--workspace` / `--panel` are accepted.
2. Run `/Applications/cmux.app/Contents/Resources/bin/cmux select-workspace --help` and confirm `--workspace` is accepted.
3. Open a cmux workspace with split panes, run a Clawd-tracked agent in one pane, then click the session in Clawd and confirm keyboard focus lands in the correct pane, not just the correct workspace.

If that passes, I am comfortable merging this PR. Remaining cleanup items such as more precise cmux log reason names, a comment on the 400ms delay, or a more canonical Spotlight query can be handled later and should not block merge.

PR #267, opened 2026-05-11
feat(copilot-hook): forward Copilot CLI session name to Clawd

XiangZheng2002:
## Problem
 
 Renaming a Copilot CLI session via `/rename` updates
 `~/.copilot/session-state/<sid>/workspace.yaml` (`name:` field), but
 `hooks/copilot-hook.js` never read it — so `body.session_title` was never
 set and Clawd's HUD / Dashboard kept falling back to `path.basename(cwd)`.
 Multiple Copilot sessions in the same repo were indistinguishable.
 
 Codex (`readCodexThreadName`) and Claude Code
 (`extractSessionTitleFromTranscript`) already forward their renamed
 titles; Copilot CLI was just unimplemented.
 
 ## Fix
 
 The hook now reads `name:` from `workspace.yaml` on every event and sends
 it as `session_title`. State plumbing already supports this field
 (`displayTitle` priority: alias > sessionTitle > cwd basename > id), so
 **no `src/` changes needed**.
 
 Notes:
 - Lazy refresh, same as Claude Code: takes effect on the next hook event
   after rename.
 - Auto-generated names (`user_named: false`) used as-is (matches Codex
   thread-name behavior).
 - User-set Dashboard alias still wins.
 - Path traversal blocked (`^[A-Za-z0-9._-]+$` on session id),
   `workspace.yaml` read capped at 16 KB, zero new dependencies.
 - Hook refactored into pure helpers + `require.main === module` guard
   for testability — same shape as `hooks/clawd-hook.js`.
 
 ## Tests & docs
 
 - `test/copilot-hook.test.js`: 26 new unit tests covering YAML parsing
   edge cases (quoted/unquoted/inline-comment/CRLF/indented), file
   lookup, path-traversal rejection, and `buildStateBody` branches.
   `npm test` clean modulo 10 pre-existing failures unrelated to this
   change (verified via `git stash`).
 - `docs/guides/copilot-setup.md`: adds a "Session rename" section.

### DIFF
--- a/docs/guides/copilot-setup.md
+++ b/docs/guides/copilot-setup.md
@@ -14,3 +14,7 @@ Create `~/.copilot/hooks/hooks.json` with the following content. Replace `/path/
   }
 }
 ```
+
+## Session rename
+
+Copilot CLI stores the current session name in `~/.copilot/session-state/<sessionId>/workspace.yaml` (`name:` field). The hook reads that file on every event and forwards the name to Clawd as the session title, so `/rename` inside Copilot CLI propagates to the Session HUD and Dashboard on the next hook event (next user message, tool run, etc.). Auto-generated names (`user_named: false`) are used as-is, matching Codex thread-name behavior.
--- a/hooks/copilot-hook.js
+++ b/hooks/copilot-hook.js
@@ -3,8 +3,103 @@
 // Usage: node copilot-hook.js <event_name>
 // Reads stdin JSON from Copilot CLI for sessionId (camelCase)
 
+const fs = require("fs");
+const os = require("os");
+const path = require("path");
 const { postStateToRunningServer } = require("./server-config");
 const { createPidResolver, readStdinJson, getPlatformConfig } = require("./shared-process");
+
+const SESSION_TITLE_CONTROL_RE = /[\u0000-\u001F\u007F-\u009F]+/g;
+const SESSION_TITLE_MAX = 80;
+const WORKSPACE_YAML_MAX_BYTES = 16384; // 16 KB — workspace.yaml is tiny
+
+function normalizeTitle(value) {
+  if (typeof value !== "string") return null;
+  const collapsed = value
+    .replace(SESSION_TITLE_CONTROL_RE, " ")
+    .replace(/\s+/g, " ")
+    .trim();
+  if (!collapsed) return null;
+  return collapsed.length > SESSION_TITLE_MAX
+    ? `${collapsed.slice(0, SESSION_TITLE_MAX - 1)}\u2026`
+    : collapsed;
+}
+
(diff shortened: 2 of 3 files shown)


rullerzhou-afk (changes requested):
Thanks for the PR. The overall approach makes sense: Copilot already provides `sessionId`, and forwarding the `workspace.yaml` `name:` as `session_title` fits the existing Clawd state plumbing.

I found one issue that should be fixed before merge:

`readCopilotSessionTitle()` says path traversal is blocked by the session id regex, but `.` and `..` still pass `/^[A-Za-z0-9._-]+$/`.

For example:

- `sessionId="."` resolves to `~/.copilot/session-state/workspace.yaml`
- `sessionId=".."` resolves to `~/.copilot/workspace.yaml`

So the hook can read outside the intended per-session directory. The impact is limited because the filename is fixed to `workspace.yaml` and the value is normalized before being sent as `session_title`, but the path boundary is still not enforced.

Suggested fix:

- explicitly reject pure dot path segments, at least `.` and `..`
- ideally also resolve the final path and assert it stays under `~/.copilot/session-state/`

Please also add tests for this. The current traversal tests use `"../../../etc"` and `"a/b"`, but those are rejected only because they contain `/`, so they do not cover this bypass. A good regression test would create `~/.copilot/workspace.yaml` with `name: leaked` and assert `readCopilotSessionTitle("..") === null`.

I do not think the smaller YAML parsing edge cases need to block this PR.

XiangZheng2002:
Thanks for catching this — fix pushed in 8f96a14 (force-pushed to the
 same branch). Both layers you suggested are in:
 
 1. **Pure-dot rejection** — `if (/^\.+$/.test(trimmed)) return null;`
    catches `.`, `..`, `...` etc. that pass the charset gate.
 2. **Post-resolve containment check** —
    `path.resolve(baseDir + sid)` must start with `baseDir + path.sep`,
    so even if a future change loosens the regex or the dot rule, the
    read still cannot escape `~/.copilot/session-state/`.
 
 Three regression tests added (29 total now):
 
 - `readCopilotSessionTitle("..") === null` with `~/.copilot/workspace.yaml`
   containing `name: leaked` planted as bait — your exact scenario.
 - Symmetric `"."` test with `~/.copilot/session-state/workspace.yaml`
   as bait.
 - Pure-dot inputs of length ≥ 3.
 
 I also renamed the existing `"rejects path-traversal-style sessionIds"`
 test to `"rejects sessionIds containing path separators or empty/null"`
 since you're right that its inputs only cover the separator case, not
 the dot-segment bypass — the name was misleading.
 
 Per your note, I haven't touched the YAML parsing edge cases.

rullerzhou-afk (approved):
Re-reviewed latest 8f96a14; requested changes addressed.

PR #268, opened 2026-05-11
fix: replace wmic with PowerShell Get-CimInstance

Hwasowl:
## Problem

  Windows 11 24H2 (build 26200+) removed `wmic` from the default install.

  The hook's parent-process-chain walk called `wmic` on every PID iteration.
  On affected builds the very first call throws `ENOENT`, the `catch { break }` fires immediately,
  and `pidChain` comes back empty. `stablePid` then resolves to the hook's own short-lived PID — 
  silently breaking Dashboard / HUD session focus and every PID-dependent feature.

  ---

  ## Solution

  Replace both `wmic` call sites with `Get-CimInstance Win32_Process` — the same WMI class, different query interface.

  ### `hooks/shared-process.js` — per-iteration parent-process walk

  ```powershell
  Get-CimInstance Win32_Process -Filter "ProcessId=<pid>" |
    Select-Object Name, ParentProcessId |
    ConvertTo-Json -Compress
  ```

  No process found → empty stdout → `if (!trimmed) break` (same guard as before).
  Added an explicit `if (!info) break` after `JSON.parse` to handle the edge case where
  `ConvertTo-Json` outputs the literal string `"null"` for a null object.

  ### `hooks/clawd-hook.js` — headless (`-p`/`--print`) detection

  ```powershell
  (Get-CimInstance Win32_Process -Filter "ProcessId=<pid>" -ErrorAction SilentlyContinue).CommandLine
  ```

  Also swaps `execSync` → `execFileSync` (argument array, no shell) on both the Windows and macOS/Linux branches.

  ---

  ## Compatibility

  | Platform | Before | After |
  |---|---|---|
  | Win 11 24H2+ (build 26200+) | `pidChain=[]`, session focus broken | Restored |
  | Win 11 23H2 and below | wmic worked | PowerShell path, identical output |
  | macOS / Linux | `ps` path unchanged | No logic change |

  Timeouts are unchanged (main query 1500 ms, cmdline check 500 ms). PowerShell cold-start on affected builds is ~270 ms.

  ---

  ## Tests

  ### `test/clawd-hook.test.js` — 8 new cases (`agentPid` and headless detection)

  Previously this branch was never exercised. `execFileSync` is monkey-patched per test:

  - `agentPid` present → `agent_pid` + `claude_pid` set
  - `agentPid` absent → both fields absent
  - cmdline ends with `-p` → `headless: true`
  - cmdline has `-p <arg>` → `headless: true`
  - cmdline has `--print` → `headless: true`
  - cmdline has `--port` (prefix match) → no `headless`
  - cmdline is empty → no `headless`
  - `execFileSync` throws → `agent_pid` still set, no `headless`

  ### `test/shared-process.test.js` — 8 new cases (win32-only, skipped on other platforms)

  The PowerShell + JSON parsing path in `createPidResolver()` had no unit coverage:

  - pidChain populated by walking PS JSON output
  - walk breaks immediately on empty PS output (process not found)
  - walk breaks cleanly on `ConvertTo-Json` `"null"` output
  - `stablePid` set to terminal PID when terminal process is found
  - `detectedEditor` resolved from process name (`code.exe` → `"code"`)
  - walk stops at system boundary (`explorer.exe`), callCount asserted
  - `agentPid` detected via `agentNameSet` name match
  - `agentPid` detected via `agentCmdlineCheck` on `node.exe` cmdline probe

  **66 / 66 pass.**

  ---

  ## Note

  Per-iteration `powershell.exe` spawn is acknowledged. A single process-table snapshot to amortize startup cost is intentionally deferred to a follow-up commit — this PR is a minimal, safe drop-in replacement.

  ## References

  - [wmic deprecation — Windows deprecated features](https://learn.microsoft.com/en-us/windows/whats-new/deprecated-features)
  - [Working with WMI — PowerShell 101](https://learn.microsoft.com/en-us/powershell/scripting/learn/ps101/07-working-with-wmi)

### DIFF
--- a/hooks/clawd-hook.js
+++ b/hooks/clawd-hook.js
@@ -171,16 +171,21 @@ function buildStateBody(event, payload, resolve) {
     if (agentPid) {
       body.agent_pid = agentPid;
       body.claude_pid = agentPid; // backward compat with older Clawd versions
-      // Check if claude process is running in non-interactive (-p/--print) mode
+      // Check if claude process is running in non-interactive (-p/--print) mode.
+      // Windows 11 24H2+ removed wmic, so fall back to PowerShell Get-CimInstance.
       try {
-        const { execSync } = require("child_process");
+        const { execFileSync } = require("child_process");
         const isWin = process.platform === "win32";
         const cmdOut = isWin
-          ? execSync(
-              `wmic process where "ProcessId=${agentPid}" get CommandLine /format:csv`,
+          ? execFileSync(
+              "powershell.exe",
+              [
+                "-NoProfile", "-NonInteractive", "-Command",
+                `(Get-CimInstance Win32_Process -Filter "ProcessId=${agentPid}" -ErrorAction SilentlyContinue).CommandLine`,
+              ],
               { encoding: "utf8", timeout: 500, windowsHide: true }
             )
-          : execSync(`ps -o command= -p ${agentPid}`, { encoding: "utf8", timeout: 500 });
+          : execFileSync("ps", ["-o", "command=", "-p", String(agentPid)], { encoding: "utf8", timeout: 500 });
         if (/\s(-p|--print)(\s|$)/.test(cmdOut)) body.headless = true;
       } catch {}
     }
--- a/hooks/clawd-hook.js
+++ b/hooks/clawd-hook.js
@@ -165,29 +165,17 @@ function buildStateBody(event, payload, resolve) {
   if (process.env.CLAWD_REMOTE) {
     body.host = readHostPrefix();
   } else {
-    const { stablePid, agentPid, detectedEditor, pidChain } = resolve();
+    const { stablePid, agentPid, agentCommandLine, detectedEditor, pidChain } = resolve();
     body.source_pid = stablePid;
     if (detectedEditor) body.editor = detectedEditor;
     if (agentPid) {
       body.agent_pid = agentPid;
       body.claude_pid = agentPid; // backward compat with older Clawd versions
-      // Check if claude process is running in non-interactive (-p/--print) mode.
-      // Windows 11 24H2+ removed wmic, so fall back to PowerShell Get-CimInstance.
-      try {
-        const { execFileSync } = require("child_process");
-        const isWin = process.platform === "win32";
-        const cmdOut = isWin
-          ? execFileSync(
-              "powershell.exe",
-              [
-                "-NoProfile", "-NonInteractive", "-Command",
-                `(Get-CimInstance Win32_Process -Filter "ProcessId=${agentPid}" -ErrorAction SilentlyContinue).CommandLine`,
-              ],
-              { encoding: "utf8", timeout: 500, windowsHide: true }
-            )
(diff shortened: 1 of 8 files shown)


rullerzhou-afk (changes requested):
Request changes.

This PR is incomplete for the Windows 11 24H2 `wmic` removal issue. It replaces `wmic` in `hooks/shared-process.js` and `hooks/clawd-hook.js`, but three real runtime paths still depend on `wmic`:

- `src/state.js:1089` startup recovery still shells out to `wmic`.
- `hooks/pi-extension.ts:309` Pi process-chain lookup still calls `execFileSync("wmic", ...)`.
- `hooks/opencode-plugin/index.mjs:203` opencode terminal PID lookup still calls `wmic`.

These should be fixed before merge, otherwise the release will only restore Claude/shared hook PID detection while Pi/opencode/startup recovery remain broken on the same Windows builds.

Also strongly recommend collapsing the new Windows PID walk in `hooks/shared-process.js` into one PowerShell invocation. The current implementation cold-starts `powershell.exe` once per ancestor process; on Windows this made the real resolver tests take multiple seconds. Since these hooks run as fresh Node processes and tool hooks fire frequently, that latency will be visible.

Hwasowl:
Thanks for the review — all four spots are addressed:

1. **`src/state.js`** — startup recovery probe rewritten to `execFile('powershell.exe', …)` running `Get-CimInstance Win32_Process | Where-Object { … }`. Same boolean output contract as before.
2. **`hooks/pi-extension.ts`** — Pi extension's per-PID `wmic` walk replaced by a single `getWindowsProcessSnapshot()` + in-memory Map lookup. `ProcessInfo` shape unchanged on both Windows and Unix.
3. **`hooks/opencode-plugin/index.mjs`** — opencode plugin's per-PID `wmic` walk replaced with the same snapshot pattern via `execFileSync('powershell.exe', [...])`.
4. **`hooks/shared-process.js`** — Windows PID walk collapsed into one PowerShell invocation. Measured `createPidResolver()` test block dropped 10.7 s → 1.8 s. Added a regression test that asserts `execFileSync` is called exactly once per resolve regardless of chain depth.

While auditing for residual spawns I also caught one more: `hooks/clawd-hook.js` was spawning a separate PS to fetch the agent's CommandLine for the `-p`/`--print` headless check, even though the resolver had already seen the same string. The resolver now exposes `agentCommandLine` and `clawd-hook` reuses it — PS spawns per hook fire on Windows dropped from 3 → 1.

PR body updated with the full breakdown. Rebased onto latest `main` (Hermes integration merged) and resolved the `state.js` conflict by keeping the PS migration and folding `hermes.exe` into the `$names` list. 113 / 113 affected tests pass; no remaining `wmic` invocations in the tree.

rullerzhou-afk (approved):
Re-reviewed after the latest updates. The originally requested Windows wmic replacements in shared hooks, startup recovery, Pi, and opencode are addressed, and the full test suite passes locally. The remaining Hermes wmic path came from a later mainline change, so I'll handle that in a follow-up PR rather than blocking this one.

Issue #107, opened 2026-04-14
Feature: support settings.local.json or allow disabling hook auto-restore


VIKI623:
## Problem

Clawd on Desk watches `~/.claude/settings.json` and auto-restores hooks when they are removed (`watchSettingsForHookLoss()` in `src/server.js`). This creates friction for users who sync `settings.json` across machines (e.g. local Mac → remote GPU servers via scp).

Remote servers don't have Clawd on Desk installed, so the hooks reference non-existent paths (`/Applications/Clawd on Desk.app/...`) and cause errors. Currently the only workaround is to filter hooks out every time before syncing:

```bash
jq 'del(.hooks)' ~/.claude/settings.json | ssh remote "cat > ~/.claude/settings.json"
```

## Proposed solutions (any of the following would help)

1. **Write hooks to `settings.local.json` instead of `settings.json`**
   Claude Code supports `~/.claude/settings.local.json` with higher priority than `settings.json`. This file is intended for machine-specific config and is not meant to be synced. Writing hooks there would cleanly separate Clawd's local-only hooks from portable settings.

2. **Add a config toggle to disable hook auto-restore**
   Something like `"autoRestoreHooks": false` in Clawd's settings, so users can opt out of the file-watch restoration behavior while keeping hooks registered on first launch.

3. **Add an uninstall-hooks command**
   A `npm run uninstall:hooks` or UI button to cleanly remove hooks and stop watching, for users who want to temporarily or permanently disconnect.

## Context

- Claude Code `settings.local.json` docs: settings defined there override `settings.json` and are meant for local/machine-specific use.
- The current hook registration in `hooks/install.js` and restoration in `src/server.js` lines 225-252 are hardcoded to target `settings.json` with no user-facing override.

rullerzhou-afk:
Thanks for the report. I re-checked the Claude Code docs and looked through the current implementation, and I think points 2 and 3 are the most practical direction here.

On point 1 first: `settings.local.json` is real, but the documented file is project-local `.claude/settings.local.json`, not `~/.claude/settings.local.json`. So this is not impossible, but it would be a larger design change than just switching the write target. It would move Clawd from a global hook model ("install once, works everywhere") to a project-scoped hook model ("hooks live per repo/project"). That also means this is not really "you must set hooks manually every session," but it likely does become "you need to handle hooks per project instead of once at the user level."

For points 2 and 3, I think they should be treated as one combined design, not two separate knobs.

[...]

PR #286, opened 2026-05-13
fix: preserve third-party hooks on suspicious settings.json shrink

taehwanis:
## Summary

Skip auto-resync when `~/.claude/settings.json` shrinks suspiciously between
watcher ticks, preserving third-party hooks (maestro, secret-guard, etc.) when
external tools (e.g. `claude mcp remove --scope user`, CC-Switch) edit the file.

## Problem

When an external CLI modifies `settings.json`, the modification may leave the
file with the Clawd hook marker missing alongside other top-level keys (`env`,
`permissions`, `enabledPlugins`) and third-party hooks. The current watcher
reads this state, sees no `clawd-hook.js` marker, and triggers
`syncClawdHooks()`. `registerHooks()` then writes back a config containing only
Clawd's hooks — silently dropping user-authored and third-party hooks.

Observed in production: 33 hooks / 18 top-level keys reduced to 14 Clawd-only
hooks / 2 keys after `claude mcp remove promptbin && claude mcp add` ran in a
separate terminal while Clawd's watcher was active.

This violates the contract from #107 ("preserves third-party hooks") and the
AGENTS.md "Constraints" guarantee:

> 注册 Claude Code hook 时只能追加，不能覆盖用户已有 hook 数组
>
> (Translation: "Registering Claude Code hooks must only append, not overwrite
> the user's existing hook array")

## Fix

Maintain an in-memory snapshot of the last trusted `{ keyCount, hookCount }`.
On each watcher tick that would trigger resync, compare against the snapshot.
If the new state shows a suspicious shrink (configurable via `ctx`: default
50% hook loss or 3+ top-level keys lost), log a warning and skip the resync.

The baseline is seeded from settings.json at `start()` time when the file is
healthy, so the very first watcher event after Clawd boots can be compared
against a real snapshot (closes the cold-start race surfaced in review).

The user can recover by inspecting via Settings > Doctor, restarting Clawd,
or explicitly toggling `manageClaudeHooksAutomatically` OFF→ON in Settings
(which deliberately bypasses this guard — see Tradeoffs).

### Threshold rationale

The default 50% hook loss / 3 keys lost was calibrated against the observed
production case (33 → 14 hooks = 0.42 ratio, 18 → 2 keys = 16 lost) and
intentionally kept conservative so that a user pruning ~half of their hooks
on purpose is not blocked. Operators who want stricter or laxer behavior
can tune `ctx.suspiciousShrinkRatio` and `ctx.suspiciousKeyLossThreshold`
when constructing the watcher.

## Tradeoffs

- **Cold-start race (resolved per review feedback)** — see Fix section.
- **User intent vs external tool indistinguishable** — both trigger sync
  suspension. User can re-trigger sync manually if intentional.
- **Toggle paths bypass this guard** — when users explicitly toggle
  `manageClaudeHooksAutomatically` OFF→ON, re-enable claude-code, or otherwise
  invoke `syncClaudeHooksNow()` through the Settings UI, the resync runs
  outside the watcher callback and the snapshot check does not apply. This is
  intentional: any sync path that originates from a Settings UI action is
  treated as explicit user consent. External CLI side-effects (where the
  user's intent was managing something else, like MCP servers) are the only
  defended path.

## Tests

New cases added to existing `test/claude-settings-watcher.test.js`:

- skips auto-resync on suspicious shrink (the production race scenario)
- allows resync on first tick when `start()` found no healthy baseline to
  seed (fresh install / settings unhealthy at boot)
- allows resync on minor shrink (user removes 1 hook intentionally)
- threshold tuning via `ctx.suspiciousShrinkRatio` /
  `ctx.suspiciousKeyLossThreshold`
- optional `ctx.notifySuspiciousShrink(prev, curr)` callback fires when the
  guard trips (currently used in tests; reserved for future Doctor IPC
  integration)
- seeds the baseline on `start()` so the very first watcher event can trip
  the guard (cold-start race coverage)
- keeps blocking resync while the shrunk state persists across watcher events
  (stuck-state is intentional; recovery via Settings UI toggle is out of
  scope for this PR)
- treats non-object JSON payloads (`null`, `[]`) as unparseable and allows
  resync (malformed payloads are not treated as attacks)

Existing tests pass unchanged.

## Reproduction

Manually reproduced the destructive path both with and without this patch
on a development machine to confirm the fix targets the exact bug.

**Setup.** Healthy `~/.claude/settings.json` (18406 bytes, 17 top-level
keys, 14 `clawd-hook.js` markers, plus third-party hook commands from
`maestro-audit` and `secret-guard`). File backed up beforehand and
restored byte-wise after each run.

**Without this patch (HEAD before this commit):**

Wrote `{ "skipDangerousModePermissionPrompt": true }` (47 bytes) to
settings.json to simulate an external CLI minimize. ~1s after the write
the Electron console logged:

```
Clawd: hooks missing from settings.json — re-registering
Clawd: synced hooks (added 16, updated 0, removed 0)
```

Final state: 4716 bytes, 2 top-level keys, 14 `clawd-hook.js` markers,
**no `maestro-audit` or `secret-guard` entries — third-party hooks lost.**

**With this patch (same machine, same simulation):**

Same minimize write. ~1s after the write the console logged:

```
Clawd: settings.json shrank suspiciously — skipping auto-resync to preserve third-party hooks
```

Final state after a 10-second hold (much longer than the 1s watcher
debounce window): 47 bytes unchanged — no re-registration, third-party
entries preserved (then restored from backup as part of cleanup).

## Related

- contributes to #107 (third-party hooks preservation)
- AGENTS.md "Constraints" — append-only guarantee
- aa371b8 (Claude hook management controls) introduced safe ENOENT handling
  in `unregisterHooks`; this PR extends the same preservation guarantee to the
  watcher path

### DIFF
--- a/src/claude-settings-watcher.js
+++ b/src/claude-settings-watcher.js
@@ -178,6 +178,23 @@ function createClaudeSettingsWatcher(ctx = {}) {
     if (settingsWatcher) return false;
     const settingsDir = getClaudeSettingsDir();
     const settingsPath = getClaudeSettingsPath();
+    // Seed the trusted baseline from the current settings.json before the watcher starts,
+    // so the very first watcher event after Clawd boots (e.g. an external CLI minimize
+    // landing right after syncClawdHooks() ran) can be compared against a real snapshot
+    // instead of null. Wrapped in its own try/catch so a missing or unreadable file
+    // (fresh install, permission error) cannot prevent the watcher from starting.
+    // The settingsNeedClaudeHookResync guard inside this block also prevents seeding
+    // from a polluted state if an external CLI raced ahead of this read.
+    try {
+      const seedRaw = fsApi.readFileSync(settingsPath, "utf-8");
+      const seedPort = typeof ctx.getHookServerPort === "function" ? ctx.getHookServerPort() : null;
+      const seedExpectedPermissionUrl = buildPermissionUrl(seedPort);
+      if (!settingsNeedClaudeHookResync(seedRaw, seedExpectedPermissionUrl)) {
+        lastTrustedSnapshot = takeSnapshot(seedRaw);
+      }
+    } catch (err) {
+      console.warn("Clawd: could not seed settings baseline:", err.message);
+    }
     try {
       settingsWatcher = fsApi.watch(settingsDir, (_event, filename) => {
         if (filename && filename !== SETTINGS_FILENAME) return;
--- a/src/claude-settings-watcher.js
+++ b/src/claude-settings-watcher.js
@@ -56,6 +56,80 @@ function settingsNeedClaudeHookResync(rawSettings, expectedPermissionUrl) {
   return !hasManagedCommandHook || !hasManagedPermissionHook;
 }
 
+function countCommandHooksInEntries(entries) {
+  if (!Array.isArray(entries)) return 0;
+  let count = 0;
+  for (const entry of entries) {
+    if (!entry || typeof entry !== "object") continue;
+    if (entry.type === "http") continue;
+    if (typeof entry.command === "string") count += 1;
+    if (!Array.isArray(entry.hooks)) continue;
+    for (const hook of entry.hooks) {
+      if (!hook || typeof hook !== "object") continue;
+      if (hook.type === "http") continue;
+      if (typeof hook.command === "string") count += 1;
+    }
+  }
+  return count;
+}
+
+/**
+ * Count total command hooks across every event in the hooks object.
+ * Handles both nested format (entry.hooks[].command) and flat format (entry.command).
+ * HTTP hooks (type: "http") are excluded because they cannot encode the marker.
+ * @param {object|null|undefined} hooks
+ * @returns {number}
+ */
+function countAllHooks(hooks) {
+  if (!hooks || typeof hooks !== "object") return 0;
+  let total = 0;
+  for (const entries of Object.values(hooks)) {
+    total += countCommandHooksInEntries(entries);
+  }
+  return total;
+}
+
+/**
+ * Capture a snapshot of top-level key count and command hook count from settings.json text.
+ * Returns null when the payload is not a parseable JSON object so callers can skip comparisons.
+ * @param {string} raw
+ * @returns {{keyCount: number, hookCount: number}|null}
+ */
+function takeSnapshot(raw) {
+  if (typeof raw !== "string" || !raw.trim()) return null;
+  let parsed;
+  try {
+    parsed = JSON.parse(raw);
+  } catch {
+    return null;
+  }
+  if (!parsed || typeof parsed !== "object" || Array.isArray(parsed)) return null;
+  return {
+    keyCount: Object.keys(parsed).length,
+    hookCount: countAllHooks(parsed.hooks),
+  };
+}
+
+/**
+ * Decide whether the shrink between two snapshots looks suspicious.
+ * Two independent OR triggers — hook drop ratio reaches shrinkRatio, or top-level key drop reaches keyLossThreshold.
+ * Either snapshot missing returns false because insufficient history cannot prove an attack.
+ * @param {{keyCount: number, hookCount: number}|null} prev
+ * @param {{keyCount: number, hookCount: number}|null} curr
+ * @param {number} shrinkRatio
+ * @param {number} keyLossThreshold
+ * @returns {boolean}
+ */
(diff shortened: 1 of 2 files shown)


rullerzhou-afk:
Thanks for working on this. This is a subtle race, and the snapshot guard is a good direction. I found one startup-order issue that I think should be fixed before merge.

### 1. The guard is not seeded on normal startup

The new guard depends on `lastTrustedSnapshot`, but that value starts as `null` and is only refreshed after the watcher observes a healthy `settings.json` change.

In the normal startup path, Claude hook sync runs before the watcher starts:

- `syncClawdHooks()`
- `startClaudeSettingsWatcher()`

So the just-synced healthy settings file is never captured as the trusted baseline. If the first watcher event after startup is an external CLI shrink, `isSuspiciousShrink(null, currentSnapshot, ...)` returns `false`, and the watcher still calls `syncClawdHooks()`.

That means the production sequence this PR is trying to protect can still happen after a normal Clawd start.

I think `start()` should seed the baseline from the current settings file, ideally only when the file is already healthy:

```js
const raw = fsApi.readFileSync(settingsPath, "utf-8");
const port = typeof ctx.getHookServerPort === "function" ? ctx.getHookServerPort() : null;
const expectedPermissionUrl = buildPermissionUrl(port);

if (!settingsNeedClaudeHookResync(raw, expectedPermissionUrl)) {
  lastTrustedSnapshot = takeSnapshot(raw);
}
```

Could you also add a test for this exact cold-start case?

Suggested test shape:

1. `watcher.start()`
2. do not emit a healthy settings change first
3. set settings to the minimized/shrunk payload
4. emit `settings.json`
5. assert that `syncClawdHooks()` is not called once the baseline has been seeded from startup

### 2. Guarded shrink currently has no recovery path

When the guard trips, it returns immediately and keeps the old `lastTrustedSnapshot`. If the shrunk file is stable and later emits more fs events, the watcher will keep comparing against the old healthy snapshot and keep skipping resync.

That is safer than overwriting third-party hooks, but it can leave users in a silent stuck state where Clawd hooks are missing and auto-management no longer repairs them.

For this PR, a minimal improvement could be to make the behavior explicit and test it. Longer term, this probably needs a user-visible Doctor/Settings signal or an "accept current file as baseline" / "restore from known-good backup" flow.

### Minor edge case

`takeSnapshot()` returns `null` for values like `null` or `[]`. In those cases `settingsNeedClaudeHookResync()` can still return `true`, but the shrink guard cannot compare snapshots and will allow resync. It may be worth adding tests for malformed or non-object JSON payloads so the intended behavior is documented.

Overall, I think the approach is promising, but I would prefer to get the startup baseline seeding and cold-start test in before merging, otherwise the main race can still slip through.

taehwanis:
💡 Review reflection

Thanks for the thorough review — all three points are addressed in this push.

- **Cold-start seeding**: `start()` now seeds `lastTrustedSnapshot` from the
  current settings.json when the file is healthy, before installing the
  fs.watch listener. The seed step is wrapped in its own try/catch so a
  missing or unreadable file (fresh install / permission error) cannot block
  the watcher from starting. New test `seeds the baseline on start so the
  very first watcher event can trip the guard` covers the race.

- **Stuck state**: behavior is locked in by the new test `keeps blocking
  resync while the shrunk state persists across watcher events` (3
  consecutive shrink events, guard trips on every event). Happy to follow up
  with a separate PR along the lines you suggested (Doctor signal or
  explicit "accept as baseline" reset) if you'd like to scope it.

- **Malformed JSON edge case**: new test `treats non-object JSON payloads
  (null, array) as unparseable and allows resync` documents the intended
  fail-open behavior for `null` / `[]` payloads — these come from partial
  writes or unrelated tools, and treating them as attacks would prevent
  Clawd from recovering its own hooks.

Also addressed two minor items found while preparing this update:
- seed `try { ... } catch { ... }` now logs `console.warn("Clawd: could not
  seed settings baseline:", err.message)` so silent failures are no longer
  invisible at debug time.
- inline comment near the seed block clarifies that the
  `settingsNeedClaudeHookResync` guard inside the same block prevents seeding
  from a polluted state if an external CLI raced ahead of the read.

PR body Tradeoff 1 (cold-start race) updated to reflect the resolution.

Verified: 13 watcher unit tests pass; full test suite 2399/2403 (4
pre-existing skips). Also re-ran a manual local reproduction with real
fs.watch (BEFORE without the seed → guard bypassed; AFTER with the seed →
guard trips).

Let me know if anything else needs attention.

PR #295, opened 2026-05-15
feat: enhance Session HUD with state chips, pulse animation, prompt titles, and wider layout

xuwei0912:
## Summary

- Show granular state label chip per session (thinking/working/juggling/error/compacting/waiting/worktree/done/interrupted) with color-coded styling and dark mode support
- Animate running dot with pulse to distinguish active sessions at a glance
- Use `UserPromptSubmit` prompt text as session title fallback when no explicit `/rename` title is set
- Expand max visible rows from 3 to 5 and widen HUD from 240px to 320px
- Add missing i18n keys (`sessionError`/`Sweeping`/`Notification`/`Carrying`) for all 4 languages (en/zh/ko/ja)

## Test plan

- [ ] Start multiple Claude Code sessions and verify each row shows the correct state chip and color
- [ ] Verify running dot pulses for active sessions
- [ ] Send a prompt without `/rename` — HUD title should update to the prompt text
- [ ] Use `/rename` — verify explicit name takes priority over prompt text
- [ ] Verify up to 5 sessions expand before folding
- [ ] Check dark mode styling for all chip variants


rullerzhou-afk (commented):
Thanks for the PR. The HUD chip/pulse direction looks useful, but I found a couple of issues that should be fixed before merging.

1. `src/i18n.js` is missing the new runtime keys in the `zh-TW` locale.

The PR adds `sessionError`, `sessionSweeping`, `sessionNotification`, and `sessionCarrying` to `en/zh/ko/ja`, but `SUPPORTED_LANGS` also includes `zh-TW`. This makes `test/i18n.test.js` fail on `keeps all locale keysets aligned with English`, and at runtime the HUD would show raw keys like `sessionError` for Traditional Chinese users.

Suggested fix: add the 4 keys to the `zh-TW` block as well.

2. `test/session-hud.test.js` needs to be updated for the new HUD constants.

The PR changes `HUD_WIDTH` from 240 to 320 and `HUD_MAX_EXPANDED_ROWS` from 3 to 5, but some tests still encode the old assumptions. I see 4 failures around flipped positioning, anchor positioning, and folding behavior.

Suggested fix: update those assertions to derive expected values from `constants.HUD_WIDTH` / `constants.HUD_MAX_EXPANDED_ROWS`, and use `cap + 2` sessions in the folding test so it still exercises the folded row.

3. The new prompt-title fallback in `hooks/clawd-hook.js` should probably skip slash commands.

Right now, if `UserPromptSubmit` has no explicit/transcript title, the first prompt line is written into `session_title`. For normal prompts that makes sense, but slash commands like `/rename Better title`, `/clear`, or `/help` can briefly or persistently become the sticky session title until a later event supplies a real title.

Suggested fix: skip fallback when the first prompt line starts with `/`, and add clawd-hook tests for:
- prompt first-line fallback
- explicit/transcript title still wins
- slash commands do not become session titles

Optional: `chip-juggling` and `chip-notification` currently use the same colors in both light and dark mode, so the two states are only distinguishable by text. It may be worth giving notification a separate color.

PR #296, opened 2026-05-15
fix: detect Node.js installed via nvm/fnm/asdf/mise on remote SSH host

jryxxx:
## Summary

The remote Node.js probe script currently fails to find Node when it's installed via version managers (nvm, fnm, asdf, mise) because non-interactive SSH sessions don't source shell rc files. The existing fallback — login-shell probing with sentinel-based stdout parsing — can also fail in minimal environments.

## Fix

Added `probe_version_managers()` to the probe script, which directly sources version-manager init scripts (`nvm.sh`, `fnm env`, `asdf.sh`, `mise activate`) and then retries `command -v node`. This runs after `probe_login_shells()` and before the hardcoded candidate-path glob scan.

Before this fix, users with nvm-managed Node on Ubuntu remotes would see:
```
Remote Node.js not found. Install Node on the remote first.
```

## Test plan

- [x] All 11 `remote-ssh-node.test.js` tests pass
- [x] All 20 `remote-ssh-deploy.test.js` tests pass
- [ ] Manually verified on macOS → Ubuntu remote with nvm-installed Node.js

### DIFF
--- a/src/remote-ssh-node.js
+++ b/src/remote-ssh-node.js
@@ -63,10 +63,49 @@ probe_login_shells() {
   done
 }
 
+probe_version_managers() {
+  # nvm — source the init script so Node lands on PATH even when
+  # shell rc files are not sourced by the non-interactive SSH session.
+  for nvm_sh in "$HOME"/.nvm/nvm.sh /root/.nvm/nvm.sh; do
+    if [ -n "$nvm_sh" ] && [ -f "$nvm_sh" ] && [ -r "$nvm_sh" ]; then
+      NVM_DIR="$(dirname "$nvm_sh")" . "$nvm_sh" 2>/dev/null || true
+      p="$(command -v node 2>/dev/null || true)"
+      emit_node "$p" "nvm:sourced"
+      break
+    fi
+  done
+
+  # fnm — use its own env generator when the binary is on PATH.
+  if command -v fnm >/dev/null 2>&1; then
+    eval "$(fnm env --shell sh 2>/dev/null)" 2>/dev/null || true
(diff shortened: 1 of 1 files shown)


rullerzhou-afk:
Thanks for tackling this. The Node-side resolver change makes sense to me: trying plain PATH first, then login shells, then explicitly sourcing nvm/fnm/asdf/mise before falling back to known candidate paths is a reasonable order for non-interactive SSH sessions.

One blocker I noticed: this PR only updates `src/remote-ssh-node.js`, but `scripts/remote-deploy.sh` has a separate embedded copy of the remote Node probe in its `REMOTE_NODE_PROBE` heredoc. `AGENTS.md` explicitly says `scripts/remote-deploy.sh` and `src/remote-ssh-node.js` must keep the remote Node probe order, candidate paths, version check, and output fields aligned.

As-is, Settings > Remote SSH > Deploy / Repair Hooks gets the new nvm/fnm/asdf/mise sourcing behavior, but the documented CLI fallback:

`bash scripts/remote-deploy.sh user@host`

still uses the old probe and can continue to fail on the same class of remotes this PR is meant to fix. That also leaves users with inconsistent behavior between the Settings deploy path and the CLI deploy path.

Could we update `scripts/remote-deploy.sh` with the same `probe_version_managers` block, in the same position after `probe_login_shells` and before the candidate-path loop? It would also be worth adding a regression test that catches this drift next time. The current `remote-ssh-node` test mostly checks for `.nvm/.fnm/.asdf/.mise` substrings, but those already existed in the old candidate glob list, so it doesn't actually prove the new manager-sourcing block is present or aligned with the shell script.

PR #301, opened 2026-05-16
feat(telegram): native Bot API client with rich approval interactions

jaroenchai-t:
## Summary

  - **Fix 4 bugs** in the existing Telegram approval sidecar/settings UI (status race, config reload, stuck UI states,
  regex too restrictive)
  - **Add native Node.js Telegram Bot API client** that replaces the Go sidecar for richer interactions
  - **Enable elicitation and suggestions** for remote Telegram approval (previously blocked)

  ## Motivation

  The Go sidecar (`cc-connect-clawd`) only supports `{title, detail} → {decision: "allow"|"deny"}` — no custom buttons,
  no option selection, no text input. Additionally, running multiple instances causes `getUpdates` conflicts that
  silently break approval.

  The native client uses Node.js built-in `https` module (zero new dependencies) and calls the Telegram Bot API
  directly, enabling:

  - **Inline keyboard buttons** for Allow/Deny
  - **Permission suggestion buttons** (e.g., "Always allow Bash npm*")
  - **Single-select options** for AskUserQuestion elicitation
  - **Multi-select checkboxes** with visual ☐/☑ toggle state
  - **Text input** via Telegram reply ("Other..." option)
  - **No getUpdates conflict** — only one poller runs

  ## Breaking changes

  None. `nativeMode` defaults to `true` but setting it to `false` preserves the original sidecar behavior. The Go binary
   is still bundled for backward compatibility.

  ## Test plan

  - [x] All 2574 existing tests pass (2 pre-existing workflow failures unrelated)
  - [x] Manual test: Allow/Deny buttons on Telegram — response received correctly
  - [x] Manual test: Multi-select checkboxes — toggle multiple, submit, correct answer returned
  - [x] Manual test: Single-select options — pick one, answer mapped correctly
  - [x] Manual test: No `cc-connect-clawd.exe` process spawned in native mode
  - [x] Manual test: No getUpdates conflict errors in logs
  - [ ] macOS/Linux testing welcome

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -175,6 +175,7 @@ let agentRuntime = null;
 let floatingWindowRuntime = null;
 let codexPetMain = null;
 let telegramApprovalSidecar = null;
+let telegramNativeBot = null;
 let telegramApprovalSyncPromise = Promise.resolve();
 let telegramApprovalConfigSignature = "";
 let telegramApprovalTokenRevision = 0;
@@ -1201,6 +1202,9 @@ function focusLog(msg) {
 }
 
 function getTelegramApprovalClient() {
+  // Prefer native bot (supports rich interactions: options, checkboxes, text)
+  if (telegramNativeBot && telegramNativeBot.isEnabled()) return telegramNativeBot;
+  // Fall back to sidecar (basic allow/deny only)
   if (!telegramApprovalSidecar || typeof telegramApprovalSidecar.getClient !== "function") return null;
   return telegramApprovalSidecar.getClient();
 }
@@ -1255,6 +1259,18 @@ function getTelegramApprovalStatus() {
   const config = getTelegramApprovalPrefs();
   const token = getTelegramApprovalTokenStatus();
   const ready = telegramApprovalSettings.readiness(config, token);
+  // Native bot mode reports as "running" if enabled and configured
+  if (telegramNativeBot && telegramNativeBot.isEnabled()) {
+    return {
+      status: "running",
+      mode: "native",
+      enabled: config.enabled === true,
+      configured: ready.ready === true,
+      reason: ready.reason || "",
+      message: "",
+      tokenStored: token.tokenStored === true,
+    };
+  }
   const sidecarStatus = telegramApprovalSidecar && typeof telegramApprovalSidecar.getStatus === "function"
     ? telegramApprovalSidecar.getStatus()
     : { status: "stopped" };
@@ -1297,6 +1313,26 @@ function startTelegramApprovalSidecar() {
     }
     return false;
   }
+
+  // Native mode: use direct Telegram Bot API (supports rich interactions)
+  if (config.nativeMode !== false) {
+    const chatId = config.targetSessionKey
+      ? config.targetSessionKey.replace(/^telegram:/, "")
+      : "";
+    if (telegramNativeBot) telegramNativeBot.cleanup();
+    const { TelegramNativeBotClient } = require("./telegram-native-bot");
+    telegramNativeBot = new TelegramNativeBotClient({
+      tokenFilePath: paths.tokenEnvFilePath,
+      allowedUserId: config.allowedTgUserId,
+      chatId,
+      ttlMs: 90000,
+      log: telegramApprovalLog,
+    });
+    telegramApprovalLog("info", "native bot mode enabled");
+    telegramApprovalConfigSignature = buildTelegramApprovalSignature(config, paths, token);
+    return true;
+  }
+
   const configWrite = telegramApprovalSettings.writeBridgeConfigFile({
     fs,
     path,
@@ -1320,7 +1356,16 @@ function startTelegramApprovalSidecar() {
     }
     return true;
   }
-  if (telegramApprovalSidecar) return true;
+  if (telegramApprovalSidecar) {
+    // Config changed but old sidecar still assigned — stop it and recreate
+    // with updated config. Without this, a user-ID or session-key change
+    // would be written to TOML but the running sidecar would keep the old
+    // config until app restart.
+    telegramApprovalLog("info", "config signature changed, recreating sidecar");
+    telegramApprovalSidecar.stop().catch(() => {});
+    telegramApprovalSidecar = null;
+    telegramApprovalConfigSignature = "";
+  }
   // The bot token only ever lives at userData/telegram-approval.env on disk.
   // The sidecar reads it from there directly — Clawd's main process must never
   // pipe a token value through process.env or child env, so there is no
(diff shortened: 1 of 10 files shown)


rullerzhou-afk:
Thanks for putting this together. The motivation makes sense: for Telegram alone, a native Node Bot API client is lighter than spawning `cc-connect-clawd`, and the richer interactions here are useful.

After reviewing the direction, I do not think we should merge this PR as-is. The reason is product/architecture direction rather than the effort itself.

Clawd's remote approval path is intended to grow beyond Telegram. Once we support Feishu/Lark, DingTalk, WeCom, QQ, etc., each platform has its own card format, callback model, app review/config, long-connection/webhook behavior, and security rules. We do not want Clawd's Electron main process to become a collection of platform-specific messaging clients.

For that reason, we want `cc-connect` / `cc-connect-clawd` to remain the upstream bridge for remote approval, and Clawd should integrate with that bridge instead of replacing it with a Telegram-only native path. Otherwise we would end up maintaining two primary remote approval implementations: one native Telegram path in Clawd and one multi-platform bridge path through cc-connect.

Some of the bug fixes in this PR are still valuable. If you are willing, smaller follow-up PRs for the sidecar lifecycle/settings hardening pieces would be very welcome. But the native Telegram client replacement changes the direction too much, so I am going to close this PR for now.

I appreciate the work here, and I am sorry for the churn.

rullerzhou-afk:
Hi @jaroenchai-t,

> 抱歉重新打扰你这条 PR。我想认错——我 2026-05-17 关闭这条 PR 时给出的理由 是错的，今天回来纠正一下，并且想问你是否愿意以新的形式继续参与。

I owe you an apology and a context update.

## The closure rationale was wrong

When I closed this PR on 2026-05-17, my stated reason was that we wanted to
keep `cc-connect-clawd` as the upstream bridge for remote approval, because
"Clawd's remote approval path is intended to grow beyond Telegram" toward
Feishu / DingTalk / WeCom / QQ etc.

After several rounds of internal review and reality-checking my own dogfood
usage, that premise turned out to be wrong:

1. I actually only use Telegram. Discord / Feishu were speculative roadmap
   items with no real user pull.
2. My real pain point isn't rich approval at all — it's completion
   notification, status visibility, and reverse prompt injection. The
   approval moment is already adequate as Allow / Deny.
3. With those two corrections, the cost of maintaining the Go sidecar
   fork (29-commit upstream drift, binary distribution, narrow HTTP
   schema) isn't justified.

In other words: **your technical direction in this PR was correct**.
Closing it was a mistake based on a premise I can no longer defend.

## What v0.9.0 will do

Clawd v0.9.0 reverses the v0.8.0 architecture decision. The architecture
direction is exactly what your PR proposed: retire the Go sidecar, embed
a Node-native Telegram bot client directly in Clawd Electron, no fork,
no binary distribution. The `cc-connect-clawd` fork will be archived.

The **product scope** is different though. v0.9.0 focuses on three pain
points that the rich-approval direction in this PR doesn't target:

- **R1**: completion notification — agent finishes → push a "Done" 
  message to Telegram so the user knows when to come back.
- **R2**: status visibility — user sends `/status` in Telegram → see 
  current activity (files touched, commands run, etc.).
- **R3 Stage 1**: remote prompt injection — user sends free text in 
  Telegram → Clawd shows a new bubble class on the desktop offering 
  to copy that text to clipboard for pasting into Claude Code.

Release scope is split: v0.9.0 stable ships the bot client, approval 
parity (so v0.8.0 functionality isn't lost on the new transport), R1 
notification, and lightweight migration. R0 (activity ingestion) / R2 / 
R3 ship behind a feature flag default-off in the same release, and get 
promoted to default-on in v0.9.1 once dogfood validates them.

`permission_suggestions` / `AskUserQuestion` rich approval — the core of
this PR — is deprioritized, because dogfood confirms the approval moment
isn't the bottleneck. The interactive-button work in this PR is not
discarded in spirit; it informs how R3 will render Telegram-side UI.

## Your code's value

After review, I think ~250–350 LOC of this PR is directly reusable as
scaffolding for v0.9.0's `bot-client.js`:

- Telegram Bot API call wrappers (`_apiCall` style)
- long-polling loop with offset advance
- callback / message dispatch
- MarkdownV2 escape
- inline keyboard builders

The ~250 LOC approval state machine (`requestApproval`,
`requestWithSuggestions`, `requestElicitation`, pending-approval TTL,
sidecar-compat dual mode) doesn't fit the new scope and won't be carried
forward.

## Two ways forward — your choice

**Option 1**: I reopen this PR. You update it to drop the approval state
machine and keep just the bot client scaffolding. We work together to
land it as the v0.9.0-alpha starting point. You get co-author credit on
the merge commit and the v0.9.0 release notes call out your contribution.

**Option 2**: If you don't have time or interest, I write `bot-client.js`
from scratch in the v0.9.0-alpha branch, with explicit attribution:

- Commit message:
  `Bot client design and polling/dispatch structure influenced by PR #301 by @jaroenchai-t`
- v0.9.0 release notes acknowledgments section names you
- This PR stays closed but visible, linked from v0.9.0 docs as design input

Either way, **your work doesn't get buried**.

## Timeline

If I don't hear from you by **2026-05-25** (7 days from now), I'll proceed
with Option 2 so the v0.9.0 timeline isn't blocked. This isn't a deadline
on you — it's a commitment from me that I won't leave the question hanging
indefinitely. You can always reach out later and I'll work with you on
where v0.9.1+ goes.

Thanks for the work you put into this PR, and again, I'm sorry for closing
it on a rationale I can no longer stand behind.

— 鹿鹿 / @rullerzhou-afk

rullerzhou-afk:
Hi @jaroenchai-t,

Coming back to close the loop here, and to give credit I owe you.

After this PR went quiet, I rebuilt Telegram remote approval as a native Node bot from scratch — an independent reimplementation, not a copy of this PR's code, to avoid any licensing ambiguity — and it shipped in **v0.9.0**, including the rich-approval and completion-notification capabilities that the native direction unlocked.

What I want to acknowledge clearly: **you were the first to propose moving Telegram approval off the bundled Go `cc-connect-clawd` sidecar onto a native Node bot.** That turned out to be the right direction, and it's the one the project ultimately took. My original closure rationale on 2026-05-17 (keeping the Go sidecar as a multi-platform bridge) was wrong, which is why I reopened this PR to say so.

When v0.9.0 shipped, its release notes credited the native rework end-to-end to me and didn't mention your earlier PR. That was an oversight on my part, and this comment puts the correction on the record: the native Telegram bot in Clawd was **influenced by the direction you proposed in #301**, and I'm grateful for it.

I'm leaving this PR open rather than closing it, so it stays visible as the earliest record of that idea. Thank you for the work, and for the nudge in the right direction.

PR #305, opened 2026-05-19
fix: smooth Clawd SVG diagonal rendering

YOIMIYA66:
## Summary
- Smooths diagonal and rotated edges in the default Clawd SVG animations by switching root SVG rendering hints to `geometricPrecision`.
- Applies the change only to runtime SVG files referenced by `themes/clawd/theme.json`.
- Preserves theme state mappings, animation timing, renderer logic, and non-Clawd themes.

## Problem context
- The previous crisp pixel rendering pass used `shape-rendering="crispEdges"`, which keeps hard pixel stair-steps on rotated or diagonal shapes.
- Clawd has many animated SVGs with rotated limbs, tools, sparks, and mini-mode poses, so root-level hard-edge rendering makes diagonal edges visibly jagged.
- The desired behavior is smoother diagonal rendering for default Clawd without changing the broader theme runtime.

## What changed
- Updated every runtime Clawd SVG referenced by the built-in Clawd theme to use `shape-rendering="geometricPrecision"` on the root `<svg>`.
- Added a focused Node test that derives the runtime SVG set from `themes/clawd/theme.json` and asserts all 35 referenced SVG roots use `geometricPrecision`.
- Kept local per-element rendering overrides intact and did not modify Cloudling, Calico, template, source, old, dist, or About hero assets.

## Behavior after this PR
- Default Clawd SVG animations ask Chromium's SVG renderer to prioritize geometric precision, improving diagonal and rotated edge smoothing.
- Existing animation playback, theme loading, mini mode, hitboxes, sounds, and state selection behavior remain unchanged.
- Non-Clawd themes continue to use their existing asset rendering behavior.

## Validation
- `node --test test\clawd-svg-rendering.test.js test\theme-loader.test.js test\theme-schema.test.js`
- `pwsh -NoLogo -NoProfile -Command "git diff --check"`
- `pwsh -NoLogo -NoProfile -Command "git diff --cached --check"`

## Platform note
- Automated validation ran on Windows in the local Electron project checkout.
- Manual Electron UI QA is still recommended for visually checking diagonal edges in Clawd working, reaction, mini-mode, and idle animations.

### DIFF
--- a/assets/svg/clawd-collapse-sleep.svg
+++ b/assets/svg/clawd-collapse-sleep.svg
@@ -1,4 +1,4 @@
-<svg xmlns="http://www.w3.org/2000/svg" viewBox="-15 -25 45 45" width="500" height="500" shape-rendering="crispEdges">
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="-15 -25 45 45" width="500" height="500" shape-rendering="geometricPrecision">
   <defs>
     <style>
       /* Collapse (0-1s) → Sleep breathing + Zzz (1s onwards, infinite) */
--- a/assets/svg/clawd-error.svg
+++ b/assets/svg/clawd-error.svg
@@ -1,4 +1,4 @@
-<svg xmlns="http://www.w3.org/2000/svg" viewBox="-15 -25 45 45" width="500" height="500">
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="-15 -25 45 45" width="500" height="500" shape-rendering="geometricPrecision">
   <defs>
     <style>
       .body-exhausted {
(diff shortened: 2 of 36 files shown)


rullerzhou-afk:
Thanks for chasing the visual issue here. I agree that the mini enter animation can look too jagged with root `shape-rendering="crispEdges"`, especially because the character is rotated and moves through sub-pixel positions.

After reviewing the affected assets, I think this PR is broader than the bug. Most Clawd runtime SVGs were using the default rendering behavior already, and changing all of them to `geometricPrecision` would turn a localized mini-enter fix into a global rendering policy for the whole default theme.

I'm going to close this one and take a narrower follow-up: only remove the root `crispEdges` hint from the mini enter assets that exhibit the issue, with a focused regression test around those files. Appreciate the careful pass here.

PR #335, opened 2026-05-25
feat: flash tray icon when AI task completes

linnin233:
## Summary

Add configurable system tray icon flashing when Claude Code (or any supported AI agent) completes a task. The tray icon toggles between the normal Clawd crab icon and an orange highlight circle.

## Changes

- `src/main.js`: Add `flashTaskbar()` and `stopTrayFlash()` using `tray.setImage()` toggle
- `src/state.js`: Call `flashTaskbar()` on task completion
- `src/prefs.js`: Add `flashTaskbarOnComplete`, `flashIntervalMs`, `flashDurationMs` prefs
- `src/settings-actions.js`: Register validators for new prefs
- `src/settings-tab-general.js`: Add collapsible settings group with toggle + sliders
- `src/settings-i18n.js`: 5-language translations (en/zh/zh-TW/ko/ja)
- `src/menu.js`: Expose `getTray()` for main process
- `assets/tray-icon-flash.png`: New orange highlight icon

## Settings

- **Toggle**: Enable/disable (default: on)
- **Flash interval**: 200-2000ms (default: 500ms)
- **Flash duration**: 0-60000ms (default: 5000ms, 0 = forever)

## Test Plan

- [x] `npm test` passes (2815 pass, 8 pre-existing failures unrelated)
- [x] Manual: task completion triggers tray flash
- [x] Manual: settings toggle disables flash
- [x] Manual: interval/duration sliders work
- [x] Manual: duration=0 flashes until tray click

?? Generated with [Claude Code](https://claude.com/claude-code)


rullerzhou-afk:
Thanks so much for the contribution, @linnin233 — really nice idea, and the i18n coverage across 5 languages is appreciated. I gave the branch a careful read with a second reviewer and we'd love to merge this, but found a few things worth fixing first. Sharing them below in rough order of severity:

**1. The new `assets/tray-icon-flash.png` looks corrupted (blocker)**

The PNG header parses fine, but the IHDR chunk length is wrong:

```
89 50 4E 47 0D 0A 1A 0A   ← PNG signature (OK)
0D 0D 0D 0D               ← IHDR chunk length — should be 00 00 00 0D
49 48 44 52               ← "IHDR"
```

A spec-compliant decoder (including Electron's `nativeImage`) will reject this. It looks like the file may have gone through a line-ending conversion at some point — the `0D 0A` inside the PNG signature is specifically there to detect that, and the bytes right after look like `0A → 0D` substitution. Could you re-export the asset and re-add it? A quick `file assets/tray-icon-flash.png` before committing should confirm it's a valid PNG.

Related: in `src/main.js` the load path only checks `fs.existsSync()` — it'd be worth also checking `!image.isEmpty()` after `createFromPath` so a future bad asset fails loud instead of silently flashing an empty icon.

**2. The auto-stop `setTimeout` isn't tracked, so back-to-back completions cut each other off**

In `flashTaskbar()`:

```js
if (trayFlashTimer) clearInterval(trayFlashTimer);  // interval is cleared
...
setTimeout(() => { stopTrayFlash(); }, durationMs || 5000);  // not stored
```

If two task-completion events fire within `durationMs` of each other, the first call's `setTimeout` is still in flight and will stop the second flash partway through. Fix: store the timeout in a module-level `trayFlashStopTimer` and `clearTimeout` it alongside the interval at the top of `flashTaskbar()` and inside `stopTrayFlash()`.

**3. macOS menubar template icon gets clobbered**

`src/menu.js` creates the tray with `tray-iconTemplate.png` + `setTemplateImage(true)` on macOS, but the flash logic always caches and restores `tray-icon.png` (resized to 32×32). After the first flash on macOS, the menubar icon permanently loses its template behavior (it'll render as a fixed-color image instead of adapting to light/dark menubar).

Two options:
- Platform-gate the whole feature to Windows/Linux, or
- In the cached `trayFlashNormalIcon` branch, detect `process.platform === 'darwin'` and load `tray-iconTemplate.png` with `setTemplateImage(true)` instead.

**4. `flashIntervalMs` / `flashDurationMs` aren't registered for in-place settings patching**

`SESSION_CLEANUP_NUMBER_KEYS` in `settings-tab-general.js` is the gate that tells `patchInPlace()` how to update number-input controls without a full re-render. The two new number prefs aren't in that set, so when they change, `patchInPlace()` falls through to `generalSwitches.get(key)`, fails to find a switch, returns `false`, and the whole General tab re-renders. Because the new "Flash taskbar" group is `defaultCollapsed: true`, it'll fold back up after each save — annoying when you're trying to dial in a value.

Adding them to a number-keys set (or a new `FLASH_NUMBER_KEYS` set following the same pattern as `SESSION_CLEANUP_NUMBER_KEYS`) and mounting them via `mountedControls` should keep edits in place.

**5. Minor — naming/i18n**

The feature is described as "flash taskbar" / "任务栏闪烁" / "タスクバー点滅" etc., but the implementation flashes the **system tray icon**, not the Windows taskbar entry (which would be `BrowserWindow.flashFrame(true)` — a separate UI concept). Could you rename the strings to "tray icon" across all five locales? Otherwise users will look at the bottom taskbar wondering why nothing's flashing.

---

Again, thanks for picking this up — once those are addressed I'd be happy to merge. Let me know if any of the fixes above need more context.

linnin233:
Thanks for the thorough review, all five points addressed:

  1. PNG corruption — regenerated a valid 32×32 RGBA PNG via System.Drawing, and added
  !img.isEmpty() guard after createFromPath
  2. setTimeout tracking — added trayFlashStopTimer, cleared in both flashTaskbar() and
  stopTrayFlash()
  3. macOS template icon — flashTaskbar() now detects process.platform === 'darwin' and loads
  tray-iconTemplate.png with setTemplateImage(true)
  4. In-place patching — added FLASH_NUMBER_KEYS set, wired into patchInPlace() validation and
  sync loops (they already land in sessionCleanupControls via buildNumberInputRow)
  5. Naming — renamed to "tray icon" / "系统托盘" / "系統匣" / "트레이 아이콘" /
  "トレイアイコン" across all 5 locales

  Ready for another look.

rullerzhou-afk (approved):
Thanks for the contribution and for addressing the review points so quickly. I gave the updated branch another pass; the fixes look good, and I'm going to merge this now.

linnin233:
Thank you for your recognition. This is my first PR. I will continue to work hard and make more contributions to the project 💪

PR #358, opened 2026-05-28
feat: block system sleep while agent tasks are in progress

zxypro1:
Add a "keep awake while working" toggle (default off). When enabled, Clawd holds a powerSaveBlocker as long as any non-headless session is in a running state (anything other than idle/sleeping) and releases it once tasks finish or the toggle is turned off.


rullerzhou-afk:
Thank you so much for this PR — it's a really clean, thoughtful contribution, and it was a pleasure to read through. A few things I genuinely appreciated:

- It's default-off, so there's no behavior change for existing users — much appreciated.
- The wiring is complete end to end (schema → mirror setter → action validator → settings tab → effect router), and you even filled in all five locales. Thank you for being so thorough.
- `prevent-app-suspension` is exactly the right blocker type here — it blocks system sleep but still lets the display turn off.
- The `before-quit` release plus the `try/catch` guards mean the blocker won't leak.
- I really liked that you hooked `reconcile` into `broadcastSessionSnapshot` rather than `setState` — that also quietly covers the "last working session gets deleted on process exit" path, which a `setState`-only hook would have missed. Nice instinct.

I just have one small suggestion and a couple of totally optional follow-ups — nothing blocking, and please feel free to push back on any of it.

**A small thing, if you have a moment — running `reconcile` before the UI broadcasts**

`reconcilePowerSaveBlocker()` currently runs after `broadcastDashboardSessionSnapshot` / `broadcastSessionHudSnapshot` / `repositionFloatingBubbles` inside the broadcast callback. The snapshot signature is advanced *before* the callback runs (`emitSessionSnapshot` sets `lastSessionSnapshotSignature`, then broadcasts), and the whole callback is wrapped in a swallowing `try/catch` in `state.js`. So if one of those UI calls happens to throw synchronously, `reconcile` would get skipped — and because the signature has already moved, an identical follow-up snapshot won't re-broadcast, which could leave the blocker in the wrong state until the next state/settings change. It's a low-probability path and it self-heals on the next change, so it's really minor — but since hardening it is cheap, moving `reconcilePowerSaveBlocker()` to the top of the callback (or giving it its own `try/finally`) would close the gap. The case worth guarding against is the "miss-acquire" one — a task running while the machine is still allowed to sleep.

**Optional — a tiny unit test for the state mapping**

The router wiring is already tested, which is great. If you're up for it, the one piece not yet covered is `anySessionInProgress()`'s state→in-progress mapping. Pulling it out into a small pure helper and adding a state-table test (idle / sleeping / thinking / working / juggling / headless, etc.) would be a nice way to lock the intended semantics in place. Completely optional, though.

**A question on semantics — no change needed**

While an agent is waiting on a permission prompt, the session stays in `working`/`thinking` (the permission path preserves the prior state), so the blocker stays held during that wait. My read is that this is actually the behavior you'd want — you'd like the machine awake so you can come back and approve — so I think it's fine as-is. Just flagging it in case "while working" was meant strictly as "actively computing."

Thanks again for taking the time to build this — really appreciate the care that went into it. 🙏

zxypro1:
Thanks for the thorough and generous review — really appreciated. 🙏 Pushed a follow-up addressing both actionable points.

1. Reconcile before the UI broadcasts

Good catch. Since the snapshot signature is advanced before the callback runs and the callback is wrapped in a swallowing `try/catch`, a synchronously-throwing UI call could skip the reconcile and leave us in the "miss-acquire" state (task running while sleep is still allowed) until the next change. I moved `reconcilePowerSaveBlocker()` to the very top of the `broadcastSessionSnapshot` callback, ahead of the dashboard/HUD broadcasts and reposition, so the blocker update can't be skipped.

2. Unit test for the state mapping

Done. I pulled the state→in-progress logic out of `anySessionInProgress()` into a pure `isSessionInProgress(session)` helper in `state-session-snapshot.js` (next to `deriveSessionBadge`, since it mirrors the same "running" semantics), and added a state-table test covering `idle` / `sleeping` / `working` / `thinking` / `juggling` / `notification` / `attention` / headless / nullish. `main.js` now just iterates sessions and delegates to the helper.

3. Semantics of waiting on a permission prompt

Your read matches my intent — while waiting on a permission prompt the session keeps its prior `working`/`thinking` state, so the machine stays awake and you can come back to approve. Treating that as in-progress is deliberate.

Thanks again for the care you put into this review!

rullerzhou-afk:
Thank you again for the thoughtful PR and for turning around the follow-up so carefully. I merged #358.

I also added a tiny follow-up commit on main to make the tests/comments around persisted session states a little more explicit, but that is just maintainer cleanup on our side and there is no action needed from you.

Really appreciate the care here: default-off behavior, the right \\prevent-app-suspension\\ mode, complete settings/i18n wiring, and the quick review response. This is a very nice contribution.

PR #370, opened 2026-05-30
feat(usage): Claude/Codex subscription rate-limit gauges (local + remote SSH)

Tonic-Jin:
## What

Adds **subscription usage gauges** for Claude (Claude Max/Pro) and Codex CLI to the desktop pet — small gauges showing 5h / weekly rate-limit utilization, with a Settings tab to toggle providers, polling interval, and display mode. Usage is read from **local + remote (Remote SSH)** sources.

## Why

If you work against a Claude/Codex subscription, there's no at-a-glance signal of how close you are to the 5-hour / weekly limit until you hit it. These gauges surface that next to the pet.

## How the Claude side avoids rate-limiting (important)

The Claude usage endpoint (`/api/oauth/usage`) is a **per-access-token bucket that 429s after only ~5 requests and stays sticky for 30-60 minutes**. Naive polling (even 60s) reliably self-inflicts a rate limit — this is well documented across community usage monitors. So the Claude path is a **hybrid**:

1. **statusLine stdin tap (primary, zero API calls).** `hooks/clawd-statusline-tap.js` registers as a Claude Code `statusLine` command. CC pushes `rate_limits` on stdin every prompt; the tap extracts **only** `rate_limits` (no session_id/cwd/transcript leak), atomically writes `~/.clawd/statusline.json` (mode 600, 30s throttle), and transparently chains any existing statusLine via `--chain` (stderr ignored so a broken chain can't dirty the statusline). For users who actively run CC, the gauge is fed entirely from stdin — **the usage API is never called.**
2. **API fallback (only when the snapshot is stale/absent).** If CC isn't running (snapshot stale) the gauge falls back to the local + remote usage API, now at a **5-minute default** (was 60s) with a **429 backoff** (honors `Retry-After`, else 30min, capped) and a **last-good cache** so a transient failure/429 keeps the last value instead of blanking.

Wiring: the tap is registered/removed alongside the Claude Code integration sync (startup + enable/disable), gated by `manageClaudeHooksAutomatically` + claude-code enabled, mirroring the other Claude hooks. It restores any chained original on uninstall and never touches a statusLine it doesn't own.

### Codex side

Reads `~/.codex/sessions` rollout JSONL locally and over SSH for remote profiles; merged freshest-wins by JSONL record timestamp. Codex records rate_limits directly in its JSONL, so there's no API and no rate-limit concern.

### Shared

- `src/usage-remote-fetch.js`: one SSH fetcher (spawn / timeout / child-tracking / `close`-vs-`exit` stdout-drain guard) for both providers.
- All paths are `os.homedir()` / `path.join` based; **no new npm dependencies** (Node stdlib only); Win/macOS/Linux window handling; en/zh/ko/ja strings.
- Missing credentials/sessions/snapshot degrade gracefully to a hidden gauge — no crash.

## Security

- statusLine tap persists **only** rate_limits; all other stdin fields (session_id, cwd, transcript_path) are discarded. Snapshot file is mode 600.
- Tokens travel only inside the SSH-encrypted channel (remote snippet streams over stdout; no plaintext temp files).
- `requestJson` error bodies are truncated with `access_token`/`refresh_token` values redacted so nothing downstream can log credentials.
- Credentials rewrite (token refresh) preserves file mode (600) via atomic tmp+rename.

### Token refresh detail

The Claude OAuth **token refresh** call must NOT send a `user-agent: claude-code/...` header — the token endpoint 429s requests posing as the real CLI, while the identical request with only `Content-Type` returns 200. Endpoint is `https://console.anthropic.com/v1/oauth/token` with the public PKCE `client_id` (not stored in credentials, so it's a constant). The usage GET still needs the UA; only refresh omits it.

## Tests

`node --test test/usage-*.test.js`: **71 pass / 0 fail** — parsers (incl. sub-1% regression: a real 1% must not render as 100%), freshest-wins merge, statusLine tap (extract / throttle / chain passthrough / no-leak / stderr ignored), snapshot reader (fresh/stale/missing/malformed), hybrid fetchClaude (snapshot-skips-API / stale-falls-back), 429 backoff + last-good retention, install wiring, and remote SSH fetch for both Claude and Codex.

## Question for maintainers

The gauge defaults to **`enabled: true`**. If your convention is opt-in for new features, say so and I'll flip it to `false` (it degrades to hidden when no data is present, so it's harmless either way).


rullerzhou-afk:
Thanks for the work here. I like the problem this is trying to solve: having a quick sense of Claude/Codex quota before starting work is genuinely useful, and I do think this belongs in Clawd in some form.

I do not think this is mergeable as-is yet. There are a few technical blockers, plus one UX direction point I would like to align on before more work goes into polishing the floating gauge surface.

## Technical issues

### Blocker: packaged builds appear to install a non-Node statusLine command

`hooks/statusline-install.js` defaults `nodeBin` to `process.execPath`, and `src/integration-sync.js` calls `registerStatuslineTap()` without passing a node path. The resulting command is built as:

```js
"${nodeBin}" "${scriptPath}"
```

In a packaged Electron app, `process.execPath` is the Clawd/Electron executable, not the system Node binary. The existing hook installers avoid this by using `resolveNodeBin()` from `hooks/server-config.js`; this new installer looks like the outlier.

User impact:

- the statusLine tap may not run in packaged builds;
- the gauge then falls back to the Claude usage API path, which is exactly the path this PR is trying to avoid;
- an existing user statusLine chained behind the tap may stop rendering;
- Claude Code may launch the Clawd executable on every prompt instead of running the tap script.

Please route this through the same Node resolution path as the other installers. If Node cannot be resolved, I would rather leave the existing statusLine untouched than write a broken command.

### High: the new wiring test mutates real user/CI agent configs

`test/usage-statusline-wiring.test.js` uses `isAgentEnabled: () => true` and then calls `syncEnabledStartupIntegrations()`. That function iterates the non-Claude syncers in `src/integration-sync.js`, while the test only stubs the Claude-specific implementations.

In practice, running the usage tests triggered real hook/plugin sync logs for other agents. This test should not touch real Gemini/Cursor/Copilot/Kiro/Kimi/Qwen/Codex/opencode/OpenClaw/Hermes configs.

Please scope the test to Claude only, for example by making `isAgentEnabled(id)` return true only for `claude-code`, or by fully stubbing every syncer it can reach.

### Medium: the 0600 assertion is not portable to Windows

`test/usage-statusline-tap.test.js` asserts `fs.statSync(...).mode & 0o777 === 0o600`. On Windows this is not a reliable POSIX permission check; I see the usage test suite fail there with 70 pass / 1 fail on this assertion.

There is already a local precedent in `test/telegram-approval-settings.test.js` that skips the tight mode assertion on `win32`. Please use the same pattern or guard this assertion by platform.

### Medium: disabling the usage gauge does not stop the statusLine tap

The tap lifecycle is currently tied to Claude integration sync/stop, not to `usageGauge.enabled` or the Claude provider toggle. Disabling the usage gauge stops the runtime/window, but the statusLine tap can remain installed and continue writing `~/.clawd/statusline.json`.

The data is limited to `rate_limits`, so this is not as severe as the packaged-command issue, but the product boundary matters: if a user disables the usage feature, we should stop collecting/writing usage snapshots too, or make that lifecycle explicit and user-visible.

## Product direction / UX alignment

The part I want to align on is the default surface. I am hesitant to add another always-on floating window next to the pet by default. We already received feedback that the Session HUD felt intrusive when it was always visible, and that is why it moved toward an on-demand interaction. I worry that a persistent usage gauge next to the pet would run into the same issue, even if the underlying information is valuable.

A direction that feels more aligned with Clawd would be something pet-attached and on-demand: hidden by default, then shown when the user clicks the pet, perhaps as two lightweight arc/rainbow indicators above the pet for the active provider's 5h and weekly usage. The Dashboard can still own the full detailed Usage view, including all providers, local/remote sources, freshness, reset times, and errors.

So this is not a rejection of the feature idea. I just want to avoid spending more time polishing an always-on floating gauge if that is not the surface we want to ship. Let's align on the UX direction before continuing with this implementation.

Tonic-Jin:
@rullerzhou-afk Apologies for the long silence — my day job took over and I only caught up today. Thanks for the detailed review back in May.

Since then #590 has landed, and I've read through it — it covers the Claude Code (local) side of this PR with a much cleaner statusline/quota-bucket infrastructure, and I agree with the on-demand UX direction. That makes most of this branch obsolete — I'll close this in favor of a fresh PR once we've aligned on the remaining scope below.

@Yike-Ye on the Codex part — either way works for me: I can finish it, or you can take it, whichever suits you. I have an active Codex subscription, so I can validate it against real data in either case — I'll keep an eye on the PR.

As for the remote SSH side (pulling Claude/Codex usage from remote machines over SSH, freshest-wins merge): for someone who mainly works on remote machines — which is exactly my setup — the usage feature is only really useful with remote collection, so I think it makes sense for me to carry this part forward. It would be a fresh PR on top of the merged quota-bucket/claude-rate-limits infrastructure — no new floating window, data surfacing in the Dashboard account-usage section like #590, with remote sources as additional groups. And since statuslines are inherently local, it would fetch remote data over SSH rather than deploy statuslines remotely. @rullerzhou-afk @Yike-Ye would that direction work for you both?

Yike-Ye:
Thanks @Tonic-Jin! The direction sounds right to me. Since you have an active Codex subscription (real-data validation) and mainly work on remote machines, both the Codex and SSH parts fit your environment much better than mine — happy for you to carry both. Feel free to ping me with any questions about the current #590 code (quota-bucket / claude-rate-limits / the Dashboard account-usage rendering) you'll be building on.

By the way, not sure whether you are working on Windows or macOS, but I am more than happy to run the test on my mac if you need.

rullerzhou-afk:
Welcome back @Tonic-Jin — thanks for re-scoping against what landed in the meantime.

Direction confirmed on my side:

- Closing this PR in favor of a fresh one on top of the #590 infrastructure: yes.
- You carrying both the Codex part and the remote SSH part: yes — @Yike-Ye's split works for me too.
- No new floating window; surfacing in the Dashboard account-usage section: exactly the surface I was hoping for after the May review.

Your "statuslines are inherently local, pull over SSH instead" read matches the codebase as it stands: `hooks/install.js` skips statusline registration on `--remote` installs, with a comment that says *"remote/SSH statusline support is an intentional non-goal"*, and the SSH deploy manifest doesn't ship `claude-statusline.js` at all. So SSH pull genuinely is the only channel for remote quota — you won't be competing with a passive path.

Three design points to settle in the fresh PR, from digging through the current plumbing:

1. **Quota needs a home that isn't a session.** Today `claudeQuota` lives on session records only — `updateSessionMetadata` in `src/state.js` drops the write (and never creates a session) when no session exists. Your headline scenario — "check remote quota before starting work" — is precisely the case where the remote has no live session, so the fresh PR needs a per-host, session-independent store in the state layer, not just UI changes.
2. **Per-source grouping in the Dashboard, as you proposed.** `resolveQuotaForDisplay` in `src/dashboard-renderer.js` is single-slot freshest-wins across all sessions of an agent — the comment above it assumes "it's the same number regardless of which session reported it most recently". That assumption breaks the moment local and remote are different subscriptions, so grouping by source (local + one group per remote host), keeping freshest-wins within a source, is the right shape.
3. **Say what the pull actually executes.** `rate_limits` only exists as the statusline stdin payload — it isn't persisted anywhere on the remote disk, and there's no prior art in the repo for querying it on demand. So the design should state the mechanism explicitly (one-shot API/CLI query on the remote using its own credentials vs. a small remote-side cache fed by something), and surface staleness in the UI either way — the absolute-`resetAt` convention in `hooks/quota-bucket.js` exists for exactly that.

Guardrails, mostly continuations of the May review:

- All SSH invocations go through the existing `remote-ssh-runtime` profiles (`buildSshArgs`) — same lesson as the statusline installer in May: ride the existing resolution paths.
- Pull lifecycle follows the toggles: disabling the usage feature (or a remote profile) stops all pulling for it; prefer on-demand triggers (Dashboard open / manual refresh) over a resident timer.
- No key/credential management in the app — keep delegating to the user's ssh config/agent via the existing profile fields.

One small heads-up: the `CLAWD_REMOTE` branch in `hooks/claude-statusline.js` currently has no real callers (SSH doesn't ship the file; the WSL install registers the statusline without setting that env var), so if you build on it, treat it as unexercised rather than battle-tested.

With that, scope is aligned — feel free to close this one and open the fresh PR whenever you're ready. Happy to review early drafts; I'm on Windows daily so I can cover that side, and @Yike-Ye has kindly offered macOS.

Tonic-Jin:
Thanks for the detailed confirmation — direction received, and agreed on points 1 and 2: the fresh PR adds a session-independent per-source quota store in the state layer, and the Dashboard groups per source (local + one group per remote host), keeping freshest-wins within a source only.

On point 3, one mechanism decision I want in front of you explicitly, because after reading the deploy plumbing I did **not** build a forward-SSH pull — correcting my earlier "pull over SSH" phrasing:

**Mechanism (draft branch): ride the existing remote hook channel.** The deploy already ships hook scripts that POST to `127.0.0.1:<remoteForwardPort>` — the reverse tunnel the runtime maintains — with `CLAWD_REMOTE=1` (`codex-remote-monitor.js` is the precedent). The branch extends exactly that:

- `claude-statusline.js` joins the deploy manifests; `install.js --remote` registers it with the same env-prefix command form remote hooks already use. Its `rate_limits` POSTs are `metadata_only`, so the B2 semantics hold.
- `codex-remote-monitor.js` forwards the `rate_limits` already carried by rollout `token_count` events (currently dropped), freshness-gated against restart replays.
- The previously-unexercised `CLAWD_REMOTE` branch is now unit-tested; I've also manually exercised it on a Linux remote.

**Why not a one-shot pull:** live rate-limit usage exists in exactly two places — the statusline stdin payload and the OAuth usage API; nothing persists it on the remote disk (verified against real transcripts). The API path works but is fragile: my May branch needed a UA-drop plus Retry-After backoff to stop 429s, and kept it as a rate-limited fallback only. So an on-demand pull either leans on that API or needs "a small remote-side cache fed by something" — and a statusline POSTing through the tunnel *is* that feed, minus the polling and the cache file.

**Against your guardrails — including two gaps I'm owning rather than claiming away:**

- No new SSH invocations, no credential handling: data rides the tunnel the runtime already maintains via the existing profiles; nothing reads `.credentials.json` on either side. Only `{usedPercent, resetAt}` plus a session/host label ever leave the remote.
- No resident processes we own: the statusline runs per refresh inside Claude Code's own lifecycle; the only remote daemon is the already-shipped Codex monitor behind its existing opt-in. Honest tension with "prefer on-demand": push is continuous-while-active rather than on-demand — it fires exactly while a session is live on that machine (which is when the numbers move from that machine's side) and goes fully quiet otherwise. A Dashboard-open refresh trigger can be added on top without changing the transport if you want one.
- **Gap 1 — remote lifecycle:** today there is no undeploy path (true of the existing remote hooks too), so deleting or disconnecting a profile would leave the statusline registered on the remote, POSTing into a dead local port. Before the PR I'll add a remote-unregister step: ship `uninstall.js` in the manifests and run it on profile delete/disconnect, mirroring `stopCodexMonitor`.
- **Gap 2 — staleness:** `resetAt` renders as a countdown/date today, but two things are still missing and go into the store rework: buckets must expire once `now > resetAt` (the windows reset on wall-clock, so a stale bar would otherwise mislead high), and quiet sources get an explicit "as of N minutes ago" label. One caveat that applies to pull and push alike: usage from outside any Clawd-connected machine (claude.ai in a browser) stays invisible until the next report — the same limitation the local path already accepts.
- Cold start: you're right that "check before starting work" is exactly where a session-bound design fails — that is what the session-independent store fixes (last-known numbers survive session end and app restart, with the expiry and staleness handling above). Until it lands, the branch can't serve that case; it's the first rework item.

Occupied slots (my own machines run a third-party statusline): an **opt-in per-profile chain mode**, isolated in its own commit so you can judge it separately — deploy captures the user's original `statusLine` object verbatim into a sidecar file, takes the slot with `--chain`, re-feeds stdin to their command (stdout inherited, stderr swallowed), and unregister restores the original object. Default behavior everywhere stays never-touch, exactly as in #590.

Draft branch: https://github.com/Tonic-Jin/clawd-on-desk/tree/feat/usage-remote-quota — rework order from here: session-independent store with expiry/staleness, per-source grouping, remote unregister; I'll open the PR after those land. If push-through-tunnel doesn't sit right with you, please say so before I polish further — the extractor/normalization layers don't care how the bytes arrive, so switching the transport would be a contained change.

Closing this PR now in favor of the fresh one, as agreed.

PR #372, opened 2026-05-30
fix(hooks): use bash-compatible command format on Windows

Ac-spider:
## Summary

- Remove `shell: "powershell"` and `&` call-operator prefix from Windows hook commands
- Claude Code runs hooks with bash on Windows, so PowerShell syntax causes execution failures
- Use the same quoted command format as macOS/Linux

## Problem

On Windows, `buildCommandHookSpec` generated:
```json
{
  "shell": "powershell",
  "command": "& \"node\" \"script\" args"
}
```

But Claude Code runs hooks with bash, so the `&` (PowerShell call operator) fails in bash.

## Fix

Generate the same format as other platforms:
```json
{
  "command": "\"node\" \"script\" args"
}
```

No `shell` field, no `&` prefix. The `syncCommandHook` function automatically removes stale `shell` metadata from existing hooks on re-registration.

Fixes hook errors and `$`-in-path issues on Windows.

### DIFF
--- a/hooks/install.js
+++ b/hooks/install.js
@@ -824,8 +824,8 @@ function registerHooks(options = {}) {
       changed = true;  // format was normalized, need to persist
     }
 
-    // Local Windows hooks must use explicit PowerShell invocation because Claude
-    // Code defaults command hooks to bash on Windows. Remote hooks stay on the
+    // Local Windows hooks use bare commands (no shell wrapper) because Claude
+    // Code runs command hooks with bash on Windows. Remote hooks stay on the
     // legacy POSIX/bash-compatible form; see buildCommandHookSpec().
     const desiredHook = buildCommandHookSpec(nodeBin, hookScript, event, {
       platform,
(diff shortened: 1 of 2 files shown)


rullerzhou-afk:
Thanks for looking into this. I agree there is a real edge case here: a `$` character inside a Windows path can be problematic when the command is evaluated by PowerShell, so the concern is valid.

However, I’m going to close this PR as-is because removing `shell: "powershell"` takes Claude Code hooks back to the exact command shape that caused #147. The bare `"node" "script" Event` form is executed through Claude Code’s default Windows shell path, which can be Git Bash / WSL bash, and that was the root cause fixed by 782f1ef. The reporter confirmed that the PowerShell shell + `& "node" ...` format resolved the Windows startup hook error.

So I don’t want to trade a rare `$`-in-path edge case for a known regression that already affected Windows users.

The right follow-up would be either:

- narrowly fix PowerShell quoting so `$` in paths is preserved, while keeping `shell: "powershell"`, or
- explore Claude Code’s exec-form hook config (`command` + `args`) so we avoid shell quoting entirely, but only after a real Windows Claude Code smoke test confirms it works with our async/timeout fields.

Thanks again for spotting the `$` path issue. That part is useful; we just need to fix it without reintroducing #147.

PR #373, opened 2026-05-30
fix(gemini): use bash-compatible command format on Windows

Ac-spider:
## Summary

- Gemini hook commands on Windows used the default PowerShell `&` prefix, but Gemini CLI runs hooks with bash - the same bug fixed for Claude hooks in #372
- Changed `buildGeminiHookCommand` to default to `windowsWrapper: none` (bare command, no `&` prefix)
- Updated test to match the new expected output

## Test plan

- [x] All 13 Gemini install tests pass
- [x] Verified buildGeminiHookCommand produces bare command format on Windows by default
- [x] Verified explicit windowsWrapper: cmd override still works

### DIFF
--- a/hooks/gemini-install.js
+++ b/hooks/gemini-install.js
@@ -33,7 +33,8 @@ function buildGeminiHookEntry(command) {
 }
 
 function buildGeminiHookCommand(nodeBin, hookScript, event, options = {}) {
-  return formatNodeHookCommand(nodeBin, hookScript, { ...options, args: [event] });
+  const opts = options.windowsWrapper !== undefined ? options : { ...options, windowsWrapper: "none" };
+  return formatNodeHookCommand(nodeBin, hookScript, { ...opts, args: [event] });
 }
 
 function replaceEntry(target, source) {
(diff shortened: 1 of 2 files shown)


rullerzhou-afk:
Thanks for the follow-up here. I’m going to close this PR as-is for the same broad reason as #372: the premise does not match the current Gemini CLI hook runner.

Gemini CLI’s current hook implementation executes command hooks through its platform shell configuration. On Windows that path is PowerShell, not bash. With that execution model, the bare `"node" "script" "event"` form is not executable PowerShell syntax; it is exactly why Clawd currently keeps the `& "node" ...` prefix for Gemini hooks. A quick PowerShell smoke check confirms the bare form fails with an unexpected-token parser error, while the `&` form runs correctly.

There is also a product-context wrinkle now: Google’s May 19 transition announcement says Gemini CLI will stop serving requests for Google AI Pro / Ultra and free individual users on June 18, 2026, while enterprise/API-key access remains supported. So we should be especially conservative about changing this integration unless there is current, agent-specific evidence that the existing Windows command format is broken.

If there is a separate path-quoting edge case, we should handle that narrowly without removing the PowerShell invocation. Thanks again for looking into these Windows hook paths; this one just points in the wrong direction for Gemini CLI.

Reference: https://developers.googleblog.com/an-important-update-transitioning-gemini-cli-to-antigravity-cli/

PR #375, opened 2026-05-30
feat: add LAN WebSocket bridge for PWA mobile clients

Bynlk:
## 概要

在桌面端新增一个轻量 WebSocket 桥接服务器，让 PWA 手机客户端可以实时查看会话状态、远程审批权限请求。

**核心原则：零侵入**——`state.js`、`permission.js`、`server.js` 完全不动，仅在 `src/main.js` 加了 ~18 行 init/start/cleanup hooks。

## 改动清单

| 文件 | 行数 | 说明 |
|------|------|------|
| `src/network/lan-ws-server.js` | +280 | WebSocket + HTTP 桥接服务器 |
| `pwa/` (6 文件) | +900 | PWA 客户端 |
| `test/lan-ws-server.test.js` | +230 | 9 个测试用例 |
| `docs/mobile-protocol-v1.md` | +180 | Protocol v1 协议文档 |
| `src/main.js` | +18 | 启动/停止 hooks |
| `src/settings-ipc.js` | +30 | PC 端设置页 IPC |
| `src/settings-tab-mobile.js` | +90 | PC 端 Mobile 标签页 |
| `src/settings.html/css/i18n` | +90 | 样式 + 国际化 |

## 工作原理

```
Clawd Desktop (Electron)
  ├── HTTP Server (:23333) — 原有，不动
  └── LAN WS Bridge (:23334) — 新增
        ├── ws://host:23334/ws?token=<hex>
        └── /mobile/* (PWA 静态文件)

PWA (手机浏览器)
  ├── 自动连接（URL参数 → 历史记录 → Bridge API）
  ├── 会话状态实时展示
  └── 权限审批（允许 / 拒绝）
```

## 安全

- Token 认证（32位 hex，WS 握手时验证）
- 仅局域网暴露（0.0.0.0，不转发端口的话外网不可达）
- 频率限制：60 msg/60s per client
- 无效 Token → close code 1008

<details>
<summary>运行测试</summary>

```bash
node --test test/lan-ws-server.test.js
```

覆盖：协议版本、启动绑定、静态文件、Token 拒绝、快照、状态广播、会话删除、权限审批流程
运行 9 个测试全过
</details>

<details>
<summary>限制</summary>

- 无 TLS（仅局域网）
- 会话状态最多 2s 延迟（轮询间隔）
- `tool_output` 未桥接（计划 v2）
- Token 不可主动撤销

</details>

<details>
<summary>协议 v1 详细规范</summary>

见 `docs/mobile-protocol-v1.md`

服务端消息类型：
- `snapshot` — 连接时全量会话状态
- `state` — 增量更新
- `session_deleted` — 会话移除
- `permission_request` / `elicitation_request` — 需要审批
- `permission_dismissed` — 已解决

客户端发送：`permission_response` / `elicitation_response`

</details>


rullerzhou-afk:
Thanks for working on this. I like the product direction here: a local-first mobile companion for session preview and remote approval fits Clawd pretty well, and I agree this should not become a remote shell/control plane.

That said, I don’t think this version is mergeable yet. The current LAN bridge has a much weaker authorization boundary than our existing Telegram approval path.

Blocking issues I found:

1. `/api/connection-info` returns the WebSocket token without authentication  
   In `src/network/lan-ws-server.js`, the unauthenticated `/api/connection-info` endpoint returns `{ port, token, lanIp }`, and also includes `getSettingsSnapshot()` when available. Since the WebSocket auth relies on that same token, any device on the same LAN can fetch the token and then connect as an approved client. This effectively defeats the token check.

2. The bridge starts by default and binds to `0.0.0.0`  
   `src/main.js` starts `_lanWss.start()` unconditionally, and the server listens on `0.0.0.0`. There is no Settings enable/disable gate, no pairing step, and no device confirmation. For an approval-capable endpoint, this needs to be explicit opt-in.

3. Mobile clients can directly resolve permissions  
   Once connected, `permission_response` maps straight into `resolvePermissionEntry(..., "allow" | "deny")`. That is the right eventual capability, but only after the auth/pairing boundary is at least as strong as Telegram approval.

4. Elicitation appears broken end-to-end  
   The bridge sends `prompt/options`, but Clawd’s AskUserQuestion flow uses `questions[]`. The PWA replies with `answers: { value: ... }`, while the server normalizes by looking up `answers[q.question]`. So elicitation cards likely render empty or resolve with empty answers.

5. PWA Settings tab has a runtime error  
   `pwa/app.js` calls `this.settingsRenderer.fetchPcSettings()`, but `SettingsRenderer` does not define that method.

I’d suggest splitting this into smaller steps:

M1:
- Mobile bridge is disabled by default.
- User explicitly enables it in Settings.
- Start with read-only session preview only.
- Do not expose the long-lived token from an unauthenticated LAN endpoint.
- Avoid sending full `cwd`, `sessionTitle`, recent events, or settings snapshot unless sanitized.
- Fix the PWA Settings tab crash.

M2:
- Add pairing / device confirmation.
- Add token rotation and revoke/disconnect controls.
- Then enable mobile Allow/Deny.
- Align filtering with Telegram approval: no DND/native-fallback/hidden/headless/opencode/elicitation/passive-notification approvals unless intentionally designed.
- Rework elicitation using the real `questions[]` contract and verify with a real Claude Code AskUserQuestion flow.

So my recommendation is: **request changes** for this PR as-is, but keep the direction. I think a safer M1 read-only mobile preview would be a good first slice.

Bynlk:
Thanks for the thorough breakdown. Agreed on all the blocking issues — especially the `/api/connection-info` token leak and the default-on binding.

I'll split this out: new branch off `main`, M1 only (read-only session preview, disabled by default, no unauthenticated endpoints). Will link back here once the PR is open and close this one.

Bynlk:
New PR: #391

PR #387, opened 2026-05-31
feat: wire Hermes Agent into permission bubble and elicitation system

29206394:
## Summary
- Hermes Agent now uses the unified `/permission` route and bubble UI for clarify (elicitation) and general tool permission approval, replacing the legacy `/clarify` native dialog
- Plugin `pre_tool_call` hook intercepts clarify tool → converts to elicitation format → POSTs to `/permission` → Clawd shows the same bubble UI as Claude Code's AskUserQuestion
- General tool permission approval supported via `CLAWD_HERMES_PERMISSION_TOOLS` env var
- All fallback paths return 204 (no-decision) so Hermes falls through to its native flow when Clawd is unavailable, DND, or agent is disabled
- Fixed `plugin.yaml` missing `hermes_version` field which prevented Hermes from loading the plugin

## Changed files
- `agents/hermes.js` — enable `permissionApproval` and `interactiveBubble` capabilities
- `hooks/hermes-plugin/__init__.py` — replace `_post_clarify` with `_post_permission`, add `_handle_clarify_tool` (elicitation), `_handle_permission_request` (general approval)
- `hooks/hermes-plugin/plugin.yaml` — add `hermes_version`, bump to 0.2.0
- `src/server-route-permission.js` — add Hermes agent branch (DND/disabled/bypass → 204, elicitation + permission → bubble)
- `src/permission.js` — add `isHermes` resolution (elicitation returns answers, permission returns allow/deny), bubble close → no-decision
- `test/registry.test.js` — update Hermes capabilities assertions

## Test plan
- [x] `npm test` — all existing tests pass (10 pre-existing failures unrelated)
- [x] curl simulation: `POST /permission` with `agent_id: "hermes"` shows elicitation bubble, returns `{decision, answers}`
- [x] Real Hermes Agent: clarify tool triggers Clawd bubble, user selection returned correctly

### DIFF
--- a/agents/hermes.js
+++ b/agents/hermes.js
@@ -18,8 +18,8 @@ module.exports = {
   },
   capabilities: {
     httpHook: false,
-    permissionApproval: false,
-    interactiveBubble: false,
+    permissionApproval: true,
+    interactiveBubble: true,
     sessionEnd: true,
     subagent: false,
   },
--- a/hooks/hermes-install.js
+++ b/hooks/hermes-install.js
@@ -55,6 +55,41 @@ function pathExists(filePath) {
   }
 }
 
+function discoverHermesProfileHomes(hermesHome) {
+  const profilesDir = path.join(hermesHome, "profiles");
+  let entries = [];
+  try {
+    entries = fs.readdirSync(profilesDir, { withFileTypes: true });
+  } catch {
+    return [];
+  }
+
+  const homes = [];
+  for (const entry of entries) {
+    if (!entry || !entry.isDirectory()) continue;
+    const profileHome = path.join(profilesDir, entry.name);
+    if (!pathExists(path.join(profileHome, "config.yaml"))) continue;
+    homes.push(profileHome);
+  }
+  homes.sort((a, b) => a.localeCompare(b));
+  return homes;
+}
+
+function hermesHomesForSync(options = {}) {
+  const hermesHome = resolveHermesHome(options);
+  const homes = [hermesHome];
+  if (options.syncProfiles === false) return homes;
+
+  const seen = new Set(homes.map((home) => path.resolve(home)));
+  for (const profileHome of discoverHermesProfileHomes(hermesHome)) {
+    const resolved = path.resolve(profileHome);
+    if (seen.has(resolved)) continue;
+    seen.add(resolved);
+    homes.push(resolved);
+  }
+  return homes;
+}
+
 function hermesCommandCandidates(options = {}, hermesHome = resolveHermesHome(options)) {
   const env = options.env || process.env;
   const platform = options.platform || process.platform;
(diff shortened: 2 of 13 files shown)


rullerzhou-afk:
Thanks for wiring Hermes into the permission bubble path. I reviewed this against current `origin/main`, and I don't think this is ready to merge yet.

Requesting changes. The main blockers are:

**Needs rebase first**

`src/permission.js` and `src/server-route-permission.js` conflict with current `origin/main`, and the conflicting area overlaps the recent Telegram/Copilot/opencode/Codex permission routing changes.

**Dead Hermes permission toggle**

`agents/hermes.js` now advertises `permissionApproval` / `interactiveBubble`, so Settings renders the `permissionsEnabled` sub-switch. But `src/prefs.js` still defaults Hermes to only `{ enabled: true }`, `normalizeAgents()` strips `permissionsEnabled`, and `agent-gate` treats the missing flag as enabled. A user can turn the Hermes permission bubble off in the UI, but the setting will not persist and the bubble will still appear.

**Hermes permission metadata is dropped**

The plugin posts `cwd`, `agent_pid`, and process metadata such as `source_pid`, `pid_chain`, and `editor`, but the Hermes `/permission` branch only passes `{ agentId: "hermes" }` into `updateSession()` and does not store those fields on the permission entry. That means Hermes permissions lose folder/session context and terminal focus metadata, unlike Codex/Qwen.

**Clarify interception uses the error channel**

The current clarify interception returns `{"action":"block","message":"User selected: ..."}` from `pre_tool_call`. Hermes' documented contract says `block` short-circuits the tool and returns the message to the model as a tool error, not as a successful tool result. Please either provide a real Hermes transcript showing this is intentionally consumed as the clarification answer, with a code comment explaining the compromise, or remove the clarify interception until Hermes exposes a result-injection hook.

**Needs targeted tests**

Before merge, please add tests for:
- prefs persistence for `agents.hermes.permissionsEnabled`
- normal Hermes permission request entry creation and response shape
- elicitation/clarify entry behavior
- DND/disabled/subgate fallback returning no-decision/204
- bubble creation failure returning fallback without leaking a pending entry

**Other fixes before merge**

- `_post_permission` should use a short Clawd probe/header check before the long blocking POST, and should respect the no-server cooldown. Otherwise a non-Clawd service on a candidate port can block a Hermes tool call for up to 600s.
- Runtime dismiss for Hermes should send the existing Hermes no-decision/204 response instead of destroying the socket.
- `hermes_version` in `plugin.yaml` appears to be ignored by the current official Hermes plugin manifest parser; please remove it or point to the Hermes version/source that actually consumes it.

Once rebased, please rerun the relevant registry, prefs, permission route/bubble, and Hermes plugin tests.

29206394:
Thanks for the detailed review. I rebased and addressed the requested changes. Summary by item:

### Rebase
Rebased onto current `origin/main` and resolved the overlapping permission-route conflicts with the recent Telegram/Copilot/opencode/Codex routing changes.

### Hermes permission toggle persistence
Fixed the dead Hermes sub-switch by adding Hermes permission/notification flags to prefs defaults/normalization:
- `agents.hermes.permissionsEnabled`
- `agents.hermes.notificationHookEnabled`

Updated prefs tests so the Hermes Settings sub-switch now persists instead of being stripped by `normalizeAgents()`.

### Hermes permission metadata
Added Hermes session metadata normalization and propagation. Hermes `/permission` entries now keep the plugin-posted context:
- `cwd`
- `agent_pid` / `agentPid`
- `source_pid` / `sourcePid`
- `pid_chain` / `pidChain`
- `editor`

This is used in both normal permission requests and elicitation/clarify entries, and is passed through `updateSession()` like the Codex/Qwen paths.

### Clarify interception / error-channel compromise
Kept the interception but documented the compromise in code. Hermes currently has no result-injection hook for `pre_tool_call`; returning `{"action":"block","message":"User selected: ..."}` is translated by Hermes into a tool result shaped as an error. I added a docstring explaining that this should migrate to a cleaner result-injection hook when Hermes exposes one.

I also ran a real Hermes transcript to verify the model intentionally consumes this as the clarification answer:

```text
session: 20260601_122638_a3596a

user:
Use the clarify tool to ask exactly: Pick one? with choices Red and Blue.
After the user response, reply with exactly: CHOSEN=<answer>.
Do not ask in normal text; use the clarify tool.

assistant tool call:
clarify({
  "question": "Pick one?",
  "choices": ["Red", "Blue"]
})

tool result:
{"error": "User selected: Blue"}

assistant final:
CHOSEN=Blue
```

Raw message sequence from `~/.hermes/state.db`:

```text
--- 73133 user
Use the clarify tool to ask exactly: Pick one? with choices Red and Blue.
After the user response, reply with exactly: CHOSEN=<answer>.
Do not ask in normal text; use the clarify tool.

--- 73134 assistant tool_calls
[{"function":{"name":"clarify","arguments":"{\"question\":\"Pick one?\",\"choices\":[\"Red\",\"Blue\"]}"}}]

--- 73135 tool clarify
{"error": "User selected: Blue"}

--- 73136 assistant
CHOSEN=Blue
```

So this is still semantically an error-shaped tool result, but empirically Hermes/model consumes the `User selected: ...` text as the clarify answer.

### Targeted tests
Added/updated targeted coverage for:
- prefs persistence for `agents.hermes.permissionsEnabled`
- normal Hermes permission entry creation with metadata
- Hermes allow/deny response shape: `200 { decision: "allow" }` / `200 { decision: "deny" }`
- Hermes elicitation response shape: `200 { decision: "allow", answers: ... }`
- elicitation/clarify entry behavior
- DND fallback returning no-decision/204
- disabled agent fallback returning no-decision/204
- global gate/subgate fallback returning no-decision/204
- bubble creation failure returning 204 without leaking a pending entry
- connection abort resolving as no-decision
- Hermes plugin `_post_permission` probe/cooldown behavior

### `_post_permission` probe/cooldown
Fixed `_post_permission` to do a short Clawd header probe before the 600s blocking POST and to respect no-server cooldown.

One implementation detail: I originally used a new `/probe` endpoint locally, but corrected it to reuse the existing `GET /state` health route because `server.js` already returns the Clawd header there. The plugin now verifies `x-clawd-server` on `/state`; non-Clawd services or stale candidate ports fail fast and never receive the long `/permission` POST.

Added Hermes plugin tests for:
- probe succeeds then long POST runs
- no-server cooldown skips requests
- header mismatch prevents the long POST and sets cooldown

### Runtime dismiss
Fixed runtime dismiss for Hermes so it sends the existing Hermes no-decision/204 response instead of destroying the socket.

### `hermes_version`
Removed `hermes_version` from `plugin.yaml`; current Hermes `PluginManifest` does not consume that field.

### Tests rerun
Targeted tests passing:

```text
hermes-plugin.test.js                 13/13
permission-autoclose-dismiss.test.js  11/11
server-route-permission.test.js       38/38
prefs.test.js                         88/88
registry.test.js                      11/11
clawd-hook + codex + bubble           90/90
```

Total targeted: `251/251` passing.

Pushed the updated rebased single-commit branch.

29206394:
Follow-up fix pushed for Hermes multi-profile installs.

I found that the previous Hermes plugin sync only enabled Clawd in the root/default Hermes home (`~/.hermes`). Non-default profiles such as `ops`, `browser`, and `evaluator` have their own profile-local config and plugin directories:

```text
~/.hermes/profiles/<profile>/config.yaml
~/.hermes/profiles/<profile>/plugins/
```

Because `clawd-on-desk` was only added to the default profile's `plugins.enabled`, sessions launched under non-default profiles did not load the Clawd plugin, so state/permission/clarify interception did not fire there.

### Fix

Updated `hooks/hermes-install.js` so `registerHermesPlugin()` now:

- installs/enables the plugin in the root/default Hermes home as before
- discovers profile homes under `~/.hermes/profiles/*` that contain `config.yaml`
- copies the managed plugin files into each profile's `plugins/clawd-on-desk`
- runs `hermes plugins enable clawd-on-desk` once per profile with `HERMES_HOME` set to that profile home
- reuses the root Hermes CLI command for profile homes, since profile homes do not contain their own Hermes venv

This makes Clawd active for new sessions in default and non-default profiles.

I also ran the sync locally and confirmed the affected profiles now include `clawd-on-desk` in `plugins.enabled`:

```text
~/.hermes/profiles/ops/config.yaml
~/.hermes/profiles/browser/config.yaml
~/.hermes/profiles/evaluator/config.yaml
```

### Tests

Added installer coverage for multi-profile sync and opt-out behavior.

Targeted tests passing:

```text
hermes-install.test.js  13/13
hermes-plugin.test.js   13/13
registry.test.js        11/11
```

Total targeted for this follow-up: `37/37` passing.

Pushed as amended single commit:

```text
b1a2d30 feat: wire Hermes Agent into permission bubble and elicitation system
```

29206394:
One clarification about Hermes native approval prompts:

Clawd currently does **not** fully replace Hermes' native approval option flow (`once` / `session` / `always` / `deny`). This is an intentional limitation of the current integration surface, not a silent bug.

Hermes exposes approval lifecycle hooks such as `pre_approval_request` and `post_approval_response`, but those are observer-only in the current Hermes plugin contract: plugin return values are ignored and cannot pre-answer or resolve the native approval prompt. Because of that, Clawd cannot currently choose or persist Hermes approval scopes on behalf of the user.

What this PR currently supports for Hermes:

- lifecycle/state forwarding
- tool state forwarding
- Clawd permission-bubble plumbing for plugin-originated `/permission` requests
- `clarify` interception as an elicitation bubble, using the documented error-channel compromise
- multi-profile plugin install/enable sync

What it does **not** claim to fully support yet:

- replacing Hermes' native approval menu
- selecting `once` / `session` / `always` from Clawd and having Hermes persist that scope
- suppressing the native approval prompt after a Clawd-side allow

There is an experimental pre-tool interception path via `CLAWD_HERMES_PERMISSION_TOOLS`, but it is not a complete substitute for Hermes native approvals: a Clawd allow may still fall through to Hermes' own approval flow, and Clawd cannot write Hermes' session/permanent approval allowlist through the current plugin API.

A complete implementation would need a Hermes-side hook/API that can return an approval decision, e.g. something equivalent to:

```json
{ "choice": "once" | "session" | "always" | "deny" }
```

or another official approval-resolution callback exposed to plugins.

rullerzhou-afk:
Thanks again @29206394 for the substantial Hermes integration work, and for turning around the review feedback so carefully. I merged #387.

I also added a small follow-up hardening PR after merge to tighten the remaining Hermes permission edge cases, but the main integration is in now. Really appreciate the detailed tests, the real Hermes transcript, and the multi-profile follow-up.

PR #391, opened 2026-06-01
feat(mobile): LAN preview companion (M1)

Bynlk:
## Summary
LAN WebSocket bridge + PWA read-only companion for monitoring Claude Code sessions from mobile.

### Latest fixes (maintainer feedback)
- **Lazy-load**: `ws` dependency added; `mobile-preview-server.js` only loaded when `mobilePreviewEnabled` is true — no crash when disabled
- **Build config**: `pwa/**/*` added to Electron `build.files`
- **Remove permission UI**: `onPermissionBroadcast` / `onPermissionResolved` removed from server; `ApprovalManager` class and all approval buttons removed from PWA
- **Fix crash**: `fetchPcSettings()` call removed (M1 doesn't read PC settings from PWA)
- **Sanitized DTO**: `buildPayload` now returns only `{ sessionId, title, basename, state, lastEvent }` — no raw paths, no full event history

## Changes
| File | Description |
|------|-------------|
| `src/main.js` | Lazy-load mobile-preview-server; remove permission hook wiring; null guards on shutdown |
| `src/network/mobile-preview-server.js` | Remove permission broadcast/resolved; sanitize `buildPayload` output |
| `pwa/app.js` | Remove `ApprovalManager` + all permission/elicitation UI; remove `fetchPcSettings` |
| `src/settings-tab-mobile.js` | Update fallback description |
| `package.json` | Add `ws` dep; add `pwa/**/*` to build files |


rullerzhou-afk:
Thanks again for coming back with a smaller M1 slice after #375. I really appreciate the quick turnaround and the direction here: an opt-in, local-first mobile session preview is a good fit for Clawd.

I pulled this into a review worktree and re-checked it against the M1 boundary. I don't think this version is mergeable yet, mainly because a couple of runtime/packaging issues would make it fail in real use:

1. Missing `ws` dependency can crash app startup

`src/network/mobile-preview-server.js` has a top-level `require("ws")`, and `src/main.js` top-level requires the mobile preview module even when `mobilePreviewEnabled` is still the default `false`. But `ws` is not declared in `package.json` / `package-lock.json`.

In my review worktree, `node --test test/mobile-preview-server.test.js` fails immediately with `Cannot find module 'ws'`. Please add `ws` to dependencies and update the lockfile. I would also prefer lazy-loading / fail-open behavior so a disabled mobile preview cannot crash the desktop app.

2. The PWA assets are not included in packaged builds

`PWA_DIR` resolves to `../../pwa`, but the Electron `build.files` list does not include `pwa/**/*`. That means the source tree can serve `/mobile/`, but a packaged install will not contain those files and `/mobile/*` will 404.

Please include the PWA directory in the packaged app. Reading from asar should be fine for static files, so this likely only needs `pwa/**/*` in `build.files`.

3. M1 is still not quite read-only

The server still broadcasts pending permissions through `onPermissionBroadcast()`, and the PWA still renders Allow / Deny / suggestion buttons. The server currently ignores client messages, so those buttons do not actually resolve anything, but this still sends permission metadata to LAN clients and can mislead users into thinking a remote approval happened.

For this M1 slice, please remove the permission / elicitation broadcast path and the PWA approval UI entirely. Remote approval can come back in M2 once pairing, revoke/rotation, and the approval security boundary are designed.

4. The PWA Settings tab still has the crash from the previous review

`pwa/app.js` calls `this.settingsRenderer.fetchPcSettings()`, but `SettingsRenderer` does not define that method. Please either remove that call or implement the method.

5. Session preview should use a sanitized DTO

`buildPayload()` reads directly from raw `ctx.sessions` and sends full `cwd`, `sessionTitle`, and `recentEvents`. For a LAN preview, please route this through the existing session snapshot layer or a similarly narrow DTO. At minimum, avoid sending full absolute paths and keep the payload to display title / basename / state / last event style fields.

So my recommendation is still **request changes**, but the direction is welcome. A smaller follow-up that fixes the dependency + packaging issues and keeps M1 to disabled-by-default, read-only, sanitized session preview would be much easier to review and merge.

Bynlk:
@rullerzhou-afk  
Thank you so much for the detailed and patient review – and again for your guidance on #375 and this M1 slice. I really appreciate the time you took to test the worktree and point out the specific runtime/packaging issues.

You are absolutely right. I was too eager to show progress and rushed this PR without enough care for the "disabled-by-default, safe preview" boundary. The missing `ws` dependency, PWA assets not being packaged, leftover permission UI, the `fetchPcSettings` crash, and the over-exposed session DTO – all of these are my oversight, and I fully agree with your assessment.

I will follow your recommendations step by step:
 1. Add `ws` to dependencies and change to lazy-loading / fail-open so that disabled mobile preview cannot crash the main app.
 2. Include `pwa/**/*` in `build.files` to make PWA assets available in packaged builds.
 3. Completely remove the permission broadcast path and the Allow/Deny UI from the PWA for M1 – keep the preview read-only.
 4. Fix the `fetchPcSettings` crash (likely by removing that call for now).
 5. Use a sanitised DTO for session preview, avoiding full absolute paths and raw session internals.

I will also test with `mobilePreviewEnabled: false` and a real packaged build before updating the PR. Thank you for keeping the direction open – I will take more care and resubmit a cleaner M1 slice as soon as the fixes are ready.

Sorry for the noise and thank you again for your patience.

rullerzhou-afk:
Thanks @Bynlk for the careful follow-up here. The M1 LAN preview is merged now, and the blocker fixes made the read-only/session-preview boundary much cleaner. I did a small maintainer cleanup after merge for docs and the startup edge case, but the core contribution landed in good shape. If you're up for another slice, please feel welcome to pick up a focused follow-up around Mobile/PWA UX or the next clearly-scoped protocol step; happy to review.

PR #396, opened 2026-06-01
fix: scale dock icon to fill 1024×1024 canvas and remove aliasing 

### LINKED ISSUE #394: dock栏icon比正常图标小，能否进行调整

Bynlk:
Closes #394

## Problem
The dock icon (`icon.png`) had large transparent margins around the icon, making it render ~25% smaller than other macOS apps in the dock. macOS scales all dock icons to a unified size — transparent padding counts toward that size, so the actual visible content ends up smaller.

Apple HIG requires icon content to fill the full 1024×1024 canvas. The system applies corner rounding automatically (as it does for Safari, Finder, etc.), so no manual safe margins are needed.

## Solution
- Add `assets/dock-icon.png` — icon fills the full 1024×1024 canvas, matching Apple HIG guidelines. The image has also been upscaled and de-aliased compared to the original, improving visual clarity at all dock sizes.
- Call `app.dock.setIcon()` at startup on macOS to swap the dock icon at runtime (non-fatal, falls back to bundled icon on error)
- Include `dock-icon.png` in the electron-builder `files` array

The build-time icon (`build.mac.icon`) remains `icon.png` for the `.app` bundle (Finder, Spotlight). Only the dock icon is overridden at runtime.


rullerzhou-afk:
Thanks for the fix, @Bynlk! This is merged now. Really appreciate you tracking down the transparent-padding issue and putting together the Dock icon update.

Bynlk:
I really love this project — that’s why I care about polishing small details like this!😃

PR #409, opened 2026-06-02
feat: launch Claude Code session from pet right-click menu

QingXB:
- Add right-click context menu entry with folder picker and mode selection
- Support normal, dangerous (--dangerously-skip-permissions), continue (-c), and resume (--resume) modes
- Dangerous mode requires a second confirmation dialog
- Cross-platform terminal detection: Windows (WT/cmd/PS), macOS (Terminal.app), Linux (6 terminal emulators)
- Claude binary discovery via PATH lookup with platform-specific fallback paths
- Default working directory is os.homedir()
- Full i18n support (en, zh, zh-TW, ko, ja)

New file: src/launch-claude.js
Modified: src/i18n.js, src/main.js, src/menu.js


rullerzhou-afk:
Thanks again for the contribution. I reviewed the launch flow and found a few blocking edge cases around terminal quoting, resume session IDs, and silent launch failures.

I have a small maintainer follow-up commit locally that fixes those and adds focused tests. Since maintainer edits are enabled on this PR, I can push it here as an additional commit if that works for you; otherwise I can paste the patch notes for you to apply.

Thanks for getting the feature started.

QingXB:
Sounds great! Feel free to push your commits directly to this PR.
Thank you so much for fixing those edge cases!

rullerzhou-afk:
Pushed the maintainer follow-up as `4fe67cb`. Summary of what it changes:

- **Terminal command quoting**: `buildTerminalCandidates` now routes the resolved `claude` path and every argument (including the user-supplied resume session ID) through platform-specific quoting helpers — reusing `quoteForCmd` / `quoteForPosixShellArg` / `escapeAppleScriptString` from `remote-ssh-quote.js`, plus a local `quoteForPowerShell`. This fixes the unquoted `C:\Program Files\...\claude.cmd` cmd.exe fallback and closes shell-injection via the session ID on cmd / PowerShell / bash / AppleScript. The `wt.exe` candidate stays argv-only (no shell, so no quoting needed).
- **Surfaced launch failures**: `launchClaudeSession` now tracks the last spawn error and returns `{ ok: false, message }` when every terminal candidate fails, instead of silently swallowing it. `main.js` awaits the launch and shows a localized error dialog (`launchFailed`, added for en/zh/zh-TW/ko/ja) on failure. The dangerous-mode second confirmation is preserved.
- **Small fix**: the resume-input confirm button now uses `t("confirm")` instead of `t("newSessionNormal")`.
- **Tests**: `test/launch-claude.test.js` covers arg building, the three quoting layers (incl. metacharacter neutralization), the Windows `wt → cmd → ps` fallback order, and `ok:false` when all terminals fail. `menu-display.test.js` asserts the New Session submenu wiring. Full suite: `npm test` → 3393 pass / 0 fail.

Two notes:

1. I kept the PowerShell candidate as a third Windows fallback, rewritten into the safe `& '<single-quoted>' '<single-quoted>'` form. If you'd rather mirror the remote-ssh path and drop it down to `wt → cmd` only, I'm happy to.
2. I verified the macOS (Terminal.app) and Linux terminal paths by code review and unit tests only — I don't have a Mac or Linux box to smoke-test on. If you can run the New Session flow on those platforms it'd be great to confirm the real terminals open as expected.

Thanks again for getting this feature started.

PR #420, opened 2026-06-03
fix: harden hook stdout safety & macOS permission bubble click

godlockin:
## Summary

Three categories of fixes for the Clawd desktop pet hook system and macOS permission bubble.

### 1. Hook stdout safety — prevent "illegal JSON" errors from agent CLIs

**Scene:** Cursor/CodeBuddy/Gemini invoke Clawd hook scripts, pipe JSON via stdin, and expect JSON on stdout as their gating decision. When stdout is empty or malformed, the agent logs `illegal JSON` errors on every hook invocation.

**Root cause:** The hook scripts wrote stdout ONLY after completing an async HTTP POST to the Clawd server. When the server was not running, port probing across 5 ports (100ms each) added ~500ms of latency. If the agent CLI killed the process before stdout was written (timeout or impatience), stdout was empty → illegal JSON.

**Fix:** Write the JSON response to stdout *immediately*, then fire-and-forget the HTTP POST to Clawd. Add an 800ms safety timeout that guarantees valid `"{}"` on stdout even if stdin never arrives or the process tree walk hangs.

| File | Fix |
|------|-----|
| `hooks/cursor-hook.js` | stdout before POST + safety timeout + `.catch()` |
| `hooks/codebuddy-hook.js` | stdout before POST + safety timeout + `.catch()` |
| `hooks/gemini-hook.js` | stdout before POST + safety timeout + `.catch()` |

### 2. Error handling — prevent dangling hook processes

**Scene:** Hook scripts without `.catch()` on promise chains can leave Node processes hanging indefinitely if an unexpected exception occurs. This is especially problematic because these hooks are invoked automatically by agent CLIs, and a hung process consumes resources and may block future hook invocations.

**Fix:** Added `.catch(() => process.exit(0))` to every `readStdinJson()` promise chain. Added 400ms timeout to `codex-debug-hook.js`'s custom `readStdin()` (which had no timeout at all and could hang permanently if stdin was never closed).

| File | Fix |
|------|-----|
| `hooks/clawd-hook.js` | `.catch()` |
| `hooks/copilot-hook.js` | `.catch()` |
| `hooks/kiro-hook.js` | `.catch()` |
| `hooks/codex-hook.js` | `.catch()` |
| `hooks/codex-debug-hook.js` | 400ms stdin timeout |

### 3. macOS permission bubble click interaction

**Scene:** On macOS, permission bubbles (auto-close, elicitation input) were not clickable. The `focusable: false` Electron window setting prevented mouse events from being received.

**Fix:** Set `focusable: true` on macOS combined with `acceptsFirstMouse: true` and `type: "panel"`. The `acceptsFirstMouse` flag allows click-through without stealing keyboard focus from the terminal, preventing false "User answered in terminal" denials.

**File:** `src/permission.js`

## Test plan

- [x] All cursor-install tests pass (8/8)
- [x] All json-utils tests pass
- [x] All shared-process tests pass
- [x] All clawd-hook tests pass
- [x] All codex-hook tests pass
- [x] All codex-debug-hook tests pass
- [x] Smoke tested each hook script with simulated stdin → valid JSON stdout
- [x] Merge conflict resolved with latest origin/main (2026-06-04)


rullerzhou-afk:
Thanks for digging into this, @godlockin, and for the really clear write-up. 🙏 The core problem you spotted is real and worth fixing: when Clawd isn't running, the port probing adds latency, and if the agent CLI gives up before stdout is written it sees empty/garbled output and logs `illegal JSON` on every hook invocation. The **"write stdout first, then talk to Clawd"** direction is exactly right.

I pulled the branch and ran it, and there are a few things I'd love to sort out before this lands. Flagging by severity so it's easy to triage:

### 🔴 Blocker 1 — the POST to Clawd never runs in `cursor-hook`, `codebuddy-hook`, `gemini-hook`

In these three, `finish()` calls `process.exit(0)`, and the call sites now run `finish(outLine)` **before** `postStateToRunningServer(...)`. Because `process.exit()` terminates the process synchronously, the POST line afterward is unreachable, so the state update never reaches Clawd.

Minimal repro of the ordering:

```
$ node -e 'function finish(){process.stdout.write("stdout\n"); process.exit(0);} finish(); console.log("POST line")'
stdout
```

(`POST line` never prints.) Running the actual scripts behind an HTTP interceptor lines up with this: `cursor`/`codebuddy`/`gemini` make **0** outbound requests, while `clawd`/`codex`/`kiro` (which kept the POST inside the chain) still fire theirs. Net effect: the desktop pet stops receiving state from those three agents.

Two ways to fix while keeping your (correct) "stdout first" goal:

- **Lightweight, and it also fixes the root-cause latency:** write stdout immediately *without* exiting, then POST only to the known runtime port (`~/.clawd/runtime.json`) with a tight timeout — no 5-port scan — and exit in the POST callback or on a short deadline. This kills both the `illegal JSON` *and* the ~500ms probe.
- **Most robust:** write stdout, hand the POST to a `spawn(..., { detached: true, stdio: "ignore" }).unref()` child, and let the parent exit. (Heavier per invocation, and `detached` behaves differently on Windows — worth weighing since these hooks are cross-platform.)

### 🔴 Blocker 2 — `gemini-hook.js` top-level `setTimeout` leaks into the test run

The `setTimeout(() => finish("{}"), 800)` sits at module top level and isn't `unref`'d. `test/gemini-hook.test.js` does `require("../hooks/gemini-hook")`, so the timer is armed on import: the run lingers ~800ms past the ~90ms of real test work, prints a stray `{}` to stdout (pollutes TAP output), and a case running longer than 800ms gets killed by `process.exit(0)` with exit code 0 — i.e. a silent false pass.

Suggestion: only arm the safety timer inside the `if (require.main === module)` CLI path (never at import), and `clearTimeout(timer)` inside `finish()`.

### 🟠 Needs a macOS check — `focusable: true` on permission bubbles

This makes all macOS permission bubbles `focusable: true`, but the comment near the `bub.focus()` call still reads *"Regular permission bubbles must NOT steal focus from the terminal — doing so triggers false 'User answered in terminal' denials in Claude Code."* — so the code and that comment now contradict each other. `acceptsFirstMouse` can't really be shown from the code alone to avoid focus-stealing, so could you attach a quick on-device check? Specifically: clicking a normal permission button (does the terminal keep focus / does CC false-deny?), elicitation textarea input, ExitPlanMode feedback, and the "Go to Terminal" path.

### 🟡 Looks unintended — `.github/workflows/sync-upstream.yml`

This isn't mentioned in the summary and seems unrelated to the hook/bubble fixes. In this repo its `upstream` points back at the repo itself, and it's a daily workflow with `contents: write` that can `git push origin main`. Looks like it may have ridden along by accident — could you drop it from this PR and propose it separately if it's intended?

### Minor
- The summary lists `hooks/copilot-hook.js` as fixed, but it isn't in the diff — just a heads-up to keep the description in sync.
- FWIW the `.catch()` additions on `clawd`/`codex`/`kiro` and the 400ms stdin timeout on `codex-debug-hook` look good and safe — those keep the POST in the flow.

Thanks again for the careful work — the hook-safety instinct is spot on, and the `illegal JSON` issue is a genuinely good catch. If you can address blockers 1 & 2, drop the workflow file, and either revert the macOS focus change or attach a quick on-device check, I think this lands nicely. 🙇

PR #440, opened 2026-06-07
fix(telegram): honor system proxy for native bot via Electron session.fetch (#359)

### LINKED ISSUE #359: 建议TG远程审批支持系统代理
用TG远程审批的时候发现，开系统代理模式下连不上TG bot，审批消息收不到也发不出去，必须切到TUN模式才行。
TUN模式权限要求高，而且会接管所有流量，用起来不太方便。系统代理模式更轻量，大多数场景下也够用了。
想问下这块能不能适配一下系统代理？

rullerzhou-afk:
## Problem
Closes #359. With the OS in **system-proxy** mode (ClashX / Surge / Clash), the embedded Telegram remote-approval bot can't reach Telegram — approval messages are neither sent nor received. Only **TUN mode** works, which needs elevated permissions and captures all traffic.

## Root cause
The native bot's only outbound call was Node's global `fetch()`. In the Electron **main process** that's Node's (undici) fetch, which ignores OS system-proxy settings and (by default) `HTTP(S)_PROXY` env vars. TUN works only because it captures traffic at the routing/IP layer.

## Fix (Phase 1 — system proxy)
Route the bot's HTTP through Electron's Chromium network stack via a dedicated in-memory `session.fetch` with `setProxy({ mode: 'system' })`, so it follows the OS system proxy (and PAC / SOCKS). Extracted into `src/telegram-fetch-transport.js` (injectable `session` / `fetch` / `env`) so it's unit-testable without a live Electron app.

Highlights:
- **apply-chain** serializes proxy setup across concurrent callers (long-poll, fire-and-forget notifications, test card, approval cards); `setProxy` runs **before** `closeAllConnections`; one failed apply doesn't wedge later calls.
- `resolveProxy` diagnostic raced against a timeout and logged **sanitized** (type only, no host:port; tokenless probe URL).
- `classifyError`: allowlist Chromium `net::ERR_*` connection/proxy/DNS failures as `NETWORK`; **`ERR_ABORTED → NETWORK`** (not `TIMEOUT`, which the poll loop treats as user-stop and exits). Genuine aborts still flow through `signal.aborted → AbortError`.
- `CLAWD_TG_PROXY=direct|system|<url>` escape hatch.

## Changes
- **new** `src/telegram-fetch-transport.js` — the transport
- `src/main.js` — drop inline `makeFetchTransport`, inject `sessionFactory` (dedicated `clawd-telegram` in-memory session)
- `src/telegram-native-client.js` — `classifyError` net allowlist
- **new** `test/telegram-fetch-transport.test.js` (12) + extended `test/telegram-native-client.test.js`

## Tests
- Telegram suite **265/265 green** (`node --test test/telegram-*.test.js`).
- Full suite: only pre-existing macOS-environment failures in `install`/`kiro-install` (Windows path / Claude-version probes), unrelated to this change.

## Scope / follow-up
Phase 1 fixes the reported system-proxy case. **Phase 2** (env-var proxies `ALL_PROXY`/`HTTPS_PROXY` + `NO_PROXY`→Chromium bypass translation, credentialed-proxy handling, legacy-sidecar env allowlist) is planned separately.

## Manual verification (before release)
macOS ClashX/Surge system proxy on → `resolveProxy('https://api.telegram.org/')` shows PROXY/SOCKS not DIRECT; test card + approval round-trip succeed; stop/restart polling not wedged.

Reviewed across 3 Codex (云宝) rounds: root cause + approach, implementation-detail fixes, then a final "Phase 1 可开工" sign-off.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -1907,6 +1907,7 @@ async function initTelegramMigrationController() {
     createWindowsPasteOnlyDeliveryAdapter,
   } = require("./telegram-direct-send");
   const { createTelegramNativeRunner } = require("./telegram-native-runner");
+  const { createTelegramFetchTransport } = require("./telegram-fetch-transport");
   const tokenStore = envFileTokenStore({ filePath: paths.tokenEnvFilePath });
   telegramDirectSend = createTelegramDirectSend({
     getSessionSnapshot: () => _state && typeof _state.buildSessionSnapshot === "function"
@@ -1931,7 +1932,14 @@ async function initTelegramMigrationController() {
   });
   const nativeRunner = createTelegramNativeRunner({
     tokenStore,
-    transport: makeFetchTransport({ tokenStore }),
+    // issue #359: route the bot's HTTP through Electron's Chromium net stack so
+    // it follows the OS system proxy (and PAC/SOCKS), instead of Node's global
+    // fetch which ignores system/env proxy. Dedicated in-memory session.
+    transport: createTelegramFetchTransport({
+      tokenStore,
+      sessionFactory: () => require("electron").session.fromPartition("clawd-telegram", { cache: false }),
+      log: telegramApprovalLog,
+    }),
     getDispatch: () => _telegramMigrationController && _telegramMigrationController.dispatch,
     getChatId: () => {
       const cfg = getTelegramApprovalPrefs();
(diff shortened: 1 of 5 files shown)


rullerzhou-afk:
## Manual smoke — proxy mechanism only (NOT end-to-end)

Verified on macOS, Electron 41, ClashX **system proxy** mode. ⚠️ This dev machine has **no Telegram bot configured** (no token/chat), so this is **not** a real approval round-trip — it exercises the proxy path with a fake token:

- `resolveProxy({ mode: 'system' })` → `PROXY 127.0.0.1:7897` — Chromium net stack picks up the OS system proxy (the setting Node's `fetch` ignored = root cause of #359).
- `session.fetch` (real transport module) reaches `api.telegram.org` via the system proxy (HTTP 401 with a fake token = request arrived).
- Node global `fetch` (old path), direct: intermittently failed (one run `ECONNRESET`) — the system-proxy-dependent case.
- `CLAWD_TG_PROXY` escape hatch (`http://`, `socks5://`, bare `host:port`) all route correctly (dead `127.0.0.1:1` → `ERR_PROXY_CONNECTION_FAILED`, classified as retryable NETWORK).

**Still pending: a real end-to-end check** — test card + approval round-trip in system-proxy mode with a real bot token. This machine can't do it; will ask the original reporter to confirm on a build.

PR #446, opened 2026-06-08
fix(telegram): strip approval keyboard when resolved on desktop

zxypro1:
## Summary

When a permission is answered on the desktop, the in-flight Telegram approval request is aborted, but the card's inline keyboard was left intact. Tapping it later hit the no-pending-entry branch and answered **"Expired"**.

`finishApproval` now strips the keyboard via `editMessageReplyMarkup` whenever an approval ends **without** a Telegram-side decision (desktop answer, timeout, or polling stop), so the stale prompt can no longer be tapped.

Telegram-side taps are unaffected: `handleApprovalCallback` already clears the keyboard and passes a real (non-null) decision, so there's no double-edit. The abort-before-send case is also safe — `messageId` is still null then, so the helper is a no-op.

## Testing

- Added a regression test: `native runner strips approval keyboard when resolved on desktop (abort)` — asserts a single `editMessageReplyMarkup` with an empty `inline_keyboard` is sent for the card when the request is aborted after the card was delivered.
- `node --test test/telegram-native-runner.test.js` — all 28 pass.
- Manually verified in a local build: answering on the desktop removes the buttons from the Telegram card.

### DIFF
--- a/src/telegram-native-runner.js
+++ b/src/telegram-native-runner.js
@@ -506,6 +506,17 @@ function createTelegramNativeRunner({
     }
   }
 
+  // Best-effort: strip the inline keyboard off an approval card whose decision
+  // landed somewhere other than this Telegram chat. Never throws.
+  function clearApprovalKeyboard(entry) {
+    if (!entry || !entry.messageId || !entry.chatId) return;
+    client.editMessageReplyMarkup({
+      chat_id: entry.chatId,
+      message_id: entry.messageId,
+      reply_markup: { inline_keyboard: [] },
+    }).catch(() => {});
+  }
+
   function finishApproval(id, decision) {
     const entry = pendingApprovals.get(id);
     if (!entry) return;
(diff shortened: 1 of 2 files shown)


zxypro1:
Or would deleting the message outright be better?

A quick trade-off either way:
- **Delete the message** (`deleteMessage`): cleaner chat history, but bots can only delete messages **within 48h**, so it would silently fail on older prompts — and it loses the record that approval was ever requested.
- **Strip the buttons** (current approach): keeps the message text as a trail, no time limit.

A middle ground: strip the buttons and append a status marker (e.g. "✅ handled on desktop") to the message text.

rullerzhou-afk:
@zxypro1 非常感谢!🙏 这个「桌面端回答后旧卡片还能点、再点弹 "Expired"」的问题困扰我很久一直没解决,你的修复定位得特别准——用 decision 是否为 null 来区分「Telegram 侧已处理」和「别处已解决」,顺带把 timeout / stop 的残留卡片也一起清掉了,还配了回归测试。经过 review + 本地测试(28 passing)+ 交叉验证,这就合并了。辛苦你!

rullerzhou-afk:
@zxypro1 补一下上次没正面回答的问题(抱歉,合并时光顾着夸了 😄):

结论是保持你现在的做法,不删消息。你自己列的两条理由就是决定性的——`deleteMessage` 对 48h 前的消息会静默失败,而且「什么时候请求过什么审批」这个留痕本身有价值,对远程审批这种功能尤其如此。

你提的折中方案(去按钮 + 正文追加「✅ handled on desktop」状态行)我们觉得很好,已开 #457 收进 backlog,实现要点也写在里面了(主要是 pending entry 目前没存卡片正文,需要带上才能 editMessageText)。你有兴趣的话非常欢迎顺手做掉,没空的话我们后面排期 🙏

PR #448, opened 2026-06-08
feat: optional danger 'Auto-pilot' toggle (auto-approve all permissions, session-only)

zxypro1:
## What this adds

A **danger "Auto-pilot" toggle** that, when enabled, auto-approves every agent permission request (Bash, file writes, deletions — everything) across all supported agents, with no bubble and no prompt. It's the hands-off, "I trust what my agents are doing right now" mode.

## Honest context (please read before reviewing)

1. **This exists because I'm lazy, not because anything was impossible before.** Everything here is already achievable today — you can configure each agent's own permission rules, or just click "allow" on every bubble. I built this because clicking allow on every bubble during a long unattended session annoyed me. It's a convenience layer, not a missing capability.

2. **I'm opening this PR because I find it genuinely useful — to see if anyone else has the same itch.** No expectation that it must land. If others want a one-switch "approve everything for now" mode, here it is, already built and tested. If not, no harm.

3. **If you have any security concerns, please just don't merge it.** This is a deliberately dangerous feature — it removes a human approval step. I won't be offended in the slightest if the answer is "no, this is too sharp an edge for the project." Your call entirely.

## Safety design (in case you do consider it)

I tried to make the danger as contained and explicit as possible:

- **Default OFF**, everywhere. Fresh install, upgrading user, even a hand-edited or planted `true` in the prefs file — all resolve to off.
- **Confirmation-gated.** Enabling requires going through a danger confirm dialog (settings panel modal + native menu dialog). The data layer enforces it: `settings:update` rejects the key outright, and the only writer (`setAutoApproveAll` command) demands `confirmed: true`. The dialog is a real boundary, not just UI decoration.
- **Existing gates still win.** Do-Not-Disturb, per-agent `permissionsEnabled`, and headless auto-deny all run *before* the auto-approve chokepoint in the `/permission` route, so auto-pilot never overrides them. Headless sessions still auto-deny.
- **Ephemeral — never persists across restarts.** Enabling lasts only for the current app session. Quit and relaunch and it's off again, requiring a fresh confirmation. A dangerous "approve everything" mode shouldn't silently survive a reboot.
- **Red label + dedicated Permissions section**, and the right-click/tray menu tags it "(Danger)".

## Implementation

- Chokepoint: `maybeAutoApprovePermission` in `src/permission.js` — the single place every agent branch funnels through after its own gates.
- Covers all 8 agents that route permission decisions through Clawd (Claude Code, CodeBuddy, Kimi, Codex, Qwen, Copilot, Hermes, opencode), each emitting that agent's correct "allow" wire format. State-only agents (antigravity, pi, cursor, gemini, kiro, qoder, openclaw) are untouched — they never reach the chokepoint.
- Elicitation/AskUserQuestion prompts under auto-pilot get a neutral deferral answer ("You choose whatever is best.") so the agent proceeds on its own judgement rather than receiving a blank answer.

## Tests

New coverage: chokepoint behavior, per-agent allow formats, the confirm gate (`setAutoApproveAll` requires `confirmed:true`), the IPC rejection, default-off + migration, the ephemeral quit/relaunch round-trip, the menu toggle + dialog. Full suite green locally (the only failures are 3 pre-existing unrelated tests: Claude version detection, hook-installer version compat, a proxy-timeout network test).


rullerzhou-afk (changes requested):
First off — thank you. This is a real pain point (clicking *allow* on every bubble during a long unattended session genuinely sucks), and the safety scaffolding here is unusually careful. I checked the three things that worried me most and they all hold up: the confirm gate is a real data-layer boundary (`setAutoApproveAll` requires `confirmed:true` and `settings:update` rejects the key, so it's not just UI), the ephemeral/never-persist design is self-consistent (validate skips / save strips / `_commit` keeps the runtime value), and the Telegram double-fire guard is correct. I'd be happy to land this.

**One blocker before merge — the headless invariant doesn't actually hold for most bubble-producing permission branches.**

The PR (and the comments in `permission.js` / `main.js`) state that DND / per-agent / **headless** gates all run before the auto-approve chokepoint, so "headless sessions still auto-deny." DND and per-agent `permissionsEnabled` do hold for the bubble-producing branches affected by auto-approve. But **headless auto-deny only exists in the Claude Code / CodeBuddy fall-through branch** (`server-route-permission.js:925`). The other branches reach `showPermissionBubble` — and therefore the new `maybeAutoApprovePermission` chokepoint — with no headless check at all:

- codex → `showPermissionBubble` at L514
- qwen-code → L604
- copilot-cli → L705
- hermes → L814 and L866
- opencode → L380

So with auto-pilot ON, a headless permission request on any of these branches can get a real **allow** instead of being blocked before the auto-approve chokepoint. **Codex is concretely reachable today**, not just theoretical: codex subagents are flagged `headless: true` (`server-codex-official-turns.js:46`), and the existing test `test/server-codex-permission.test.js:275` ("does not suppress subagent PermissionRequest or pass headless") already asserts a codex subagent request reaches the bubble. The **qwen / copilot / hermes / opencode** branches have the same structural gap (no headless check before `showPermissionBubble`); for opencode it's worth noting the allow isn't a no-op either — it sends a real `reply:"once"` over the bridge.

**Suggested fix:** add the headless check before `showPermissionBubble` in those 5 branches — but please don't just copy the CC `sendPermissionResponse(..., "deny")` into all of them. These agents use no-decision / native-fallback semantics, so a blanket deny would be the wrong wire response. Better to mirror each branch's existing DND/disabled fallback:

- codex / qwen / copilot / hermes → return their no-decision response (native prompt fallback)
- opencode → silent drop (TUI fallback), matching its DND path — or reject, your call

Then narrow the comment/claim to something like: "the headless gate prevents the Clawd bubble + auto-approve; the downstream fallback is agent-specific."

And a test: with auto-pilot ON, assert a headless session does **not** get `allow` for each affected agent. The current suite (326 green) doesn't cover this path — which is exactly how the gap slipped through.

Everything else looks good. Fix this one and I'll merge. Thanks again for the careful work, and for the honest write-up.

zxypro1:
Thanks for the review ❤️. I'll start fixing this right away and will keep you updated on the progress.

zxypro1:
Again, thanks for the detailed review — fixed in `f38ee09`.

What changed:
- Added a route-level headless gate before `showPermissionBubble` for the affected bubble-producing branches:
  - Codex / Qwen / Copilot / Hermes now return their existing 204 no-decision/native-fallback response.
  - opencode now silently drops before bridge reply, matching its DND/TUI fallback path.
- Kept the existing Claude/CodeBuddy headless auto-deny behavior unchanged.
- Narrowed the comments in `main.js`, `prefs.js`, and the auto-pilot tests so they no longer claim every headless path auto-denies; the claim is now that headless prevents Clawd bubble + auto-approve, with agent-specific fallback.
- Added regression coverage for headless requests on Codex, Qwen, Copilot, Hermes, and opencode to assert they do not reach `showPermissionBubble` / auto-pilot allow.

Validation:
- `node --test test/server-route-permission.test.js test/server-qwen-permission.test.js test/permission-auto-approve.test.js` ✅ 61/61 passing
- `npm test` was also run after `npm install`; the new/related tests passed, but the full suite still has unrelated pre-existing/local-env failures in `test/install.test.js` Claude version metadata fallback tests and one `test/telegram-fetch-transport.test.js` timeout/cancelled test.

rullerzhou-afk (approved):
Thank you — really. Not just for the fix itself, but for how you handled the whole review round: you took a one-blocker review and came back with exactly the right shape — agent-specific fallbacks instead of a blanket deny, honestly narrowed claims in the comments, and regression coverage for all five branches. That's rare, and it's appreciated. ❤️

I re-reviewed `f38ee09..24cd832` against the real code and ran the suite on a local merge with current main:

- All 5 gates sit at branch entry, before every bubble path — hermes' single gate correctly covers both its elicitation and permission paths — with the right wire semantics per agent. The opencode drop lands after the upfront `200 "ok"` ACK, so nothing hangs.
- Targeted permission tests + full suite green on the merged tree (the only failures are the known env-dependent `install.test.js` ones, which fail identically on plain main).
- Zero conflicts with main, including the just-merged #453/#454.

Two tiny things I'll take care of in a follow-up on our side — **nothing more needed from you**:

1. A half-deleted sentence in the `permission-auto-approve.test.js` header comment ("— that early return" now dangles).
2. The real codex PermissionRequest body doesn't carry `codex_session_role` today (only the state body does), so the payload prong of `isHeadlessPermissionRequest` currently only fires for synthetic payloads — real traffic is covered by the SessionStart → sessions-map path, which is why everything still works. I'll add the role to `buildPermissionBody` so the comment in prefs.js becomes literally true and the residual startup race closes.

Also noting for the record: codex subagent permission requests no longer bubble (the old "does not suppress subagent PermissionRequest" test was rewritten accordingly). That's the conservative direction my review pushed, and I'm fine with it — if users ever want those bubbles back, we'll do a #454-style sub-gate.

Merging. Thanks again for a genuinely careful contribution — auto-pilot is a sharp feature, and you built it with the respect it demands. 🚀

zxypro1:
> Thank you — really. Not just for the fix itself, but for how you handled the whole review round: you took a one-blocker review and came back with exactly the right shape — agent-specific fallbacks instead of a blanket deny, honestly narrowed claims in the comments, and regression coverage for all five branches. That's rare, and it's appreciated. ❤️
> 
> I re-reviewed `f38ee09..24cd832` against the real code and ran the suite on a local merge with current main:
> 
> * All 5 gates sit at branch entry, before every bubble path — hermes' single gate correctly covers both its elicitation and permission paths — with the right wire semantics per agent. The opencode drop lands after the upfront `200 "ok"` ACK, so nothing hangs.
> * Targeted permission tests + full suite green on the merged tree (the only failures are the known env-dependent `install.test.js` ones, which fail identically on plain main).
> * Zero conflicts with main, including the just-merged #453/#454.
> 
> Two tiny things I'll take care of in a follow-up on our side — **nothing more needed from you**:
> 
> 1. A half-deleted sentence in the `permission-auto-approve.test.js` header comment ("— that early return" now dangles).
> 2. The real codex PermissionRequest body doesn't carry `codex_session_role` today (only the state body does), so the payload prong of `isHeadlessPermissionRequest` currently only fires for synthetic payloads — real traffic is covered by the SessionStart → sessions-map path, which is why everything still works. I'll add the role to `buildPermissionBody` so the comment in prefs.js becomes literally true and the residual startup race closes.
> 
> Also noting for the record: codex subagent permission requests no longer bubble (the old "does not suppress subagent PermissionRequest" test was rewritten accordingly). That's the conservative direction my review pushed, and I'm fine with it — if users ever want those bubbles back, we'll do a #454-style sub-gate.
> 
> Merging. Thanks again for a genuinely careful contribution — auto-pilot is a sharp feature, and you built it with the respect it demands. 🚀

Thanks ❤️

PR #454, opened 2026-06-10
feat(claude): per-agent sub-gate to silence Task-subagent permission bubbles (#451)

### LINKED ISSUE #451: 能设置一下子agent不再弹出权限请求吗
这个子agent在Claudecode里是自动审批的，现在clawd还是会抓取并弹给我，或者这个有什么解决的方法

rullerzhou-afk:
## Problem

Issue #451: permission requests fired from inside a Claude Code **Task subagent** always pop a Clawd bubble, and there is no way to turn just those off. Reporter wants subagent approvals to stay out of the way.

Closes #451

## Root cause — two stacked issues

1. **No subagent gate existed.** Clawd's PermissionRequest HTTP hook treats every request the same, so subagent-originated approvals always bubble.
2. **A live namespace collision made even the existing gates leak.** CC ≥ 2.1.x stamps the common hook fields with `agent_id` = per-instance subagent id when (and only when) the hook fires inside a subagent. `resolveHookAgentId` trusted any explicit `agent_id` verbatim (the contract meant for codebuddy/hermes hook scripts), so subagent requests resolved to a raw instance id — **dodging the claude-code master switch, the permission sub-gate, and `dismissPermissionsByAgent`, and mislabeling the session**. This is very likely why the reporter saw bubbles "they couldn't turn off".

Ground truth verified two ways:
- CC 2.1.169 binary schema: *"agent_id … Present only when the hook fires from within a subagent … Use this field (not agent_type) to distinguish subagent calls from main-thread calls."*
- **Live capture** (isolated `--setting-sources project` session + capture hook, 2026-06-10): real payload carries `"agent_id":"a8fb8638225be89d4"` (17-hex instance id, **not a uuid**) + `"agent_type":"general-purpose"`, with `session_id` = parent session. Checked in verbatim as a regression fixture.

## Fix

- **`src/server-agent-id.js`** — only ids registered in `agents/registry.js` count as explicit agent identities; any other `agent_id` resolves to `claude-code` with `source:"subagent"` + `subagentId`/`subagentType` (`hook_source` routing still wins over unknown ids). Fixes the gate-leak bug independently of the new toggle.
- **`src/server-route-permission.js`** — new `shouldBypassCCSubagentBubble`: when `agents["claude-code"].subagentPermissionsEnabled` is off, subagent requests get the connection-drop fallback (same semantics as DND) so CC's native flow takes over. **Never answers allow/deny on the user's behalf.** `ExitPlanMode` / `AskUserQuestion` stay exempt, mirroring the parent gate. permEntries carry `subagentId`/`subagentType`.
- **Settings plumbing** — new claude-code-only flag (default **on**, zero behavior change until opted out): prefs schema + `AGENT_FLAGS`, agent-gate reader, `_serverCtx` wiring, settings row under the permission toggle (greyed while the parent permission gate is off), i18n en/zh-CN/zh-TW/ko/ja. Flipping it off dismisses only pending subagent bubbles (`dismissPermissionsByAgent(id, { subagentOnly: true })`, same exemptions). The flag is rejected for any other agent at the command layer (Yunbao review: prevents a stray kimi-cli call from tripping Kimi's permission-state disposal).

## Verified on a real machine

Isolated live run (real CC 2.1.169 TUI in a pty, user-level hooks excluded, capture server simulating the gate-off connection drop):
- Subagent `Bash` approval fired the hook **once**, payload as above; capture server destroyed the connection.
- CC parked the approval in the **agents pane** ("← for agents / ↓ to manage") with the full native prompt — `Do you want to proceed? 1. Yes / 2. Yes, and always allow… / 3. No` — answerable, no hang, no auto-approve; the gated command never executed.
- UX note: with the toggle off, subagent prompts live in CC's agents pane rather than interrupting the main conversation flow.

## Tests

- `test/server-agent-id.test.js` (new): registry-allowlist normalization, subagent classification, hook_source precedence.
- `test/server-route-permission.test.js`: CC subagent suite — default bubble with `claude-code` identity, gate-off destroy, exemptions, headless/PASSTHROUGH ordering, CodeBuddy untouched, **verbatim live-capture fixture**, plus regression guards proving the master switch / permission sub-gate now apply to subagent requests.
- `test/permission-dismiss-by-agent.test.js` (new): real `subagentOnly` filter semantics.
- `test/agent-gate.test.js`, `test/prefs.test.js`: flag defaults/normalization scoping, setAgentFlag side effects + claude-code-only guard.
- `test/settings-renderer-browser-env.test.js`: row mounts for claude-code only, greys with the permission gate.

Full suite: 3779 pass; the 8 `install.test.js` failures are pre-existing on `main` (env-dependent version probing), count unchanged.

Reviewed by Yunbao (Codex) — no blocking findings; both "suggested" items (command-layer scoping, coverage gaps) are addressed in this branch.

## Compatibility

- Older CC versions don't send `agent_id` → the toggle has no effect there (requests keep bubbling as today).
- CodeBuddy / other agents: hook scripts stamp registry ids → routed exactly as before; payloads without `agent_id` keep defaulting to claude-code as before.


rullerzhou-afk:
**Live GUI smoke (real desktop, branch build, real Claude Code 2.1.169) — both rounds pass:**

Round 1 — toggle ON (default): real Task-subagent `Bash` approval popped a Clawd bubble; session labeled `agent=claude-code` in session-debug.log (the pre-fix code would have stamped the raw 17-hex instance id); user clicked Deny on the bubble → `IPC permission-decide: behavior=deny matched=true` → proper deny JSON returned to CC; subagent wrapped up cleanly (SubagentStop → Stop); gated command never executed.

Round 2 — toggle OFF (flipped via the new Settings row, persisted as `agents["claude-code"].subagentPermissionsEnabled: false` with `permissionsEnabled` intact): same trigger, **no bubble**; permission-debug.log shows the new path —
```
/permission hit | DND=false pending=0
claude-code subagent bubbles disabled → destroy connection, chat fallback (tool=Bash subagent=general-purpose)
```
CC parked the approval in its agents pane (`◯ general-purpose · Run touch command in tmp`), answerable there; gated command never executed.

PR #462, opened 2026-06-10
fix(focus): resolve terminal PID and switch tmux pane for tmux users

cod3hulk:
## Summary
- When running Claude Code inside tmux, the "Jump to Terminal" / pet click focus feature silently failed because the process tree walk couldn't reach the GUI terminal through the tmux server (parented to launchd, not the terminal app)
- Adds tmux client PID resolution in `hooks/shared-process.js` so `sourcePid` correctly resolves to the actual terminal (e.g. Alacritty, kitty, Terminal.app)
- Adds `scheduleTmuxPaneFocus` in `src/focus.js` to switch to the correct tmux window/pane after focusing the terminal window, using the full binary path since Electron apps don't inherit shell PATH

## Test plan
- [x] Verified `sourcePid` resolves to Alacritty PID instead of tmux server PID
- [x] Verified `focus-debug.log` shows correct PID chain including tmux client and terminal
- [x] Verified tmux pane matching logic correctly identifies target pane from pidChain
- [x] Existing tests pass (`node --test test/shared-process.test.js test/focus-mac-extras.test.js`)
- [ ] End-to-end: click pet/dashboard with multiple Claude sessions in different tmux windows → correct pane activates
- [ ] Non-tmux regression: plain terminal focus still works as before

### DIFF
--- a/hooks/antigravity-hook.js
+++ b/hooks/antigravity-hook.js
@@ -269,6 +269,7 @@ function buildStateBody(hookName, payload, options = {}) {
   if (pidMeta.detectedEditor) body.editor = pidMeta.detectedEditor;
   if (Number.isFinite(pidMeta.agentPid) && pidMeta.agentPid > 0) body.agent_pid = Math.floor(pidMeta.agentPid);
   if (Array.isArray(pidMeta.pidChain) && pidMeta.pidChain.length) body.pid_chain = pidMeta.pidChain;
+  if (pidMeta.tmuxSocket) body.tmux_socket = pidMeta.tmuxSocket;
   return body;
 }
 
@@ -304,6 +305,7 @@ function buildPermissionBody(hookName, payload, options = {}) {
   if (pidMeta.detectedEditor) body.editor = pidMeta.detectedEditor;
   if (Number.isFinite(pidMeta.agentPid) && pidMeta.agentPid > 0) body.agent_pid = Math.floor(pidMeta.agentPid);
   if (Array.isArray(pidMeta.pidChain) && pidMeta.pidChain.length) body.pid_chain = pidMeta.pidChain;
+  if (pidMeta.tmuxSocket) body.tmux_socket = pidMeta.tmuxSocket;
   return body;
 }
 
--- a/hooks/clawd-hook.js
+++ b/hooks/clawd-hook.js
@@ -421,7 +421,7 @@ function buildStateBody(event, payload, resolve) {
   if (process.env.CLAWD_REMOTE) {
     body.host = readHostPrefix();
   } else {
-    const { stablePid, agentPid, agentCommandLine, detectedEditor, pidChain, foregroundWtHwnd } = resolve();
+    const { stablePid, agentPid, agentCommandLine, detectedEditor, pidChain, foregroundWtHwnd, tmuxSocket } = resolve();
     body.source_pid = stablePid;
     if (detectedEditor) body.editor = detectedEditor;
     if (agentPid) {
@@ -432,6 +432,7 @@ function buildStateBody(event, payload, resolve) {
       }
     }
     if (pidChain.length) body.pid_chain = pidChain;
+    if (tmuxSocket) body.tmux_socket = tmuxSocket;
     if (shouldReportForegroundWtHwnd(event) && foregroundWtHwnd) {
       body.wt_hwnd = String(foregroundWtHwnd);
     }
--- a/hooks/codebuddy-hook.js
+++ b/hooks/codebuddy-hook.js
@@ -83,7 +83,7 @@ readStdinJson()
     const sessionId = (payload && payload.session_id) || "default";
     const cwd = (payload && payload.cwd) || "";
 
-    const { stablePid, agentPid, detectedEditor, pidChain } = resolve();
+    const { stablePid, agentPid, detectedEditor, pidChain, tmuxSocket } = resolve();
 
     const body = { state, session_id: sessionId, event };
     body.agent_id = "codebuddy";
@@ -95,6 +95,7 @@ readStdinJson()
       if (detectedEditor) body.editor = detectedEditor;
       if (agentPid) body.agent_pid = agentPid;
       if (pidChain.length) body.pid_chain = pidChain;
+      if (tmuxSocket) body.tmux_socket = tmuxSocket;
     }
 
     // Answer CodeBuddy immediately so it never sees empty stdout, but don't
--- a/hooks/codex-hook.js
+++ b/hooks/codex-hook.js
@@ -196,12 +196,13 @@ function shouldReportForegroundWtHwnd(event) {
 }
 
 function applyLocalProcessFields(body, resolve, options = {}) {
-  const { stablePid, agentPid, detectedEditor, pidChain, foregroundWtHwnd } = resolve();
+  const { stablePid, agentPid, detectedEditor, pidChain, foregroundWtHwnd, tmuxSocket } = resolve();
   const sourcePid = options.preferAgentPid && agentPid ? agentPid : stablePid;
   body.source_pid = sourcePid;
   if (detectedEditor) body.editor = detectedEditor;
   if (agentPid) body.agent_pid = agentPid;
   if (pidChain.length) body.pid_chain = pidChain;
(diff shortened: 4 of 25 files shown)


rullerzhou-afk (changes requested):
Thanks for this PR — the client-PID bridge is a clever way around the daemonized-server dead end, and I can confirm the core design works. I did a deep review with live verification on a real tmux 3.6b server (multi-window/pane and multi-session/multi-client setups). Bottom line: the happy path works end-to-end, but I hit two issues that should be fixed before merge, plus a few optional follow-ups.

**What I verified working (zero mocks):**
- `scheduleTmuxPaneFocus` end-to-end against a live server: with the agent's pane in window 1 and the active window parked on window 2, the chain `[pane_pid, server_pid]` correctly switched active window 2→1 and active pane 1.1→1.0.
- The hook-side bridge from inside a real pane: `pidChain` came back as `[pane-zsh → tmux server → tmux client → …]` with `client_pid` matching `list-clients` exactly, and it degrades gracefully when no known terminal exists in the client's ancestry.

### Must-fix 1: `test/focus-cmux.test.js` fails on any machine that has tmux installed

`should skip cmux detection on non-macOS platforms` asserts the Linux path makes no `ps -o comm=` calls (test/focus-cmux.test.js:322). The new `scheduleTmuxPaneFocus` call in the Linux branch (src/focus.js:1518) violates that — but only when tmux is actually installed, because `resolveTmuxBin()` (src/focus.js:1092) probes with a real `fs.statSync` the test's `execFile` mock can't intercept. I reproduced both states on one machine: suite green before installing tmux, red after. CI never sees it: only the windows job runs `npm test`, and the four Unix probe paths don't exist there.

Suggested direction: make the tmux binary resolution injectable, update the cmux test's expectation deliberately, and add mocked unit tests for `scheduleTmuxPaneFocus` (happy path / non-tmux chain / pane missing from `list-panes` all fit the existing `execFile`-mock pattern — happy to share the four scenarios I used).

### Must-fix 2: wrong client resolved when the agent's session is detached

Good news first: while a client is attached to the calling pane's session, `display-message -p '#{client_pid}'` (hooks/shared-process.js:279) resolves session-scoped — I verified two sessions × two clients returns the correct client regardless of attach order, so the common case is solid.

But when the agent's session is **detached** while another session has a client (detach-and-come-back-later is a core tmux workflow), the server falls back to the *other* session's client. Verified live: client A on session `aaa` (the agent's), client B (pid 72756) on `bbb`; after detaching A, `display-message` from inside `aaa` returned 72756. `stablePid` then points at an unrelated terminal, and clicking the pet raises the wrong window — arguably worse than the silent no-op before this PR.

Heads-up: scoping with `display-message -p -t "$TMUX_PANE" …` does **not** fix this — verified it falls back identically. What does work (verified in both states):

```sh
tmux list-clients -t "$TMUX_PANE" -F '#{client_pid}'
```

- attached → returns only the pane's session's client(s)
- detached → returns empty → skip the bridge → existing fallback behavior

(`$TMUX_PANE` is its own env var; the third field of `$TMUX` is the session id, not a pane id.)

### Optional follow-ups (fine as separate issues)

- **switch-client hardening:** `select-window`/`select-pane` are invisible to a client attached to another session (verified). With must-fix 2 the resolved client is always on the agent's session, so this mostly evaporates; a `switch-client` would only guard the narrow race where the client hops sessions between hook resolution and the click.
- **Custom sockets:** the hook's tmux CLI inherits `$TMUX` so it reaches the right server, but the Electron side always talks to the default socket — `tmux -L work` users get window focus with a silently no-op pane switch.
- **Gated tests portability:** the new `skip: !process.env.TMUX` tests (test/shared-process.test.js:130) never run in CI, and fail for tmux users without a GUI terminal in the client's ancestry (ssh, IDE terminals, headless — reproduced). Mocked tests would cover this deterministically.
- `resolveTmuxBin` misses MacPorts (`/opt/local/bin/tmux`) and nix paths.
- Long-term, `#{window_id}`/`#{pane_id}` targets are sturdier than `session:index.index` strings (tmux ≥3.x normalizes `.`/`:` in session names to `_`, so today's parsing is safe; ids would future-proof it).
- Style nit: `if (walkReachedTmux) try {` without braces (hooks/shared-process.js:278).

The mechanism itself is solid and this will be a great quality-of-life win for tmux users — with the two must-fixes in, I'd be glad to merge. Happy to share my verification scripts (mocked scenarios + live-server repro steps) if useful.

cod3hulk:
Thanks for the deep review — that detached-session repro especially is a saver. Pushed [a7be911](https://github.com/cod3hulk/clawd-on-desk/commit/a7be911) addressing both must-fixes and all the follow-ups you flagged.

### Must-fix 2 — detached-session client resolution
Swapped `display-message -p '#{client_pid}'` for `tmux list-clients -t "$TMUX_PANE" -F '#{client_pid}'`. Attached → returns the session-scoped client pid(s); detached → returns empty, the bridge is skipped, existing fallback runs. Also iterates multi-client results and only appends a winning walk to `pidChain` (intermediates buffered in `localAdds`) so a client that didn't reach a known terminal doesn't pollute the chain.

### Must-fix 1 — `focus-cmux` test + new mocked unit tests
- `resolveTmuxBin` now accepts `__setTmuxBin(path)` (exposed via `__test`) so tests can disable the bridge without touching `fs.statSync`.
- `should skip cmux detection on non-macOS platforms` now disables the tmux bridge and asserts on the *actual* cmux probes (`mdfind` and cmux AppleScript) instead of the brittle "no `ps -o comm=`" invariant — which was reasonable before this PR but is wrong now that the tmux bridge legitimately runs that probe on Linux.
- New `test/focus-tmux.test.js` with five mocked scenarios for `scheduleTmuxPaneFocus`: happy path, non-tmux chain, missing pane, missing binary, and custom socket. Used your suggested four scenarios + one for the new `-L` plumbing.

### Follow-ups (all of them, since they were cheap once we were in there)
- **`switch-client` hardening** — `switch-client -t <session>` runs before `select-window`/`select-pane` to cover the narrow race where the client hops sessions between hook resolution and the click. Went with your option (b) — after must-fix 2 the race shrinks enough that plumbing the resolved client pid through for `-c <pid>` didn't seem worth the extra surface area. Happy to revisit if you'd rather have option (a).
- **Custom sockets (`tmux -L name`)** — parse the socket name from the first comma-field of `$TMUX` in `hooks/shared-process.js`, plumb `tmux_socket` through hooks → `server-route-state.js` → `state.js` → `main.js` → `focus.js`, and prepend `-L <name>` to every tmux invocation in `scheduleTmuxPaneFocus`. Wired up across all hooks that emit `pid_chain` (clawd, cursor, codex, qwen-code, kimi, codebuddy, antigravity, qoder, kiro, copilot, gemini, pi-extension, opencode-plugin, hermes-plugin).
- **Sturdier targets** — `list-panes` now formats `#{pane_pid} #{window_id} #{pane_id} #{session_name}` and `select-window`/`select-pane` use the `@N`/`%N` ids directly. `session_name` is parsed via `parts.slice(3).join(" ")` to survive trailing whitespace, though `pane_id`/`window_id` being stable upfront means the leading fields are unambiguous.
- **`resolveTmuxBin` paths** — added `/opt/local/bin/tmux` (MacPorts), `/run/current-system/sw/bin/tmux` (NixOS), and `~/.nix-profile/bin/tmux` (nix home-manager).
- **Gated tests portability** — replaced `{ skip: !process.env.TMUX }` block in `test/shared-process.test.js` with five deterministic mocked cases via a new `test/helpers/load-shared-process-with-mock.js` (mirrors `load-focus-with-mock.js`). Cases: walk reaches tmux server happy path, walk doesn't reach tmux server, `list-clients` empty (detached) → fallback to lastGoodPid, `tmuxSocket` parsed from `$TMUX` exposed on the resolver result, `default` socket → `tmuxSocket` is null.
- **Style nit** — added braces to the `if (walkReachedTmux)` block.

### Test results
3814 pass / 8 fail. All 8 failures are in `test/install.test.js` (Claude version detection helpers) — pre-existing on macOS (Windows-path expectations), nothing my diff touched. Confirmed via `git diff --stat main..HEAD -- src/install.js src/version.js test/install.test.js` returning empty.

Would still appreciate eyes on the new mocked tests if you have your verification scripts handy — happy to fold in scenarios I missed.

PR #467, opened 2026-06-11
feat: add free roam mode — pet wanders around desk when idle

zhaoxv210:
## Summary

Adds a **Free roam** toggle to Settings → Appearance that lets Clawd wander freely around the desk when idle. Movement is smooth (~60fps eased animation), stops on mouse move, and respects screen work-area boundaries.

## Changes

- **src/roam.js** (new): Free roam module
- **src/main.js**: Wired roam context + settings mirror + tick integration
- **src/tick.js**: Calls roam.tick() during idle, cancels on mouse move
- **src/prefs.js**: Added freeRoam boolean (default: false)
- **src/settings-actions.js**: Added freeRoam validator
- **src/settings-tab-general.js**: Toggle in Appearance section
- **src/settings-i18n.js**: i18n for en/zh/zh-TW/ko/ja

## Tests

All 3883 existing tests pass (0 failures).


rullerzhou-afk:
Thanks for the PR. The settings/i18n plumbing is useful and looks reusable, but I don't think we can merge this implementation as-is.

The main blocker is that the new roam path moves the normal idle pet window directly instead of going through the existing movement/mini animation state machine. In the PR diff, `src/tick.js` calls `ctx.roam.tick()` from the normal `idle` branch, while `src/roam.js` only allows roaming when `currentState === "idle"` and mini mode is false. The actual animation then drives `win.setBounds(...)` directly in `src/roam.js:73-84`, without applying a walk/crabwalk state. By contrast, the existing menu mini entry path switches to `mini-crabwalk` first (`src/mini.js:479`) and then moves the window (`src/mini.js:492`). The result is that the idle SVG is visibly dragged around the desktop, which also matches a Windows regression we hit while trying this.

There are also two correctness issues around cancellation and window sync:

- Once `animateTo()` starts, each `step()` only checks `roamActive` and whether the window is destroyed. It does not re-check `isRoamAllowed()` / `getCurrentState()`, so an async state change to working, notification, or permission can leave the pet drifting for the rest of the 2.5s animation.
- During the animation, roam only moves the render window. `syncHitWin()` and anchored-surface repositioning only run at the end (`src/roam.js:86-88`). Existing movement code like `animateWindowX()` keeps the hit window/HUD/bubbles synchronized per frame. This means the visible pet and its input area can diverge during the roam animation, which is especially risky on Windows.

A smaller but still real issue: `ROAM_IDLE_DELAY_MS = 8000` is defined but not used; the first roam is scheduled with `ROAM_BETWEEN_DELAY_MS` instead, so it can start after about 4s rather than the intended idle delay.

I ran the related existing tests (`node --test test/tick.test.js test/mini.test.js test/prefs.test.js test/settings-actions.test.js`) and they pass, but there are no tests covering `src/roam.js`, so the green test run does not exercise the new behavior.

Recommended direction: keep the settings/i18n pieces if we still want the feature, but rework the movement path so it uses an explicit movement visual state, synchronizes hit/floating windows throughout the animation, cancels immediately on non-idle state changes, and has dedicated unit coverage for the roam timing/cancellation rules. This is a bigger behavioral rework, so please don't treat this as an obligation to keep iterating on the PR if you don't want to; we can take it from here if needed.

rullerzhou-afk:
Thanks again for the Free Roam work and for iterating on the earlier review. I pushed one small maintainer follow-up before merging: it restores idle when Free Roam is disabled mid-roam and removes a redundant per-frame window update/HUD reposition.

The PR is merged now. Appreciate the thoughtful implementation and the test coverage. We'll still do a quick Windows smoke before the next release, but the code side is in.

zhaoxv210:
Thanks for merging and the cleanup follow-up! If the Windows smoke test turns up anything, feel free to tag me and I'll help investigate.

PR #477, opened 2026-06-12
fix: validate cached Windows focus targets

lxgxhsy:
## Summary
  - Validate cached Windows focus HWNDs against current cwd title candidates before reuse.
  - Avoid caching ambiguous Windows Terminal fallback targets.
  - Match VS Code/Cursor parent windows by cwd title before focusing, preventing multi-window session jumps.

  ## Context
  When clicking a Session HUD row for one project, the HUD sent the correct session id/cwd, but Windows focus reused a cached HWND that belonged  
  to another session window. This could make clicking a later session jump to a different project/session.

  ## Test plan
  - [x] node --check src\focus.js
  - [x] node --check test\focus-windows.test.js
  - [x] PowerShell parser validates generated focus helper
  - [x] node --test test\focus-windows.test.js

### DIFF
--- a/src/focus.js
+++ b/src/focus.js
@@ -242,7 +242,7 @@ function psSingleQuotedString(value) {
   return `'${String(value || "").replace(/'/g, "''")}'`;
 }
 
-function makeFocusCmd(sourcePid, cwdCandidates, focusCacheKey = null, wtHwnd = null, focusToken = "") {
+function makeFocusCmd(sourcePid, cwdCandidates, focusCacheKey = null, wtHwnd = null, focusToken = "", cacheCwdCandidates = cwdCandidates) {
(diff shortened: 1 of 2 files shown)


rullerzhou-afk:
Thanks for the careful fix, @lxgxhsy! 🙏 Re-validating the cached HWND against the session's sourcePid + window title before reuse — and matching VS Code/Cursor parent windows by cwd title — is exactly the right call for the wrong-window jumps, and dropping the weak-fallback caches keeps the cache honest. Verified on a real Windows box across all three cases (stale HWND, multi-window VS Code/Cursor, and title-mismatch) and everything behaved correctly. Merged 🎉 Appreciate the thorough write-up and tests!

PR #505, opened 2026-06-16
feat(focus): capture and restore focus to exact window on macOS

BucleLiu:
When a Claude Code / Codex session starts or the user submits a prompt, capture the frontmost window's metadata (app name, title, bounds) via System Events AppleScript. On focus-back, use a 5-level cascade to restore focus to the exact window instead of just activating the app:
	L1 – title + bounds match (exact window)
	L2 – project-name unique match (CWD folder in window title)
	L3 – title substring match (stable prefix from editor title)
	L4 – bounds unique match (only window at captured position)
	L5 – generic app activation (existing fallback)

Skips captured-window focus for apps with precise handlers (iTerm2, Ghostty, cmux, Superset) to avoid conflicts.

### DIFF
--- a/src/agent-runtime-main.js
+++ b/src/agent-runtime-main.js
@@ -37,6 +37,7 @@ function createAgentRuntimeMain(options = {}) {
   const isAgentEnabled = options.isAgentEnabled || (() => true);
   const updateSession = options.updateSession || (() => {});
   const captureGhosttyTerminalId = options.captureGhosttyTerminalId || null;
+  const captureFrontmostWindow = options.captureFrontmostWindow || null;
   const showCodexNotifyBubble = options.showCodexNotifyBubble || (() => {});
   const clearCodexNotifyBubbles = options.clearCodexNotifyBubbles || (() => {});
 
@@ -93,6 +94,7 @@ function createAgentRuntimeMain(options = {}) {
     }
     const result = updateSession(sessionId, state, event, opts);
     maybeCaptureGhosttyTerminalId(sessionId, event, opts);
+    maybeCaptureWindowMetadata(sessionId, event, opts);
     return result;
   }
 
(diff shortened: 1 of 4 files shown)


rullerzhou-afk:
Thanks for working on this. The overall direction makes sense to me: restoring focus to the exact macOS window would be a real improvement over app-level activation.

I found a couple of issues that I think should be addressed before merging:

1. `captureFrontmostWindow` currently runs for every local `SessionStart` / `UserPromptSubmit`, before we know whether the captured metadata will ever be used.

`maybeCaptureWindowMetadata` in `src/agent-runtime-main.js` has no app gate or dedupe, and `captureFrontmostWindow` then runs a `System Events` AppleScript. However, `scheduleCapturedWindowFocus` later skips iTerm2 / Ghostty / cmux / Superset because those already have more precise handlers.

That means users of those apps still pay the capture cost, and may hit macOS Automation / Accessibility prompts, even though the captured window is discarded on restore. This also cuts against the existing macOS focus path, which deliberately prefers `open <bundle>` because it does not need Automation consent.

Could we move the skip logic to the capture side too, and preferably avoid recapturing on every prompt when the session already has valid captured-window metadata?

2. The "bounds" matching currently only compares window position, not full bounds.

In `buildAxTitleBoundsFocusScript` and `buildAxBoundsUniqueFocusScript`, the scripts compare only `wx` and `wy`. They do not compare width or height, even though capture stores `w` and `h`, and the PR description says "title + bounds" / "bounds unique".

This can produce false matches when another same-app window occupies the same top-left position, and it makes the implementation weaker than the PR description suggests. I think L1/L4 should compare `x/y/w/h`, or the PR should explicitly describe this as position-only matching.

One more risk worth validating on a real macOS machine: `capturedWindow.appName` is derived from `ps -o comm` basename and later used as `System Events process "<appName>"`. For Electron-based apps like VS Code / Cursor, we should verify that this name matches the System Events process name and that L1-L4 do not silently miss.

I also could not find tests covering the new captured-window path. It would be great to add focused tests for:
- capture-side gating / dedupe
- `capturedWindow` storage in session metadata
- L1/L4 full-bounds matching
- `extractStableTitlePrefix` behavior for common editor titles

So my vote is: good feature direction, but please tighten the capture gate and the bounds matching before merge.

PR #507, opened 2026-06-16
feat: add dizzy reaction when cursor rapidly circles the crab

IatomicreactorI:
## What

When the user rapidly circles the cursor around Clawd during idle eye-follow state, accumulated angular displacement triggers a "dizzy" animation. The existing `clawd-dizzy.svg` (spiral eyes, body sway, spinning stars) plays for 6 seconds before auto-returning to idle.

## How It Works

**Spin detection in `src/tick.js`:**
- Tracks cursor angle around the pet center in idle eye-follow state
- Accumulates absolute angular displacement with exponential decay (half-life 800ms) — slow movement decays away, only rapid circling builds up
- Threshold: 2 full circles (4 PI radians) triggers dizzy
- 12-second cooldown prevents repeated triggering

**Theme config in `themes/clawd/theme.json`:**
- Wires the `"dizzy"` state to the already-existing `clawd-dizzy.svg` asset
- 6-second display + auto-return timing

## Demo

In idle state, rapidly circle the cursor around the crab approximately 2 times → spiral eyes, body wobble, spinning stars!

## Files Changed

- `src/tick.js` — spin detection algorithm + dizzy trigger
- `themes/clawd/theme.json` — dizzy state registration

### DIFF
--- a/src/tick.js
+++ b/src/tick.js
@@ -22,13 +22,14 @@ let mainTickActive = false;
 let nextMainTickAt = 0;
 
 // ── Spin detection: tracks cursor circling to trigger dizzy animation ──
-let lastCursorAngle = null;             // last cursor angle (radians) relative to pet center
-let accumulatedSpin = 0;                // accumulated absolute angular displacement
+let lastCursorAngle = null;             // last cursor angle (radians) relative to eye-tracking origin
+let accumulatedSpin = 0;                // signed accumulated angular displacement
 let dizzyCooldownUntil = 0;             // timestamp until which dizzy cannot re-trigger
 let lastSpinTickAt = 0;                 // timestamp of last spin tick
 
-const SPIN_THRESHOLD = Math.PI * 4;     // 2 full circles to trigger dizzy
-const SPIN_DECAY_HALF_LIFE_MS = 800;    // spin memory decays with this half-life
+const SPIN_THRESHOLD = Math.PI * 4;     // 2 full circles (signed) to trigger dizzy
+const SPIN_IDLE_RESET_MS = 500;         // a pause longer than this resets the spin meter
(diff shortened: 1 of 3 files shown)


rullerzhou-afk:
Thanks for the feature, this is a fun interaction. I think there is one blocker before merge:

The spin detector lives in `src/tick.js`, so it can run for every theme, but this PR only wires the `dizzy` state and `autoReturn.dizzy` for the Clawd theme. Calico and Cloudling do not define `states.dizzy` or `timings.autoReturn.dizzy`.

That means if the user triggers the gesture while using Calico/Cloudling, `ctx.setState("dizzy")` can put the state machine into logical `dizzy`; the visual falls back to the idle SVG, but no auto-return timer is scheduled. From there, normal idle eye-follow / idle sleep logic no longer runs until some external event changes state again.

Suggested minimal fix: gate the trigger in `tick.js` on theme support, e.g. only call `ctx.setState("dizzy")` when the active theme has a real `states.dizzy` binding and a positive `timings.autoReturn.dizzy`. Unsupported themes should just keep their normal idle behavior.

A focused test would be useful too: Clawd theme fast spin triggers `dizzy`; Calico or Cloudling fast spin does not.

IatomicreactorI:
Good catch, that makes total sense — I'd missed that `tick.js` runs for every theme.

Added a `themeDizzyReady` guard in the trigger path: it only calls `ctx.setState("dizzy")` when the active theme truly has `states.dizzy` (non-empty array) **and** `timings.autoReturn.dizzy` (positive number). Unsupported themes — Calico, Cloudling, etc. — have this evaluate to `false`, so the spin math still runs internally but the threshold is never reached and idle behaviour is unchanged.

Tested: Clawd fast-spin → dizzy ✅ / Calico fast-spin → stays idle ✅ / Cloudling fast-spin → stays idle ✅

Issue #215, opened 2026-04-30
Feature request: Optional Discord Rich Presence integration


boomho:
### 背景

我想建议 Clawd on Desk 增加一个可选的 Discord Rich Presence 集成功能。这个功能不需要很严肃，更多是一个有趣的 ambient status：让 Discord 显示当前 AI coding agent 的大致状态，例如 thinking / working / waiting / idle。

我自己刚做了一个 Codex Discord Rich Presence 的本地小工具，实践下来感觉这个功能和 Clawd 的定位很适合：Clawd 已经在本地判断 agent 状态，只需要把其中一部分低敏状态同步到 Discord Rich Presence。

### 建议功能

- 在设置里提供 Discord Rich Presence 开关。
- 显示粗粒度状态即可，例如：
  - `Codex is thinking`
  - `Codex is working`
  - `Codex is waiting`
  - `Codex is idle`
- 可以复用 Clawd 已有的状态机，把 thinking / working / notification / attention / sleeping 等状态映射到 Discord 文案。
- 不建议默认显示具体项目名、文件名、路径、分支名、prompt、命令内容等敏感信息。
- 如果要显示项目名或 session 名，建议做成用户明确 opt-in。
- 更简单安全的默认方案：只显示 agent 类型和状态，不显示正在做哪个项目。

### 技术实现建议

我本地实现时采用的是这个结构：

1. Codex hook 只负责接收生命周期事件，并把最小状态写到本地状态文件，例如：
   - `SessionStart` -> idle
   - `UserPromptSubmit` / `PostToolUse` -> thinking
   - `PreToolUse` -> working
   - `PermissionRequest` -> waiting
   - `Stop` -> idle
2. 一个长期运行的本地 daemon 读取 hook state 和 session JSONL，推导当前状态。
3. daemon 通过本机 Discord Desktop 的 RPC/IPC 设置 Rich Presence activity。
4. Discord 侧只需要用户创建一个 Discord Developer Application，并填写 Application ID / Client ID；不需要 bot token，也不需要把 bot 加进服务器。
5. Presence 的 start timestamp 最好保持稳定，不要每次刷新都重置；只有文案变化或定时 heartbeat 时更新。
6. 如果同时支持 Codex subagent，建议复用“main/root agent 和 subagent 分离”的逻辑，避免 subagent 完成导致 Discord 状态误显示为整个任务完成。

### 隐私建议

[...]

PR #508, opened 2026-06-16
Discord Rich Presence (#215)

KaiC5504:
Implements opt-in Discord Rich Presence: shows the active agent and a coarse
session state (Thinking / Working / Waiting for input / Idle) on the user's
Discord profile via local IPC. Addresses #215.

**What changed**
- Privacy-locked by default — only the agent name + coarse state are published.
  Project folder name is an explicit opt-in; file/branch/command/prompt are never
  sent (covered by a unit test).
- Zero new dependencies: raw Node `net` over the Discord IPC pipe.
- Desktop-only; non-fatal when Discord isn't running (backoff reconnect).
- Settings tab with a bring-your-own Application ID. `DEFAULT_CLAWD_DISCORD_APP_ID`
  is left empty as a handoff seam — drop the official published App ID into that
  constant for the zero-friction default from #215.

---

实现可选的 Discord Rich Presence：通过本地 IPC 在用户的 Discord 个人资料上显示当前
agent 和粗粒度的会话状态（Thinking / Working / Waiting for input / Idle）。关联 #215。

**改动内容**
- 默认隐私锁定 —— 仅发布 agent 名称和粗粒度状态。项目文件夹名称为显式可选项；
  文件/分支/命令/prompt 内容从不发送（有单元测试覆盖）。
- 零新增依赖：使用原生 Node `net` 直连 Discord IPC 管道。
- 仅限桌面端；Discord 未运行时不影响程序（指数退避重连）。
- 设置页支持自带（BYO）Application ID。`DEFAULT_CLAWD_DISCORD_APP_ID` 留空作为交接位
  —— 把官方发布的 App ID 填入该常量即可获得 #215 中的零摩擦默认体验。


rullerzhou-afk:
Thanks, this is close, but I am going to request changes before merging.

The main blockers are:

1. `privacyShowProject` can leak a full cross-platform path via `path.basename(session.cwd)`. On macOS/Linux, a Windows cwd like `C:\Users\alice\Projects\secret` is published as the full path, despite the UI promising only the folder name.
2. `pickDominantSession()` should skip `hiddenFromHud` / sleeping sessions to stay aligned with HUD/session visibility; otherwise Discord can show superseded Codex sessions.
3. The default Application ID handoff is incomplete: `readiness()` supports `DEFAULT_CLAWD_DISCORD_APP_ID`, but the Settings enable switch still requires a user-saved `applicationId`.
4. Changing Application ID while connected does not reconnect, so the old `client_id` stays active until restart/disconnect.

The privacy allowlist and core IPC direction look reasonable, and the targeted tests pass, but these should be fixed before merge.

KaiC5504:
All four addressed in 1d57f08:

1. **Path leak** — the folder name now goes through `path.win32.basename`, so a Windows cwd observed on a POSIX host publishes only the folder name, never the full path.
2. **Session visibility** — `pickDominantSession` now skips `headless` / `sleeping` / `hiddenFromHud`, the exact complement of `isHudSession()`; superseded Codex sessions drop out with it.
3. **Default App ID** — the enable switch now honors `DEFAULT_CLAWD_DISCORD_APP_ID` via a preload-exposed flag, matching `readiness()`, so it's ready without a saved ID once the constant is filled.
4. **Reconnect** — changing the Application ID while connected now re-dials with the new `client_id` (race-safe, and recovers from a synchronous connect throw).

New unit tests cover all four: the path-leak assertion, HUD-aligned picking, the App-ID handoff, and the reconnect lifecycle via a fake IPC socket.

KaiC5504:
Follow-up pushed in `232df6d`.

Heads-up: while wiring point 3 (enable switch honoring the default App ID), `1d57f08`'s new `require("./discord-presence-settings")` in `preload-settings.js` broke the Settings window — the settings preload is sandboxed on Electron 41, where a relative `require` throws, so `settingsAPI` never loaded and the window opened blank. Caught in live testing.

`232df6d` fixes it without weakening the sandbox: the "default App ID present" flag is now injected via `webPreferences.additionalArguments` and read from `process.argv` (same pattern as `preload-hit.js`). Point 3's behavior is unchanged. Also added: `before-quit` now stops the presence bridge, `stop()` resets the reconnect backoff, and regression tests for all three. Suite green.

rullerzhou-afk:
Merged — thank you for the careful work across all three rounds. The privacy allowlist, the raw-`net` IPC bridge, and the reconnect/race handling all held up well under review and on a real machine.

Notes on the merge:

- I resolved the conflicts with latest `main` while merging (the Feishu channel from #511 landed in the meantime, plus a sidebar-icon refactor — tab icons are now SVGs resolved by id in `settings-icons.js`, so I added a gamepad icon for `discord-presence` there).
- Live-verified today on Windows 11 + Discord stable, with a real Claude Code session: handshake/READY, all four coarse states, the project-name opt-in (folder name only), App-ID readiness gating in the Settings tab, and the external icon URL — the `assets/icon.png` fallback renders fine on the profile card, so BYO users indeed need no uploaded asset.

A few small polish items I'll handle myself in a follow-up (no action needed on your side): the Save button wraps its label in CJK locales, the developer-portal link should route through the existing `openExternalSafe` helper instead of `target="_blank"`, zh terminology alignment, greying the project-name toggle while presence is disabled, and truncating `state` to Discord's 128-char limit for very long folder names.

The `DEFAULT_CLAWD_DISCORD_APP_ID` handoff seam stays as designed; I'll fill in the official Application ID separately.

PR #510, opened 2026-06-16
feat(qoderwork): support QoderWork IDE agent in clawd

xiaoshidefeng:
## Feature

  The desktop pet now reacts to QoderWork IDE in real time. When running tasks in QoderWork, the pet
  plays corresponding animations (thinking, working, done, error, etc.), and the Session HUD panel shows
  the user's query and execution status.

  Consistent with Qoder and Reasonix, Phase 1 is state-only — the pet observes QoderWork's permission
  events but never answers permission decisions. QoderWork's native permission flow stays in control.

  ## Technical

  - **Agent registration**: Wire QoderWork into registry, prefs, settings, cleanup, integration sync,
  doctor descriptors, and installer (`~/.qoderwork/settings.json`, hook-based)
  - **Hook script** (`hooks/qoderwork-hook.js`): Read QoderWork's hook stdin payload, map to Clawd
  states, POST to the local server. Stdout always writes `{}` (no gating)
  - **Install script** (`hooks/qoderwork-install.js`): Register 10 hook events in
  `~/.qoderwork/settings.json`

  ### Hook design decisions (verified against real payloads)

  - **PermissionRequest / PermissionDenied → `working`**: QoderWork fires 40+ permission events per task.
  Mapping to `notification` caused constant pet flashing. Phase 1 never answers decisions — silent
  handling is correct.
  - **Session title from high-quality sources only**: Priority is `payload.session_title` → `prompt`
  first line → `parent_business_info.name`. No `cwd` fallback sent — avoids workspace IDs (e.g.
  `mqgw60jiigjsjcid`) overwriting good titles via the server's sticky `||` chain.
  - **No token/usage data**: QoderWork hook payloads don't include token counts or model info — HUD
  doesn't display these.

  ## Tests

  - Added QoderWork cases in `registry.test.js` and `doctor-agent-descriptors.test.js`
  - Full suite passes (4077/4078, 1 failure is a pre-existing flaky proxy timeout test, unrelated)


rullerzhou-afk:
Thanks for adding QoderWork support. I did a read-only pass over the PR and the state-only contract looks good: the hook always returns `{}`, and I did not find a path where Clawd answers QoderWork's native permission flow.

One issue looks worth fixing before merge:

- `hooks/qoderwork-hook.js` passes `new Set(["QoderWork.exe"])` / `new Set(["QoderWork"])` into `createPidResolver`, but `hooks/shared-process.js` lowercases process names before matching (`proc.Name.toLowerCase()` on Windows and `basename(commOut).toLowerCase()` on macOS/Linux). That means `agent_pid` will be missed on all three platforms. `source_pid` can still be present, so the main state/HUD path is not broken, but agent-level reachability/cleanup that depends on `agentPid` will not work as intended. The small fix is to use lowercase names in the resolver set, e.g. `qoderwork.exe` / `qoderwork`.

Non-blocking follow-ups I noticed:

- `agents/qoderwork.js` says startup recovery watches the IDE process name, but `src/state.js` still uses a hard-coded recovery process query and currently only includes `qodercli` / `qoder-cli`, not `QoderWork`.
- The registry `eventMap` says `PermissionRequest` / `PermissionDenied` map to `notification`, while the live hook maps them to `working` / `PreToolUse`. The runtime behavior seems intentional, but the metadata/test now gives a different impression.
- If QoderWork sends an internal cwd like `~/.qoderwork/workspace/<id>` and no title source is available, the HUD can still fall back to displaying that workspace id via `path.basename(cwd)`. The helper functions to detect/resolve that cwd appear to be present but not wired into `buildStateBody`.
- It would be useful to add `qoderwork-hook` / installer tests similar to the existing Qoder coverage; the current tests mostly validate registry/doctor metadata.

After the `agent_pid` casing fix, I would be comfortable treating the rest as follow-up cleanup.

xiaoshidefeng:
Thanks for the thorough review! All points addressed in 72b3f1d:

**Blocking fix:**
- ✅ `agent_pid` casing — lowered to `qoderwork.exe` / `qoderwork` in `createPidResolver` to match `shared-process.js` normalization.

**Non-blocking follow-ups (also fixed in this commit):**
- ✅ Startup recovery — added `qoderwork.exe` (win32) and `pgrep -x 'qoderwork'` (darwin/linux) to `detectRunningAgentProcesses` in `state.js`.
- ✅ `eventMap` metadata — updated `PermissionRequest`/`PermissionDenied` from `"notification"` to `"working"` in `agents/qoderwork.js` to match the runtime hook behavior.
- ✅ Workspace ID leak — added `.qoderwork/workspace/<id>` path detection in `sessionDisplayTitle` (`state-session-snapshot.js`) so the HUD falls through to the truncated session ID instead of displaying the raw workspace ID.
- ✅ Tests — added `test/qoderwork-hook.test.js` (22 tests) and `test/qoderwork-install.test.js` (9 tests) covering hook mappings, `buildStateBody`, session title resolution, command-line detection, workspace cwd detection, installer idempotency, third-party hook preservation, uninstall, and more. Also updated the existing `registry.test.js` assertions to match the new eventMap values.

All 75 related tests pass with zero regressions.

rullerzhou-afk:
Merged! 🎉 Your QoderWork integration is now in main via #598 (f42f970), with your authorship preserved in the squash commit (`Co-Authored-By: xiaoshidefeng`).

Since this PR was opened from your fork's `main` branch, I couldn't push the merge-conflict resolution and follow-ups here, so I carried your commits into a maintainer branch instead. On top of your two rounds:

- Resolved conflicts with the Kimi Code changes that landed on main (#593)
- Skipped pid resolution for `PermissionRequest` / `PermissionDenied` — since QoderWork waits on the hook's stdout and these fire 40+ times per task, this cuts each permission event from ~650ms to ~65ms on Windows
- Scoped the workspace-id HUD skip to QoderWork sessions so other agents running inside `~/.qoderwork/workspace/<id>` keep their normal title fallback
- Fixed a case-sensitivity miss in startup recovery (`pgrep` for the `QoderWork` process name)
- Small cleanups: removed the unused cwd-resolution helpers, synced stale comments, added QoderWork to the READMEs

Everything passed the full suite plus a live smoke of the whole state machine, but I don't have a QoderWork install here — once the next release ships, it would be great if you could give it a quick regression pass on your real setup (especially the permission-event latency and HUD titles). Thanks again for the thorough work and the fast review turnaround! Closing this in favor of #598.

rullerzhou-afk:
Update: QoderWork support shipped in v0.12.0.

@xiaoshidefeng one heads-up — the Windows hook command was switched to the bash-portable form in this release (same fix as Qoder #615/#597), so it parses correctly under both bash and cmd. Could you update and run a quick regression on your QoderWork setup to confirm the approval prompt fires? Thanks!

PR #528, opened 2026-06-20
Add Spanish (es) locale + mobile PWA i18n and pairing QR

Zamaniego:
## Add Spanish (`es`) locale

Complete **Spanish (Mexican, neutral)** translation, registered as a selectable language alongside `en, zh, zh-TW, ko, ja` — covering the desktop UI, the main-process dialogs, the language-picker plumbing, first-run auto-detection, **and the mobile PWA**.

### Desktop UI
- **`src/i18n.js`** — full `es` block (222 keys) + `es` added to `SUPPORTED_LANGS`.
- **`src/settings-i18n.js`** — full `es` block (644 keys, incl. the 19 function-valued strings) + a `langSpanish: "Español"` label in every locale.
- **`src/bubble-renderer.js`**, **`src/settings-ipc.js`**, **`src/settings-animation-overrides-main.js`**, **`src/codex-pet-main.js`** — `es` for permission bubbles and native dialogs.
- **`src/settings-tab-general.js`** — `es` in `LANGUAGE_OPTIONS` + the label-key map.
- **`src/tutorial-renderer.js`** — `Español` in the onboarding `LANG_LABELS`.

### Make `es` actually selectable & auto-detected
- **`src/prefs.js`** — add `es` to the `lang` schema enum **and** map `es-*` device locales to `es` in `mapLocaleToLang` (fresh-install auto-seed).
- **`src/settings-actions.js`** — add `es` to the `lang` `requireEnum` (without this the picker silently reverts any non-built-in language).

### Mobile PWA (`pwa/`) — full i18n that follows the desktop language
The mobile companion shipped with every string **hard-coded in Chinese** (`<html lang="zh-CN">`, inline text). Extracted them into a proper string catalog (`pwa/i18n.js`) covering the **full desktop locale set — en, zh, zh-TW, ko, ja, es** — and routed all UI text (session states, hook-event labels, relative time, toasts, empty-state, settings, nav) through a `t(key)` helper in `app.js` / `data-i18n` in `index.html`. The LAN server injects the desktop app's selected language as `window.__CLAWD_LANG__` (and sets `<html lang>`), so **the PWA shows whatever language you picked on the desktop**, with document/browser fallback. Placeholders (`{n}`, `{label}`) preserved; keysets verified at parity across all six locales.

Mobile UX improvements alongside: the whole session card is now tap-to-expand/collapse (touch-friendly, not just the chevron); the expanded view shows the **assistant's latest output** — the text it just wrote (`assistantLastOutput`, now included in the LAN payload along with `contextUsage`) — plus the full recent-event list (server now sends up to 25); and the service worker is now network-first for the HTML (and bumped to a versioned cache) so the PWA always reflects the current desktop language instead of a stale cached copy.

Further mobile polish: per-session **context-usage %** shown on each card (highlighted when high), richer **push notifications** (fires on task completion, needs-input, attention and error) with haptic feedback and an in-app "enable notifications" control in Settings.

The phone also shows the **Clawd mascot**: the LAN server serves the active theme's animation SVGs (`/mobile/pet/<file>.svg`, strictly path-sanitized) and the PWA displays the pet reacting to the most active session's state.

Also fixed iOS **Add to Home Screen**: the home-screen icon launched the manifest's `start_url` (`/mobile/`) with no token, and iOS isolates that storage from Safari, so the QR-paired session never reached it. The PWA now appends the pairing params to the manifest link, and the LAN server reflects them into `start_url` — **only echoing the `host/port/token` the requester already presented in the query, never the server's own token**, so an unauthenticated `GET /mobile/manifest.json` still gets a token-less manifest. The PWA also prefers the freshest saved token for that host:port (surviving token rotation), so the icon reconnects without re-pairing. (`src/network/mobile-preview-server.js`, `pwa/app.js`.)

### Desktop: QR code for mobile pairing
The Mobile settings panel previously showed only the pairing URL / IP / port / token as text. Added a scannable **QR code** of the pair URL so a phone can pair straight from the camera. It's rendered in the trusted main process via the dependency-free `qrcode-generator` (added to `dependencies`) and inserted as a generated SVG (no user HTML). (`src/settings-ipc.js`, `src/settings-tab-mobile.js`.)

### Tests
Extended the `SUPPORTED_LANGS`, bubble locale-count, `mapLocaleToLang`, and `lang`-enum assertions. **Full suite (4186 tests) passes.** `test/i18n.test.js` confirms the `es` keyset and placeholder parity across runtime, Settings, bubbles, and main-process dialogs.

Placeholders, markdown links, and proper nouns (Clawd, Codex, Claude Code, Telegram, SSH, HUD…) are preserved verbatim. Happy to adjust any wording to your preference.


rullerzhou-afk:
@Zamaniego Apologies for the three-week silence — this deserved a faster response.

First: you found a real problem. The PWA shipping with hard-coded Chinese strings was a genuine blind spot, and the QR pairing + the iOS Add-to-Home-Screen fix are both genuinely wanted. Thank you for the careful writeup and for keeping placeholder/keyset parity across locales.

That said, this PR bundles at least seven separately-decidable changes (es locale, PWA i18n architecture, QR pairing, the iOS `start_url` fix, assistant output in the LAN payload, richer push notifications, the mobile mascot, plus SW/cache changes), and most of the mobile side overlaps with work in flight: @Bynlk owns the Mobile / PWA line, and #533 / #494 are about to land changes to the same files. Merging this as one box would either invalidate that queue or force painful rebases on everyone.

Here's the split I'd like, ordered by how fast each can land:

1. **QR pairing** (desktop settings + `qrcode-generator`) — separate small PR; I'll review and land it quickly.
2. **iOS Add-to-Home-Screen fix** — separate PR. The token-echo logic in the manifest route needs a careful security pass, but the bug is real and the fix is wanted.
3. **PWA i18n architecture** (string catalog + follow-desktop-language) — right direction; please coordinate ordering with @Bynlk so it doesn't collide with #533's PermissionBanner rework. Landing after his next slice and rebasing on it is probably smoothest.
4. **Spanish locale (desktop)** — yes, I want this. Please rebase onto current `main` first: the settings surface changed a lot since your branch point (#645–#647), so the `settings-i18n.js` keyset needs a refresh. Don't worry about future upkeep — new UI strings get filled in for all locales as part of normal maintenance here — though as the resident Spanish speaker, your eyes on translation quality will always be welcome.
5. **Assistant output on the phone, push notifications, mobile mascot** — each of these is a product decision, not a patch. Assistant output in particular changes *what travels over the LAN* (conversation content rather than status), which needs a privacy default/toggle story first — I've opened #656 for that discussion and would love your input there. Please hold these three for now.

@Bynlk — flagging this to you since two-thirds of it lives on your turf. The i18n catalog and some of the UX polish here may be worth absorbing into your slice ordering where it helps rather than conflicts; your call on sequencing.

Thanks again — the direction sense here is good, and I'd genuinely like the QR + iOS pieces in soon.

PR #531, opened 2026-06-20
Back up settings.json before injecting hooks

chrono-meta:
### Problem
The hook installer mutates the user's global `~/.claude/settings.json` in place — adding a set of lifecycle hooks plus the blocking HTTP `PermissionRequest` hook — but does not snapshot the prior file. The atomic temp+rename write (`writeJsonAtomic`) protects against a *half-written* file, but it gives the user no way to undo the change, and this is a shared config the user did not author.

### Fix
`writeJsonAtomicWithBackup` / `...Async` already exist in `json-utils.js` — they just weren't wired into the install path (`registerHooks` / `registerHooksAsync` called plain `writeJsonAtomic`). This routes both paths through the backup-aware write, **only when the settings file already existed** (a fresh install has nothing to snapshot).

**Behavior change:** by default, installing over an existing settings file now also writes a one-time backup. Opt out with `backup: false`.

The backup is written next to the settings file as `settings.json.clawd-cleanup-<timestamp>.bak` (reusing the existing `uniqueBackupPath` naming) and its path is printed to the install console — restore by copying it back. The path is also returned as `result.backupPath` (new additive field on the install result).

This may also help with restore asks such as #360, though uninstall/restore UX is out of scope here.

### Tests
Adds a `Hook installer settings backup` suite — existing-file backup captures the original content, no backup for a new file, `backup: false` opt-out, and async parity — and updates the sync/async parity test to compare everything except the environment-specific `backupPath`. `install.test.js`: 59/59 pass.

Unrelated: `test/roam.test.js` is timing-flaky on the full suite (a different sub-test fails per run) on clean `main` too; this change does not touch it.

### DIFF
--- a/hooks/install.js
+++ b/hooks/install.js
@@ -781,8 +781,10 @@ function registerHooks(options = {}) {
 
   // Read existing settings
   let settings = {};
+  let preExisting = false;
   try {
     settings = JSON.parse(fs.readFileSync(settingsPath, "utf-8"));
+    preExisting = true;
   } catch (err) {
     if (err.code !== "ENOENT") {
       throw new Error(`Failed to read settings.json: ${err.message}`);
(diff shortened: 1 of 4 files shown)


rullerzhou-afk:
Thanks for this, @chrono-meta — merged into main! Snapshotting the user's `settings.json` before injecting hooks is exactly the kind of safety net a shared global config deserves.

For transparency on what shifted during review: I added a rolling cap so the backups stay bounded. The settings watcher re-registers hooks whenever another tool (e.g. CC-Switch) strips them, so without a cap the `.bak` files would accumulate indefinitely. I also tightened a few edges around it:

- order backups by their **filename timestamp** rather than `mtime` — `fs.copyFileSync` inherits the source file's `mtime` on Windows, so an mtime-based prune could end up deleting the backup it just created;
- prune only **after** the atomic write succeeds, so a failed write keeps the prior backups;
- a `COPYFILE_EXCL` guard against a copy-name race.

Your backup logic is the foundation it all builds on. Really appreciate the contribution — hope to see more!

chrono-meta:
Thank you for the merge and the detailed explanation. The rolling cap was something I missed, and I’m especially impressed that you caught the issue with ⁠fs.copyFileSync⁠ inheriting the source mtime on Windows — thanks to you, the backup logic is much more robust now. Pruning only after a successful atomic write and adding the ⁠COPYFILE_EXCL⁠ conflict guard are both great touches.

Thank you for the warm welcome to my contribution, and I’ll definitely look for more ways to contribute in the future!

PR #547, opened 2026-06-23
fix: keep mini restore on snapped display

royhuang91:
## Summary

Fix a multi-monitor mini-mode restore issue where Clawd could jump to the neighboring display after exiting edge-snapped mini mode.

## Problem

On a setup with three monitors arranged horizontally, the pet can be snapped to an internal monitor seam while part of its window crosses into the neighboring monitor.

Two reproducible cases:

1. The pet starts on the middle monitor, is dragged to the middle monitor's left edge, and more than half of the pet crosses toward the left monitor. After releasing, it snaps to the middle monitor's left edge. When exiting mini mode, the pet jumps to the left monitor.
2. The pet starts on the right monitor, is dragged to the right monitor's left edge, and more than half of the pet crosses toward the middle monitor. After releasing, it snaps to the right monitor's left edge. When exiting mini mode, the pet jumps to the middle monitor.

The snap target is correct, but restore was using the saved pre-mini position to choose the nearest display. When that saved position's center was over the neighboring display, restore clamped to the wrong screen.

## Fix

Keep the existing restore behavior based on `preMiniX` / `preMiniY`, but when exiting mini mode, clamp the restore target to the display where the pet is currently snapped.

This preserves the original restore position as much as possible while keeping restore ownership tied to the current snapped display.

## Tests

Added regression coverage for three-monitor internal-seam restore cases:

- middle monitor left-edge snap restores onto the middle monitor
- right monitor left-edge snap restores onto the right monitor

Verified with:

```bash
node --test test/mini.test.js test/pet-window-runtime.test.js
git diff --check -- src/mini.js src/pet-window-runtime.js test/mini.test.js

### DIFF
--- a/src/mini.js
+++ b/src/mini.js
@@ -406,11 +406,13 @@ function exitMiniMode() {
   const size = _getSize();
   const visualState = ctx.doNotDisturb ? "idle" : ctx.resolveDisplayState();
   const visualFile = visualState ? ctx.getSvgOverride(visualState) : null;
+  const restoreWorkArea = getAttachedMiniWorkArea();
   const clamped = ctx.clampToScreenVisual(preMiniX, preMiniY, size.width, size.height, {
     state: visualState,
     file: visualFile,
+    workArea: restoreWorkArea,
   });
-  const wa = ctx.getNearestWorkArea(clamped.x + size.width / 2, clamped.y + size.height / 2);
+  const wa = restoreWorkArea || ctx.getNearestWorkArea(clamped.x + size.width / 2, clamped.y + size.height / 2);
   const mEdge = Math.round(size.width * 0.25);
   // Prevent right-edge re-snap
   if (clamped.x >= wa.x + wa.width - size.width + mEdge - SNAP_TOLERANCE) {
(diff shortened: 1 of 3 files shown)


rullerzhou-afk:
Thanks for the great fix — merged! 🎉

Really appreciated the quality here: the root-cause analysis in the description (saved pre-mini center landing on the neighboring display), the defensive fallback in `getAttachedMiniWorkArea()` that reverts to the old behavior when the display layout no longer matches, and the three-monitor regression tests. Everything checked out in review and the full test run.

One ask: my dev machine is single-display, so I could only verify this with the mocked-display tests. Once the next release ships, it would be great if you could re-run your two three-monitor repro scenarios on your real setup and confirm the restore stays on the snapped display. I'll ping you here when it's out.

Thanks again for the contribution!

rullerzhou-afk:
Update: shipped in v0.12.0.

@royhuang91 thanks again for the fix and the thorough root-cause writeup. Once you've updated, could you confirm mini-restore lands correctly on your multi-monitor / snapped-display setup? Want to make sure it holds on real hardware.

PR #548, opened 2026-06-23
fix(state): celebrate bg-only Claude Stop with final assistant text (#406)

rullerzhou-afk:
## What & why

Under Claude Desktop / Code Desktop a user-visible short reply often ends **without** Clawd playing the happy/attention completion animation.

Root cause: the #406 Stop completion gate treats any `background_tasks_count > 0` Stop as live background work and holds the session `"working"` indefinitely. Claude Desktop leaves a `background_tasks` entry attached to the **final** Stop even after the assistant reply is done, so the turn never completes — no happy, no HUD `done` badge, no Telegram push — until stale cleanup drops it back to idle. The user just sees "reply finished, no celebration".

## Fix

Treat a bg-only Stop that **already carries the assistant's final text** as tentative rather than permanent live work:

- **Hard gates unchanged** (always hold `"working"`, never auto-promote):
  `session_crons_count > 0`, `stop_hook_active === true`, and `background_tasks_count > 0` **without** final assistant text.
- **New:** `background_tasks_count > 0` **with** final assistant text gets a 2s quiet window. A forward-progress event (`UserPromptSubmit` / `PreToolUse` / `PostToolUse` / …) in the window cancels the completion; a quiet window promotes it (happy + `done` badge + Telegram push, exactly once).

`CLAWD_COMPLETION_DEBOUNCE_MS` still overrides for every session kind (`0` = celebrate immediately).

## Behavior changes worth noting

- Because the assistant emits prose on nearly every turn, the no-text hard-gate branch only fires for a zero-prose turn — so in practice the bg "hold working" path becomes the 2s-debounce path. This is the intended relaxation of #406 for the spurious-bg case.
- **Headless:** a bg-only Stop with final text now waits 2s instead of holding forever. (Plain headless #449 debounce unchanged.)
- `CLAWD_COMPLETION_DEBOUNCE_MS=0` now makes bg+text celebrate immediately (consistent with "0 = fully off"); crons / stop_hook_active / bg-without-text still hard-gate regardless of the env var.

## Privacy

No new data leaves the hook. The gate reads only `!!assistantLastOutput` (a boolean). `assistantLastOutput` already flowed into the snapshot / Telegram body before this change, and background task command/description are **never** forwarded by the hook.

## Tests

`node --test test/state.test.js test/completion-notify-integration.test.js` → **225 pass / 0 fail**.

Covers both the new quiet-window path and the hard-gate boundaries:
- bg-without-text → hard hold; bg+text → debounce → celebrate; resume cancels inside the window.
- `session_crons` + text, `stop_hook_active` + text, and bg + crons + text → still hard-held (final text does **not** relax them).
- `CLAWD_COMPLETION_DEBOUNCE_MS=0` + bg+text → immediate celebration.
- interactive default 2s window; Telegram pushes exactly once.

## Real-machine smoke ✅

Verified end-to-end against the running app (patched branch, single instance, default port `127.0.0.1:23333`), driving the real `POST /state` hook endpoint with `sendToRenderer` / `playSound` temporarily traced (instrumentation reverted afterwards):

| Scenario | Gate decision (`session-debug.log`) | Renderer received |
| --- | --- | --- |
| bg=1 **+ final text** | `action=debounce-working` | `working` → **+2.0s** → `play-sound complete` + `state-change ["attention","clawd-happy.svg"]` |
| bg=1 **no text** | `action=hold-working` | `working` only — no attention/complete over 3s |
| bg=1 + text, then `PreToolUse` @0.8s | `stop-debounce cancel by=PreToolUse` | no attention/complete (resume cancels) |

The renderer is objectively told to show `clawd-happy.svg` + play `complete` exactly ~2s after the bg-only Stop; the #406 hard gate and the in-window cancel both still behave as designed.

Context: #406 / #421, #449 / #453, #447.

### DIFF
--- a/src/state.js
+++ b/src/state.js
@@ -100,6 +100,11 @@ const COMPLETION_CANCEL_EVENTS = new Set([
 // UserPromptSubmit lands within hook-spawn latency (~0.3s) of the Stop, so a
 // 2s quiet window absorbs it; a genuinely final Stop just celebrates 2s late.
 const HEADLESS_COMPLETION_DEBOUNCE_MS = 2000;
+// Claude Desktop can leave a background_tasks entry attached to a user-visible
+// final Stop even after the assistant reply is complete. Treat that bg-only
+// Stop as tentative, not permanently working, when the hook also extracted the
+// assistant's final text.
+const BACKGROUND_TASKS_COMPLETION_DEBOUNCE_MS = 2000;
 function getCompletionDebounceMs(headless) {
   const raw = process.env.CLAWD_COMPLETION_DEBOUNCE_MS;
   const n = Number.parseInt(raw, 10);
@@ -111,6 +116,12 @@ function getCompletionDebounceMs(headless) {
   // user otherwise wants the celebration the instant the turn ends. Headless
   // sessions default to the #449 window above.
   return headless ? HEADLESS_COMPLETION_DEBOUNCE_MS : 0;
+}
+function getBackgroundTasksCompletionDebounceMs(headless) {
+  const raw = process.env.CLAWD_COMPLETION_DEBOUNCE_MS;
+  const n = Number.parseInt(raw, 10);
+  if (Number.isFinite(n) && n >= 0 && n <= 10000) return n;
+  return Math.max(getCompletionDebounceMs(headless), BACKGROUND_TASKS_COMPLETION_DEBOUNCE_MS);
 }
 let lastSessionSnapshotSignature = null;
 let lastSessionSnapshot = null;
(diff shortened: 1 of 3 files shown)


rullerzhou-afk:
**Real-machine smoke passed ✅** (added a "Real-machine smoke" section to the description with the full evidence).

Ran the patched branch as a live app (single instance, default port `:23333`) and drove the real `POST /state` hook endpoint:

- **bg=1 + final text** → held `working` (`action=debounce-working`), then at **+2.0s** the renderer got `play-sound complete` + `state-change ["attention","clawd-happy.svg"]` → the happy animation fires. ✅
- **bg=1, no text** → `action=hold-working`, no celebration over 3s → #406 protection intact. ✅
- **bg=1 + text, then `PreToolUse` @0.8s** → `stop-debounce cancel by=PreToolUse`, no false happy → in-window resume cancels correctly. ✅

Instrumentation (temporary `sendToRenderer`/`playSound` trace) was reverted; the commit is unchanged. Only the actual on-screen pixels need a human glance, which was confirmed live.

PR #560, opened 2026-06-26
Fix Windows startup recovery followups

rullerzhou-afk:
Summary: async Claude PATH lookup to avoid New Session main-process stalls; debounce Windows display-metrics geometry recovery; re-top hit window after render topmost loss; use hard-coded WMI -Filter for startup recovery. Tests: npm test (4215 pass, 0 fail, 13 skipped). Manual QA: lock/unlock hit-test passed; single-display DPI/scale change passed; Claude New Session no 5s freeze passed. Not covered: multi-monitor hot-unplug/cross-DPI, no multi-display hardware available.

### DIFF
--- a/src/launch-claude.js
+++ b/src/launch-claude.js
@@ -1,6 +1,7 @@
 "use strict";
 
-const { spawn, execFileSync } = require("child_process");
+const { spawn, execFile } = require("child_process");
+const { promisify } = require("util");
 const { platform, homedir } = require("os");
 const path = require("path");
 const fs = require("fs");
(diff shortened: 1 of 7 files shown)


rullerzhou-afk:
Follow-up QA note: during the next macOS real-machine QA pass, verify that after display-metrics changes (resolution/scale or external-display changes), the pet remains visible across Spaces. This PR debounces handleDisplayMetricsChanged by 400ms, which also delays reapplyMacVisibility on the pure display-metrics path.

PR #562, opened 2026-06-26
Float the pet over fullscreen games (#538)

### LINKED ISSUE #538: Bug: Windows hit window steals focus from fullscreen applications / games
## Description
When running a fullscreen application (e.g., a game) on Windows, the pet periodically causes the fullscreen app to lose focus and minimize back to desktop. This happens every ~5 seconds, making it impossible to use fullscreen apps while Clawd is running.

## Root Cause
Two issues in combination:

1. **Hit window is focusable on Windows**  
   `pet-window-runtime.js` line 545: `focusable: !isLinux` — this makes the hit window focusable on Windows. Combined with the topmost watchdog calling `setAlwaysOnTop(true, "pop-up-menu")` every 5 seconds, Windows briefly activates the pet window and deactivates the fullscreen game.

2. **Topmost watchdog runs regardless of foreground app state**  
   `topmost-runtime.js`: the watchdog reasserts `setAlwaysOnTop` every 5 seconds with no regard for whether the user is in a fullscreen app.

## Fix Applied (tested working)
1. Changed hit window to `focusable: false` on all platforms
2. Created `win-fullscreen-detect.js` using koffi to call Windows API (`GetForegroundWindow` + `GetMonitorInfoW`) and detect when the foreground window covers the entire monitor
3. In the topmost watchdog tick, skip `setAlwaysOnTop` when a fullscreen foreground window is detected

## Environment
- OS: Windows 11 Pro (Build 26200)
- Clawd version: 0.10.0 (running from source)

## Additional note
The macOS equivalent (`mac-window.js`) handles fullscreen coexistence extensively via `NSWindowCollectionBehavior`. Windows has no similar handling and relies solely on `"pop-up-menu"` window level, which is not sufficient for fullscreen apps.

rullerzhou-afk:
## What

Fixes #538 — on Windows the pet interrupts fullscreen games/videos, popping back on top roughly every ~5 seconds.

## Root cause

The topmost watchdog (`src/topmost-runtime.js`) re-asserts the pet + hit windows to the `"pop-up-menu"` level every 5s, and `guardAlwaysOnTop` fights back the moment anything else grabs topmost. Neither path is aware of the foreground app, so over a fullscreen game the pet keeps clawing its way back on top every watchdog tick — exactly the "every ~5 seconds" interruption #538 describes.

## Fix

Add a small, degradable Windows probe and make the watchdog stand down when a real fullscreen app owns the foreground.

- **`src/win-fullscreen-detect.js` (new)** — a koffi probe (`GetForegroundWindow` + `GetWindowRect` + `MonitorFromWindow` + `GetMonitorInfoW`) that reports whether the foreground window covers its *whole monitor* (`rcMonitor`, not the work area, so a merely maximized window is **not** treated as fullscreen). The geometry decision is split into a pure `rectCoversMonitor()` for testing.
- **`src/topmost-runtime.js`** — new injected `isForegroundFullscreen` option (defaults to "never fullscreen"). When it returns true, the watchdog tick and the `always-on-top-changed` guard skip the topmost re-assert / nudge for the pet + hit windows, while still running the non-focus-stealing `keepOutOfTaskbar` maintenance. Permission bubbles / HUD keep re-asserting — those are deliberate interruptions the user must act on.
- **`src/main.js`** — wires the real probe in next to the other FFI blocks.

### Why not `focusable: false`

The original #538 sketch proposed flipping `hitWin` to `focusable: false`. That would regress the #545 drag fix (the `WS_EX_NOACTIVATE` input-routing path) and break its `test/pet-window-runtime.test.js:163` contract. This foreground-aware watchdog is the refined version of the same idea — it keeps `focusable: true` untouched and instead removes the *interruption* mechanism.

### Safety

The probe is best-effort: off Windows, or on **any** FFI load/call failure, it returns "never fullscreen", so a broken probe can never hide the pet — it just falls back to today's always-reassert behavior.

## Verification

- `npm test`: **4233 pass / 0 fail**. Covered: pure `rectCoversMonitor` cases (fullscreen / maximized / DPI-tolerance / null) and the full probe decision chain via an injected fake koffi; watchdog + guard standing down under fullscreen while keeping taskbar maintenance; normal reassert unchanged when not fullscreen.
- **Real-machine (Windows 11, 2752×1152 display):**
  - Detection: a fullscreen game reports `win[0,0 → 2752,1152]` == monitor → fires; maximized/normal windows (e.g. `[946,214 → 1804,1104]`) never false-fire.
  - Behavior: launched a real fullscreen game across several attempts — the pet sits quietly in the corner and no longer interrupts gameplay; it returns to top within a watchdog cycle on exiting fullscreen.

## Scope

Size-growth (#408) and the other Windows-audit items are unaffected. This only touches the fullscreen-foreground interruption path.


rullerzhou-afk:
Follow-up after real-machine testing and a codex review pass — three commits on top of the existing #538 stand-down work:

- **`7015e52`** — core: the pet now floats over a foreground fullscreen app (default on) instead of standing down, staying draggable and never stealing the game's focus.
- **`36665ed`** — fast focusable poll: the hit window's activation used to ride the 5s topmost watchdog, leaving up to ~5s after entering fullscreen where an early drag could kick an exclusive-fullscreen game out. It now runs on its own ~1s poll, decoupled from the topmost/float decision, and syncs once up front so starting while already fullscreen is covered immediately.
- **`91bdde6`** — overlay toggle removed from Settings, the pref kept as an escape hatch: for borderless-fullscreen games (the common case) turning it off can't drop the pet behind the game anyway (a Windows limit), so it was a non-choice. The pref (default on), the stand-down logic, and the i18n strings all stay, so the toggle can be restored in one line if exclusive-fullscreen games ever need it.

Verified: borderless-fullscreen drag-over works on real hardware; `node --test test/*.test.js` → 4242 pass / 0 fail. The codex review found no Blocker/High; its Medium (up-front sync) and the Low/Nit (conservative help text; drop the now-unused pref from the in-place re-render set) are folded into the commits above.

Known gap: behavior over exclusive-fullscreen (older true-fullscreen) games is unverified — which is exactly why the pref stays as an escape hatch rather than being deleted.

PR #582, opened 2026-07-02
fix(antigravity): stop first-token stall when the hook runner never closes stdin (#568)

### LINKED ISSUE #568: Antigravity IDE/App hook can add ~10s+ first-token latency before model request
## Summary

When the Clawd Antigravity hook is installed via `~/.gemini/config/hooks.json`, local Antigravity IDE/App conversations on macOS can wait roughly 10 seconds before the model request is sent. Disabling the hook immediately restores normal first-token latency.

This looks specific to the local Antigravity IDE/App hook runner path. Antigravity CLI and Antigravity IDE SSH conversations were normal in the same environment.

## Environment

- Platform: macOS on MacBook
- Product: Antigravity 2.0 App / Antigravity IDE local conversations
- Clawd hook file involved: `~/.gemini/config/hooks.json`
- Hook target: `/Applications/Clawd on Desk.app/Contents/Resources/app.asar.unpacked/hooks/antigravity-hook.js`

## Reproduction / control tests

Observed behavior:

| Environment | Result |
| --- | --- |
| Same Mac, local Antigravity 2.0 App / IDE conversation | First-token latency around 20s |
| Same Mac, Antigravity CLI | Normal |
| Same Mac, Antigravity IDE SSH conversation | Normal |
| Same network, Linux Antigravity 2.0 App | Normal |
| Same Mac, after moving `~/.gemini/config/hooks.json` away | Normal |

The last row is the key confirmation: moving `hooks.json` out of the way and restarting Antigravity fixed the delay immediately.

## Log evidence

In Antigravity IDE logs, there was a gap between the Clawd `PreInvocation` hook and the actual model request:

```text
13:16:35.921 jsonhook__clawd_PreInvocation_0_0 executing command
13:16:46.884 URL: https://daily-cloudcode-pa.googleapis.com/v1internal:streamGenerateContent?alt=sse
```

That is about 11 seconds before `streamGenerateContent` is even requested, so this does not look like model/backend latency. It appears to be a local preflight hook delay.

## Suspected cause

The generated shell wrapper in `hooks.json` starts by reading stdin into a temp file, roughly:

```sh
cat > "$in_file"
```

For the local Antigravity IDE/App hook runner, stdin may not be closed promptly, or the payload/runtime behavior may differ from the CLI path. Since `PreInvocation` is synchronous and runs before the model request, waiting for hook timeout directly delays first-token latency.

This may be a CLI-oriented Antigravity adapter working fine for CLI usage, but not being safe for local IDE/App conversations where hook stdin lifecycle / timeout behavior differs.

## Workaround

Remove or move the Clawd Antigravity hook config:

```sh
mv ~/.gemini/config/hooks.json ~/.gemini/config/hooks.json.disabled
```

Then fully restart Antigravity / Antigravity IDE.

## Suggested fix direction

A few possible mitigations:

1. Avoid installing `PreInvocation` for Antigravity IDE/App by default, or make it opt-in.
2. Add a hard stdin read timeout / nonblocking fail-open behavior before `cat > "$in_file"` can hit the IDE hook timeout.
3. Make state reporting fully asynchronous for non-permission hooks, so it cannot block model request startup.
4. Detect IDE/App vs CLI paths and use a safer adapter for IDE/App.

The important part is that a Clawd state hook should never block Antigravity from sending the model request.

rullerzhou-afk:
Fixes #568

## Root cause

The Antigravity IDE/App (macOS local chat) hook runner keeps the hook's stdin open, unlike the CLI which closes it promptly. The fail-open wrapper read stdin in the foreground (`cat > "$in_file"`), **outside** the watchdog that only guarded the node child, so the wrapper blocked until the 10s `hooks.json` timeout killed it — delaying the first token by 10s+ and dropping the fallback stdout on the floor. This matches the ~11s gap between `PreInvocation` and `streamGenerateContent` in the reporter's logs.

The Windows command had the same latent bug in `[Console]::In.ReadToEnd()` (unbounded synchronous read); nobody had hit it yet.

## The fix

**POSIX** — the stdin read now runs as a background `cat` under its own 2s watchdog. Two non-obvious details, both verified in a real shell:

- Background lists get `/dev/null` as stdin in non-interactive shells (a bare `cat > f &` reads 0 bytes), so a `3<&0` group redirection hands the real stdin to the background cat. Chosen over `exec 3<&0`: if fd 0 is somehow absent, the group redirection fails soft (pid stays empty, the empty `in_file` walks the existing fail-open path) while a failed `exec` would abort a POSIX shell entirely.
- When the watchdog cuts the read, whatever bytes already arrived stay in `in_file`; the node hook already fails open on partial/empty JSON. In the common IDE case (payload written, fd just never closed) the full payload survives and state reporting keeps working.

**Windows** — `[Console]::In` is a `SyncTextReader` whose `ReadToEndAsync()` executes synchronously (the naive `ReadToEndAsync().Wait(ms)` fix still blocked in testing). The fix reads the raw stdin stream via `StreamReader([Console]::OpenStandardInput())`, whose async read honors `Wait(2000)`. On timeout the payload is dropped (no partial reads without EOF) and the hook fails open on an empty payload.

**Timeout budget** — the child watchdog shrinks 8s → 6s so stdin (2s) + child (6s) stays under the outer 10s hook timeout with ~2s of headroom for shell/PowerShell startup; without headroom the outer timeout could kill the wrapper before the fallback line is printed. Measured worst path (never-closed stdin + hung child): 9.5-9.7s at 2+7 — too close to 10s for a PowerShell cold start under AV scanning — vs ~8.3s at 2+6 on both platforms. A test guards the budget at ≥2s headroom.

**Watchdog hardening** — both watchdog subshells now redirect stdout/stderr to `/dev/null`. A killed watchdog orphans its `sleep`, and an orphan holding the wrapper's stdout would stall any hook runner that waits for pipe EOF rather than process exit — which would have re-introduced a 2s stall via the new stdin watchdog.

## Verification

Repro harness: spawn the generated wrapper with a stdin pipe that is written to (or not) but never closed, the way the IDE runner behaves.

| Scenario | Before | After |
|---|---|---|
| POSIX, stdin never closed | blocked until pipe EOF / outer timeout (14.2s in harness) | **2.37s**, payload delivered intact |
| POSIX, stdin closed promptly (CLI) | ~0.3s | 0.41s — no regression |
| Windows, stdin never closed | blocked (killed at 12s backstop) | **2.37s**, fails open on empty payload |
| Windows, stdin closed promptly | 372ms | 365ms — no regression |

- Full suite: `node --test test/*.test.js` — 4369 tests, 0 failures.
- New tests: never-close and write-without-close repros on both platforms (async spawn with an open stdin pipe — `spawnSync` without `input` closes stdin immediately and cannot model the IDE runner), structural asserts on the generated commands, elapsed lower bounds so the repros cannot silently degrade into close-path tests, and a budget guard.
- Upgrade path: an existing pre-fix install is auto-rewritten on the next startup sync (`updated: 4`) on both platforms; users only need the new version to start once.

## Known tradeoffs

- When the runner never closes stdin, each blocking hook event still pays up to the 2s stdin watchdog (down from 10s+). The 2s floor is the fail-safe read window, not eliminable without a protocol change.
- Windows only: a payload that arrives without EOF is dropped at the timeout (the async read cannot surface partial data), so the hook falls back to an empty payload. POSIX keeps the arrived bytes.
- `StreamReader` decodes stdin as UTF-8 instead of the console code page — for JSON piped between processes this is a correctness improvement for non-ASCII payloads.

## Status

- [x] Windows behavior verified on a real machine (this table)
- [x] Codex review round done — both findings addressed in 5ed5eac (budget headroom 7s→6s; tests now bound stdout-EOF time, not just exit, and pin both watchdog redirects)
- [x] macOS regression verified on a real machine (2026-07-07): POSIX-only tests green (real runs), CLI path unaffected, live-IDE hook overhead 0–1s on Antigravity 2.2.1, pet state updates flowing, upgrade path exercised — details in the verification comment below

All verification gates are green. Note from the macOS pass: Antigravity 2.2.1's runner closes stdin promptly, so the live stall only reproduces on older builds (or via the never-close harness).

### DIFF
--- a/hooks/antigravity-install.js
+++ b/hooks/antigravity-install.js
@@ -34,7 +34,11 @@ const ANTIGRAVITY_HOOK_EVENTS = [
   "Stop",
 ];
 const DEFAULT_HOOK_TIMEOUT_SECONDS = 10;
-const FAIL_OPEN_CHILD_TIMEOUT_SECONDS = 8;
+// #568 budget: stdin timeout + child timeout must stay below
+// DEFAULT_HOOK_TIMEOUT_SECONDS with headroom for shell startup, or the outer
+// hooks.json timeout kills the wrapper before the fallback line is printed.
+const FAIL_OPEN_CHILD_TIMEOUT_SECONDS = 7;
+const FAIL_OPEN_STDIN_TIMEOUT_SECONDS = 2;
 
 function fallbackStdoutForEvent(event) {
   return stdoutForAntigravityEvent(event);
@@ -52,6 +56,12 @@ function normalizeFailOpenTimeoutSeconds(options = {}) {
   const raw = Number(options.failOpenTimeoutSeconds);
   if (Number.isFinite(raw) && raw > 0) return Math.max(1, Math.floor(raw));
   return FAIL_OPEN_CHILD_TIMEOUT_SECONDS;
+}
+
+function normalizeStdinTimeoutSeconds(options = {}) {
+  const raw = Number(options.stdinTimeoutSeconds);
+  if (Number.isFinite(raw) && raw > 0) return Math.max(1, Math.floor(raw));
+  return FAIL_OPEN_STDIN_TIMEOUT_SECONDS;
 }
 
 function quoteWindowsProcessArg(value) {
@@ -82,6 +92,7 @@ function quoteWindowsProcessArg(value) {
 function withFailOpenShellFallback(command, event, nodeBin, options = {}) {
   const fallback = quoteShellSingleArg(fallbackStdoutForEvent(event));
   const timeoutSeconds = normalizeFailOpenTimeoutSeconds(options);
+  const stdinTimeoutSeconds = normalizeStdinTimeoutSeconds(options);
   const validatorScript = [
     "let s='';",
     "process.stdin.setEncoding('utf8');",
@@ -105,9 +116,22 @@ function withFailOpenShellFallback(command, event, nodeBin, options = {}) {
     // Do not trap TERM: macOS bash 3.2 may print run_pending_traps warnings when the watchdog is killed.
     "cleanup(){ trap - EXIT HUP INT TERM; [ -n \"$watchdog\" ] && kill \"$watchdog\" 2>/dev/null; [ -n \"$pid\" ] && kill \"$pid\" 2>/dev/null; rm -f \"$in_file\" \"$out_file\"; }",
     "trap cleanup EXIT HUP INT",
-    "cat > \"$in_file\" 2>/dev/null || :",
+    // #568: IDE/App hook runners may never close our stdin, so the stdin read
+    // needs its own watchdog. Background lists get /dev/null as stdin in
+    // non-interactive shells; the 3<&0 group redirection hands the real stdin
+    // to the background cat (and fails soft if fd 0 is somehow absent).
+    "{ cat <&3 > \"$in_file\" 2>/dev/null & pid=$!; } 3<&0",
+    // Watchdog subshells redirect stdout/stderr: a killed watchdog orphans its
+    // sleep, and an orphan holding our stdout would stall a hook runner that
+    // waits for pipe EOF instead of process exit.
+    `( sleep ${stdinTimeoutSeconds}; kill "$pid" 2>/dev/null ) > /dev/null 2>&1 & watchdog=$!`,
+    "wait \"$pid\" 2>/dev/null",
+    "[ -n \"$watchdog\" ] && kill \"$watchdog\" 2>/dev/null",
+    "[ -n \"$watchdog\" ] && wait \"$watchdog\" 2>/dev/null",
+    "pid=",
+    "watchdog=",
     `${command} < "$in_file" > "$out_file" 2>/dev/null & pid=$!`,
-    `( sleep ${timeoutSeconds}; kill "$pid" 2>/dev/null ) & watchdog=$!`,
+    `( sleep ${timeoutSeconds}; kill "$pid" 2>/dev/null ) > /dev/null 2>&1 & watchdog=$!`,
     "wait \"$pid\" 2>/dev/null",
     "status=$?",
     "[ -n \"$watchdog\" ] && kill \"$watchdog\" 2>/dev/null",
@@ -123,6 +147,7 @@ function withFailOpenShellFallback(command, event, nodeBin, options = {}) {
 function buildWindowsEncodedFailOpenNodeHookCommand(nodeBin, hookScript, event, options = {}) {
   const fallback = fallbackStdoutForEvent(event);
   const timeoutMs = normalizeFailOpenTimeoutSeconds(options) * 1000;
+  const stdinTimeoutMs = normalizeStdinTimeoutSeconds(options) * 1000;
   const childArgs = [
     quoteWindowsProcessArg(hookScript),
     quoteWindowsProcessArg(event),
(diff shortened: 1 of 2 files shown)


rullerzhou-afk:
Codex review round — both findings confirmed and fixed in 5ed5eac:

**Finding 1 (medium): 2s+7s budget too close to the outer 10s.** Reviewer measured the worst path (never-closed stdin + hung child) at 9529ms on Windows PowerShell 5.1 and 9710ms on pwsh 7.6 — fine warm, but a PowerShell cold start under AV scanning would cross 10s, the outer runner would kill the wrapper, and the fallback line would be lost. Fixed: child watchdog 7s → 6s on both platforms (POSIX gains nothing from staying higher — the child hook normally finishes in tens of ms). Re-measured worst path at 2+6: 8342ms (win32) / 8270ms (POSIX), ~1.7s warm headroom. The budget guard test now requires ≥2s.

**Finding 2 (low/medium): never-close tests awaited exit, not stdout EOF.** Exit alone could not catch a stripped watchdog redirect — the orphaned sleep keeps the wrapper's stdout open past exit and stalls an EOF-waiting hook runner. Fixed: the repro helper now resolves only after exit AND stdout close, and every never-close test bounds the close time. Verified the assertion actually bites: with the redirects stripped, exit lands at 1.3s but stdout EOF at 7.2s, which trips the new `< 6000ms` bound. The structural test also pins the child watchdog redirect (previously only the stdin one) and rejects any unredirected watchdog subshell.

Reviewer also confirmed: the fd-3 group redirection is sound on macOS bash-as-sh and dash, empty-pid `wait` degrades safely, partial `cat` writes fail open through the child hook, and the PowerShell StreamReader timeout behaves identically on 5.1 and 7.6 with no exception escaping the local catch.

Full suite still green after the changes (4369 tests, 0 failures).

rullerzhou-afk:
## macOS real-machine pass — done (2026-07-07)

Environment: macOS 15 (Apple Silicon), Antigravity 2.2.1 (Homebrew cask), hooks installed from this branch (head 5ed5eac) into `~/.gemini/config/hooks.json`, resident pet running from the same checkout.

| Check | Result |
|---|---|
| POSIX-only tests | ✅ Real runs, not skipped: never-close repro 1095ms, write-without-close 1091ms; antigravity suite 25 pass / 0 fail (5 Windows-only skips) |
| Full suite | ✅ 4341 / 0 on the branch. `main` has since touched the same file (`b08556d2`), so I also merged main into a throwaway branch (clean merge) and re-ran: 4637 / 0 |
| Harness vs the actually-installed wrapper | Never-close stdin: the pre-fix wrapper blocked past 12s with **empty stdout**; this branch's wrapper exits in **2126ms with stdout intact**. Closed-promptly: 87ms — no regression |
| Live IDE first-token latency | ✅ Wrapped all 4 installed hook commands in a timing probe (subshell + timestamps, command untouched); across 3 real chat turns every event completed in 0–1s |
| Pet state updates | ✅ Full chain lands in session-debug.log: `thinking/UserPromptSubmit → idle/AfterAgent → attention/Stop`, PID resolution working. (Requires the Antigravity integration to be enabled in settings — disabled agents are dropped with a 204 by design) |
| Upgrade path | ✅ A pre-existing install was rewritten in place on the next sync (`updated: 4`), matching the description above |

### Two findings worth recording

1. **Antigravity 2.2.1's hook runner closes stdin promptly.** All three probed turns showed 0s hook overhead *even on the pre-fix wrapper*, and the stall only reproduces under the never-close harness. So #568's live-IDE stall does not manifest on current Antigravity — the reporter was presumably on an older build whose runner held stdin open. The fix is still warranted (older installs in the wild, defense in depth against runner regressions, and the same latent bug on Windows), but anyone trying to repro on 2.2.1 should use the harness, not the live IDE.
2. **The first turn of a fresh conversation pays ~8s of agy-internal setup** (a `CONVERSATION_HISTORY` assembly step, visible in the brain transcript) that is easy to misread as the hook stall. Later turns don't pay it. I initially blamed the wrapper for exactly this gap.

All gates green — un-drafting.

PR #584, opened 2026-07-02
fix(codex): don't replay attention from a metadata-only token_count write

### LINKED ISSUE #566: [Bug] 在同时配置 Claude Code 和 Codex 时，Codex 空闲会话有时会触发 thinking/working 状态
### 问题描述

我安装了 Clawd on Desk，并同时启用了 Claude Code CLI和 Codex Desktop的集成。在这种多 Agent 的配置下，我发现有时仅仅进入或启动 Codex 时，即使没有执行任何任务，Clawd 也会自动切换到 thinking 或 working 动画状态。

这会导致桌宠显示为“正在工作”的状态，但实际上 Codex 会话是空闲的。

### 复现步骤

1. 安装并启动 Clawd on Desk。
2. 启用 / 配置 Claude Code 和 Codex 集成。
3. 打开终端并进入 Codex。
4. 不提交任何 prompt，也不执行任何 Codex 任务。
5. 观察 Clawd 的动画状态。

### 实际行为

有时在进入 Codex 后，Clawd 会切换到 thinking / working 状态，即使 Codex 并没有执行任何工作。

### 期望行为

当仅仅打开或进入 Codex，但没有提交 prompt、调用工具或执行任何实际任务时，Clawd 应保持空闲状态。

thinking / working 动画应仅在 Codex 有真实活动时触发，例如提交 prompt、模型生成、工具调用或其他明确的活跃生命周期事件。

### 环境信息

* 操作系统： [macOS Sonoma 14.4]
* Clawd on Desk 版本： [v0.10.0]
* Codex Desktop 版本： [26.623.31921]
* Claude Code CLI版本： [v2.1.183]

### 补充说明

该问题出现在同时配置 Claude Code 和 Codex 的环境中。我不确定错误的活跃状态是否来源于以下原因：

* Codex 启动或会话初始化被误判为正在工作；
* 读取到了过期的 Codex 会话或日志状态；
* Claude Code 和 Codex 会话共同影响了全局最高优先级状态的判断；
* 存在未正确清理的孤立或过期会话。

我发现一个相关问题（#535），其中提到在没有会话时 Clawd 会随机显示活跃或完成状态，但我的情况更具体，是在 Claude Code + Codex 多 Agent 配置下，仅进入 Codex 就触发状态变化。

rullerzhou-afk:
Refs #535.

Codex Desktop rewrites `token_count` on focus, long after a session has gone idle. The monitor carried `tracked.lastState` forward verbatim on that write:

```js
this._emitStateChange(tracked, tracked.lastState || "idle", key);
```

For a session whose last state was the one-shot `attention`, that re-announces the finished turn. **The pet celebrates work the user already watched complete, with nothing new behind it** — a metadata refresh, animated as a turn ending.

## Why `preserveState` doesn't already cover this

The main-process consumer treats `token_count` as metadata-only, so this looked like defense in depth. It isn't. `preserveState` pins the **stored** state only:

```js
// src/state.js
const preservedState = preserveState && existing ? existing.state : null;
```

while the one-shot branch plays whatever state it was handed, bypassing `resolveDisplayState()` entirely:

```js
// src/state.js
if (ONESHOT_STATES.has(state)) { … }
```

`attention` is in `ONESHOT_STATE_NAMES`. So a session stored as idle still animates. The carry has to be filtered at the source.

## Fix

`token_count` carries only sustained states (`thinking` / `working`) forward; a one-shot re-emits as `idle`. `BACKFILL_SNAPSHOT_STATES` → `SUSTAINED_ACTIVE_STATES`, since the same set now answers both "may the backfill snapshot synthesize this?" and "may a metadata write carry this?" — one question, one set.

## Verification

Two tests. `test/codex-log-monitor.test.js` locks the fix — confirmed to fail on `main`, where the second `_pollFile` re-emits `attention`.

`test/state.test.js` pins the premise underneath it, since that premise lives in another file and the monitor depends on it silently. Driving `state.js` directly with an idle session and a metadata-only `token_count` carrying `attention`:

```
carried state = attention  →  renderer told to play ["attention"]   ← the bug
carried state = idle       →  renderer told nothing                 ← the fix
```

`preserveState: true` is passed in both. It pins the stored state and does nothing for the visual, which is the whole reason the carry has to be filtered upstream. The test is characterization, not endorsement: if `preserveState` ever grows to cover one-shots it will fail, and whoever does that learns the monitor's filter became redundant instead of finding out by deleting it. Verified to discriminate — hand the same call `idle` and the assertion fails.

Full suite on the rebased branch: **4966 pass / 0 fail** (4984 total, 18 skipped).

## Scope — this PR used to be about #566

Earlier revisions tried to fix #566 (a finished Codex session replaying as live `thinking → working → attention`) by re-anchoring both replay guards on per-file attach time instead of monitor start, plus a filename-staleness signal. **That work is withdrawn.** Review found two regressions that no amount of tuning removes:

- **A short turn discovered late is swallowed whole.** A turn finishing >5s before the monitor first sees the file reads as history and emits nothing, where `main` emits the full turn. Reachable within the ordinary discovery latency (~1.5s when the day dir is already scanned, up to ~6.5s via the `_getActiveDayDirs` cache).
- **The restart-replay window widens 1.5s → 5s.** Restarting inside it re-emits a turn the pet already showed.

They are the same contradiction from two sides. Both guards answer "is this content history?" by comparing timestamps, and no threshold separates *"a turn we never saw"* from *"a turn we saw, then forgot"* — **the two produce byte-identical files**. A wider window swallows less and replays more; a narrower one does the reverse.

The filename signal was also unnecessary. Its stated justification was timestamp-less historical lines defeating the line-level guard — but every state-bearing line in 209k lines of real rollouts across 383 files carries a timestamp, and upstream introduced the `RolloutItem` envelope this monitor parses *in the same commit* as per-line `RolloutLine.timestamp`. A recognizable line without a timestamp isn't rare; it's structurally impossible.

#566 needs bookkeeping, not inference: the monitor already knows how far it has read, but that offset rides the `_tracked`/`_retiredTracked` eviction and is lost once ~150 rollouts of churn push it out. Tracked separately — see the follow-up issue.

## Not included

The `SUSTAINED_ACTIVE_STATES` set is shared with `_emitBackfillSnapshot`, which is untouched. #660 rewrites this same `token_count` branch to attach quota data and keeps `tracked.lastState || "idle"` verbatim; whoever lands second needs to keep the `carry`, which #660's own tests would not catch:

```js
const carry = SUSTAINED_ACTIVE_STATES.has(tracked.lastState) ? tracked.lastState : "idle";
this._emitStateChange(tracked, carry, key, codexQuota ? { codexQuota } : null);
```

### DIFF
--- a/agents/codex-log-monitor.js
+++ b/agents/codex-log-monitor.js
@@ -38,7 +38,11 @@ const ACTIVE_SESSION_WINDOW_MS = 5 * 60 * 1000;
 // replay it silently (backfill) instead of emitting stale transitions. A
 // file written within the grace window is a live session and emits normally.
 const BACKFILL_GRACE_MS = 5 * 1000;
-const BACKFILL_SNAPSHOT_STATES = new Set(["thinking", "working"]);
+// States that are ongoing rather than one-shot. Safe to re-synthesize from a
+// backfill snapshot, and safe for a metadata-only token_count write to carry
+// forward. A one-shot (attention, sweeping, …) must never be carried by
+// either: re-emitting it replays a finished turn's celebration.
+const SUSTAINED_ACTIVE_STATES = new Set(["thinking", "working"]);
 
 function finiteNonnegativeNumber(value) {
   const n = Number(value);
@@ -438,7 +442,23 @@ class CodexLogMonitor {
       if (contextUsage) {
         tracked.contextUsage = contextUsage;
         if (!tracked.backfilling) {
-          this._emitStateChange(tracked, tracked.lastState || "idle", key);
+          // token_count is a metadata refresh, not a turn boundary: Codex
+          // Desktop rewrites it on focus long after a session went idle.
+          // Carrying lastState verbatim therefore re-announces whatever the
+          // session last did, and for a finished turn that is the one-shot
+          // `attention` — the pet celebrates a turn the user already saw
+          // complete, with no new work behind it (#535).
+          //
+          // preserveState does NOT cover this. It only pins the *stored*
+          // state (src/state.js: `preservedState = preserveState && existing
+          // ? existing.state : null`); the one-shot branch keys off the state
+          // passed in and plays it regardless, bypassing resolveDisplayState.
+          // So a stored-idle session still animates. The carry has to be
+          // filtered here.
+          const carry = SUSTAINED_ACTIVE_STATES.has(tracked.lastState)
+            ? tracked.lastState
+            : "idle";
+          this._emitStateChange(tracked, carry, key);
         }
       }
       return;
@@ -635,7 +655,7 @@ class CodexLogMonitor {
 
   _emitBackfillSnapshot(tracked) {
     const snapshotState = tracked.lastState;
-    if (!BACKFILL_SNAPSHOT_STATES.has(snapshotState)) {
+    if (!SUSTAINED_ACTIVE_STATES.has(snapshotState)) {
       if (tracked.contextUsage) {
         this._emitStateChange(tracked, "idle", "event_msg:token_count");
       }
--- a/test/codex-log-monitor.test.js
+++ b/test/codex-log-monitor.test.js
@@ -487,6 +487,50 @@ describe("CodexLogMonitor", () => {
     });
   });
 
+  it("does not replay attention from a metadata-only token_count write (#535)", () => {
+    // token_count is a metadata refresh, not a turn boundary — Codex Desktop
+    // rewrites it on focus, long after a session went idle. Carrying lastState
+    // verbatim re-announces the finished turn's one-shot `attention`, and the
+    // pet celebrates work the user already watched complete.
+    //
+    // The consumer's preserveState does not save us: it pins the stored state
+    // only, while state.js's one-shot branch plays whatever state it is handed
+    // (`attention` is in ONESHOT_STATE_NAMES). So a stored-idle session still
+    // animates unless the carry is filtered here.
+    const testFile = path.join(dateDir, TEST_FILENAME);
+    fs.writeFileSync(testFile, [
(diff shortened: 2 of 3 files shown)


rullerzhou-afk:
Codex review surfaced two confirmed findings against the first commit; both fixed in 9d27109.

**Finding 1 (medium): timestamp-less stale lines still replayed in the touched-rollout shape.** The line-level guard only inspects lines that carry a `timestamp` field. Fix: the file-level gate now also treats a rollout as silent backfill when its **filename creation stamp** is older than the active-session window, regardless of mtime — an old name with a fresh mtime is exactly the touched-stale shape, and the creation stamp is the one signal it can't fake. Works for any line shape.

**Finding 2 (medium, regression introduced by the first commit): legitimate late-arriving lines swallowed.** Long-lived Codex Desktop sessions live in old day dirs; discovery of a re-activating write lags behind (active-dir scan cache 5s + poll 1.5s), so a fresh `task_started` could land beyond the 5s line-guard margin and get skipped — pet stays idle while the user works. Fix: those sessions now enter the backfill path (old filename), where the drain is silent but `_emitBackfillSnapshot` restores the current sustained state, so the pet still shows thinking/working. Regression test pins it.

**Attribution correction from review:** the `token_count → attention` replay path I blamed for part of #535 is already neutralized in the main-process consumer (`isCodexMonitorMetadataOnlyEvent` → `preserveState: true`). The sustained-only carry stays as defense in depth for other consumers, but #535's real explanation is the main replay chain (a replayed session ending on `task_complete` → attention). Comment updated accordingly.

Both new regression tests fail on the first commit and pass here; full suite 4356 pass / 0 fail.

PR #590, opened 2026-07-02
fix(context-usage): correct Claude Code 1M detection; add Antigravity context + account-usage bars

### LINKED ISSUE #534: dashboard中的会话上下文上限判定出错
来自于v0.10.0的New Features：
Claude & Codex context usage (https://github.com/rullerzhou-afk/clawd-on-desk/pull/397) - Sessions can surface Claude and Codex context-window usage. Thanks to first-time contributor @Tsdsj.

image
image
我的模型是Sonnet 4.6 (1M context)，但是该功能下的上下文检测依然判定为200000。

Yike-Ye:
## Summary

Fixes #534 (Claude Code's context-limit detection bug), then extends into context/quota tracking for Antigravity and an account-usage summary for both, once real Anthropic docs/community sources confirmed the underlying data actually exists.

- **Root cause fix (Claude Code)**: `resolveClaudeContextLimit` (`hooks/context-usage.js`, used via `clawd-hook.js` reading the local transcript) matched a `[1m]` marker that Claude Code never actually writes to `message.model`, so every session fell back to the 200k default even on models with a real 1M-token context window. Replaced with an explicit model-name table sourced from Anthropic's docs (Opus 4.6+, Sonnet 4.6+/5, Fable 5, Mythos default to 1M).
- **New (Antigravity context usage)**: Antigravity had no context-usage tracking at all before this — the dashboard only showed it for Claude Code. Added `hooks/antigravity-context-usage.js` (`resolveAntigravityContextUsage`), which reads `context_window` straight off Antigravity's own statusline JSON — no model-name table needed there, since Antigravity already reports the real window size and fill level itself.
- **New (account-usage bars)**: a dashboard summary bar showing account-wide rate-limit quota — the same numbers `/usage` shows in Antigravity, and Claude Code's own `rate_limits.five_hour`/`seven_day` (shipped since Claude Code v2.1.80). Grouped by source, each section only appears once that CLI has actually reported data:
  - Antigravity: Gemini and Claude/GPT-via-agy, each showing 5h + weekly
  - Claude Code: 5h + weekly
- New `hooks/antigravity-statusline.js` and `hooks/claude-statusline.js` register as each CLI's `statusLine` command — never overwriting a pre-existing third-party statusline, only taking an empty or already-ours slot.
- Not wired into remote SSH deploy — deliberately left as a known gap (statuslines are inherently local-terminal; the scripts already support `CLAWD_REMOTE`/`--host`, but the deploy manifest and `install.js --remote` path weren't extended).

### References / credit
- [ag-donald/Antigravity-Database-Manager](https://github.com/ag-donald/Antigravity-Database-Manager) — reverse-engineered Antigravity's local SQLite/protobuf trajectory storage; used during investigation (`protoc --decode_raw` against the `steps` table) before settling on the statusline JSON approach used here.
- [weby-homelab/antigravity-cli-statusline](https://github.com/weby-homelab/antigravity-cli-statusline) — source for the statusline JSON field names (`context_window`, `quota`) this implementation is built against.
- [Claude Code statusline docs](https://code.claude.com/docs/en/statusline) — official `rate_limits` schema for the Claude Code side.

## Test plan

- [x] `npm test` — 4425 passing, 0 failing, 16 skipped (Windows-only)
- [x] Manually verified end-to-end against a running dev build: real Antigravity and Claude Code sessions correctly populate the dashboard's context/quota bars, independently show/hide per source, and the 5h vs weekly reset formatting (relative countdown vs absolute date) renders correctly
- [x] Verified the statusline installers never overwrite a pre-existing third-party `statusLine` in `~/.gemini/antigravity-cli/settings.json` / `~/.claude/settings.json`
- [x] Confirmed no regressions in the remote/SSH test suite (297 tests, 296 passing, 1 skipped)

--------------------------------------------------------------------

## 概述

修复 #534(Claude Code 自己的上下文上限判定错误的 bug),然后在查证过程中确认了 Anthropic 官方文档和社区资料里其实已经有真实数据支撑后,顺带把 Antigravity 的上下文用量追踪、以及两边的账号用量汇总读条也做了进去。

- **根因修复(Claude Code)**:`resolveClaudeContextLimit`(`hooks/context-usage.js`,通过 `clawd-hook.js` 读本地 transcript)原来是在找一个 Claude Code 从来不会写进 `message.model` 字段的 `[1m]` 标记,导致所有会话都被判成默认的 200000,哪怕模型本身就是真正的 1M 上下文窗口。现在换成了显式的模型名对照表(数据来自 Anthropic 官方文档:Opus 4.6+、Sonnet 4.6+/5、Fable 5、Mythos 默认就是 1M)。
- **新增(Antigravity 上下文用量)**:Antigravity 在这次改动之前完全没有上下文用量追踪——dashboard 之前只给 Claude Code 显示这个信息。新增了 `hooks/antigravity-context-usage.js`(`resolveAntigravityContextUsage`),直接从 Antigravity 自己的 statusline JSON 里读 `context_window` 字段——这边不需要模型名对照表,因为 Antigravity 自己就会把真实的上下文窗口大小和占用比例报出来。
- **新增(账号用量读条)**:dashboard 顶部的账号级用量配额读条——跟 Antigravity `/usage` 命令显示的是同一份数字,以及 Claude Code 自己的 `rate_limits.five_hour`/`seven_day`(2.1.80 版本起就有)。按数据来源分组,每个分区只有对应 CLI 真的上报过数据才会显示:
  - Antigravity:Gemini,以及经由 agy 统计的 Claude/GPT,各自显示 5 小时 + 每周
  - Claude Code:5 小时 + 每周
- 新增 `hooks/antigravity-statusline.js` 和 `hooks/claude-statusline.js`,分别注册成两个 CLI 各自的 `statusLine` 命令——绝不覆盖用户已有的第三方 statusline,只在配置为空或者已经是我们自己装的时候才接管。
- 没有接入远程 SSH 部署流程——这是有意识留下的已知缺口(statusline 本质上是本地终端的概念;脚本本身已经支持 `CLAWD_REMOTE`/`--host`,但部署清单和 `install.js --remote` 路径没有跟着扩展)。

### 参考 / 致谢
- [ag-donald/Antigravity-Database-Manager](https://github.com/ag-donald/Antigravity-Database-Manager) —— 逆向了 Antigravity 本地 SQLite/protobuf 会话存储的社区项目;排查过程中用它对着 `steps` 表跑过 `protoc --decode_raw`,后来才改用现在这套 statusline JSON 方案。
- [weby-homelab/antigravity-cli-statusline](https://github.com/weby-homelab/antigravity-cli-statusline) —— 这个项目的脚本源码给出了 statusline JSON 里 `context_window`、`quota` 等字段的确切命名,本次实现就是照这个字段结构对接的。
- [Claude Code statusline 官方文档](https://code.claude.com/docs/en/statusline) —— `rate_limits` 字段结构的出处。

## 测试情况

- [x] `npm test` —— 4425 个通过,0 个失败,16 个跳过(仅 Windows 相关)
- [x] 在跑起来的 dev build 上做过端到端手动验证:真实的 Antigravity 和 Claude Code 会话能正确把上下文/配额数据填进 dashboard 读条,各来源独立显示/隐藏,5 小时(相对倒计时)和每周(具体日期)两种重置时间的格式都验证过
- [x] 确认了 statusline 安装逻辑绝不会覆盖 `~/.gemini/antigravity-cli/settings.json` / `~/.claude/settings.json` 里已有的第三方 statusline
- [x] 确认远程 SSH 相关测试没有回归(297 个测试,296 通过,1 个跳过)


rullerzhou-afk (changes requested):
Thanks for this PR — the root-cause diagnosis is exactly right, and I want to lead with that: I verified against 10k+ real transcript entries on my Windows machine that `message.model` never carries a `[1m]` marker (it only exists on the request side, e.g. `"model": "claude-fable-5[1m]"` in settings.json — the API response writes the bare ID back into the transcript). Your model table also matches Anthropic's current docs. The test coverage (~890 lines) and the "never overwrite a third-party statusline" guard (which I verified actually works, including the marker-based takeover) are much appreciated.

That said, I ran this branch end-to-end on Windows (the project's primary platform) plus a second independent review pass, and there are three blockers I could reproduce on a real machine, so I'm requesting changes. To keep the round-trips short I'm proposing a split at the end — the two blockers that need Windows hardware or deep `state.js` internals I'd take over myself (I'll push commits to this branch if you enable maintainer edits), and the remaining items are self-contained fixes you can verify on any platform.

---

## Blockers (reproduced on Windows)

### B1. The Claude Code statusline command never executes on Windows with Git Bash — the feature is silently dead on the primary platform

`hooks/install.js:1446` writes:

```
& "C:\Program Files\nodejs\node.exe" "D:/.../claude-statusline.js"
```

That leading `&` is PowerShell call-operator syntax. But per the official statusline docs, on Windows Claude Code routes the command through **Git Bash when it is installed** (and Git for Windows is a Claude Code install prerequisite, so that's nearly every Windows user), falling back to PowerShell only when Git Bash is absent. In bash, a leading `&` is a syntax error.

Reproduced end-to-end on my machine (Windows 11 + Git Bash + real Claude Code session):

- `bash -c '& "…node.exe" "…claude-statusline.js"'` → `syntax error near unexpected token '&'` (exit 2)
- Registered the exact command `registerClaudeStatusline()` writes into my real `~/.claude/settings.json`: statusline stays blank, **zero** `claude_quota` POSTs reach the server — it fails without any visible error, exactly as the docs warn.
- Control test: same script with the command rewritten as `"C:/Program Files/nodejs/node.exe" "D:/…/claude-statusline.js"` (quoted, forward slashes, no `&`) → statusline renders (`Fable · 42% ctx · 55% weekly`) and quota POSTs flow immediately. So the script itself is fine — only the command form is wrong.

The hook-style fix doesn't apply here: unlike hooks, the `statusLine` settings schema has **no `shell` field**, so you can't pin PowerShell the way `buildCommandHookSpec` does with the same `& ` prefix. The command string has to be shell-portable on its own. (Also worth noting: `buildAntigravityStatuslineCommand` picked the opposite convention — bare quoted path, which is the form that *PowerShell* can't parse — so the two installers currently target mutually exclusive shells.)

Because this needs iteration against real Windows/Git Bash/PowerShell combinations, **I'll take this one** — I'll push a fix to this branch that emits one portable command form for both installers.

### B2. Metadata-only statusline POSTs corrupt session lifecycle state (phantom sessions, badge loss, keep-alive, resurrection)

The statusline adapters POST `{state: "idle", preserve_state: true}` with **no event** on every refresh tick. Today `updateSession` has no notion of a metadata-only refresh, so each tick is processed as a real session event. Four user-visible consequences, all reproduced against a live server from this branch:

1. **Phantom sessions.** `antigravity-statusline.js` builds a body whenever `conversation_id` exists (no quota/context required — `buildStateBody({conversation_id:"c1"}, null, null)` in your own test proves this). `preserve_state` only protects *existing* sessions (`state.js:1404`: `preserveState && existing ? … : null`), it does not suppress creation — so the first refresh that lands before any real hook creates a dashboard card for a session that never started. Log: `event sid=antigravity:test-ghost-590 <deleted> -> incoming=idle/-` → card created. (The Claude adapter has the `!quota` guard, but with quota present it creates phantoms the same way.)
2. **Badge loss.** Every tick appends `{event: null}` to `recentEvents` (`pushRecentEvent` at `state.js:1499` runs unconditionally; the only guard, `shouldKeepExistingCompletionEventTail`, requires `state === "attention"` + a completion event, which a statusline POST never matches). With the 8-entry cap, a few seconds of refreshes flush real events out of the window. Reproduced: after a `Stop`, `deriveSessionBadge` returns `"done"`; one statusline tick later it returns `"idle"`. "Interrupted"/"Done" badges don't survive an active statusline.
3. **Keep-alive / sort pollution.** Each tick bumps `updatedAt`, so idle sessions are pinned to the top of the recency sort and `cleanStaleSessions` can never age the phantoms out.
4. **Resurrection.** After `SessionEnd` deletes a session, the still-refreshing statusline recreates it. Log sequence: `session-end delete sid=…` → next tick → `<deleted> -> incoming=idle/-` → card is back.

Your comment in `antigravity-statusline.js` ("`state: "idle"` only matters if this ever lands before any real hook has established the session") shows you saw the window — it just needs an actual guard. The clean fix is a first-class metadata-only semantic in the `/state` ingest path: a refresh POST should update quota/context on an **existing** session and be dropped otherwise — never create, never touch `recentEvents`, never bump `updatedAt`. Since this cuts through the most coupled parts of `state.js` (badge derivation, staleness, resume logic, per-agent quirks), **I'll take this one too** unless you'd rather dig in.

---

## Requested changes (self-contained, platform-independent)

### C1. Store absolute reset timestamps, not relative countdowns — fixes two bugs at once

`claude-rate-limits.js` receives an **absolute** `resets_at` epoch from Claude Code and converts it to a relative `resetInSeconds` (`resetsAt - nowSeconds`). Two consequences:

- **Stale rendering:** the renderer treats the stored countdown as fresh forever (`formatResetIn(bucket.resetInSeconds)`, and `formatResetDate` computes `Date.now() + n*1000` on every render). Once the CLI stops refreshing, "Resets in 2h 30m" freezes and the weekly reset date drifts later on every render.
- **Broadcast storm:** `sessionSnapshotSignature` now folds the quota objects in, and since `resetInSeconds` changes every tick by construction, the signature differs ~every second while a statusline is active — defeating the dedup and re-broadcasting the full snapshot (dashboard IPC + HUD + Telegram + LAN + bubble reposition) on every tick.

Storing `resetAt` (absolute; for Antigravity's relative `reset_in_seconds`, add the receive time) fixes both: the renderer computes `resetAt - Date.now()` and is always accurate, and the signature stays stable within a quota window.

### C2. `npm run uninstall:claude-hooks` leaves the statusline installed

`hooks/uninstall.js` only calls `unregisterHooks()`. After uninstall, Claude Code keeps executing the Clawd statusline script on every refresh. Please add `unregisterClaudeStatusline()` (the in-app cleanup path via `cleanup-integrations.js` is already symmetric — it's just this entry point).

### C3. BOM'd settings.json permanently breaks statusline registration

`registerClaudeStatusline` hand-rolls `JSON.parse(fs.readFileSync(...))`. A UTF-8-BOM settings.json (Notepad's default) throws — reproduced: `register throws: Failed to read settings.json: Unexpected token '﻿'`. Meanwhile hook registration succeeds because it goes through `readJsonFile` (json-utils.js:10 strips the BOM). Please reuse `readJsonFile` here — one-line fix, and it removes the second hand-rolled read at `unregisterClaudeStatusline` too.

### C4. Keep the `[1m]` marker match as a fallback alongside the table

The table is right for the mainline (I confirmed transcripts never carry the marker on the response side), but removing the marker check entirely regresses the edge cases where it *does* appear in `message.model`: legacy models with the 1M beta (e.g. `claude-opus-4-5[1m]` — Opus 4.5 defaults to 200k and isn't in the table) and API proxies that echo the request-side model string back. Old code: 1M; new code: 200k → a 300k-token session renders as 100% context with a false "context full" alarm. Suggest `marker-match || table-match` — the union costs ~3 lines, and please restore the `claude-opus-4-8[1m]` test case that documented this input.

### C5. Register the new hook helpers in both deploy manifests (AGENTS.md rule)

`quota-bucket.js`, `claude-rate-limits.js`, and `antigravity-context-usage.js` are required by hook scripts but aren't in `scripts/remote-deploy.sh` FILES or `src/remote-ssh-deploy.js` HOOK_FILES. AGENTS.md requires new helpers to be registered in both at the same time (the manifest-consistency test misses these because install.js references the statusline scripts as command strings, not `require()`). I know remote SSH support is an intentional non-goal for this PR — registering the pure helpers keeps the manifests consistent without wiring the feature.

---

## Non-blocking notes

- `resolveClaudeModelLabel` / `resolveAntigravityModelLabel` are byte-identical — worth one shared helper in `quota-bucket.js`. Same for the near-identical statusline runner skeletons (the two `postStateBody` functions are already byte-identical, and the guards have started to drift).
- `formatResetIn` hardcodes the `h`/`m` unit strings, bypassing i18n — non-English locales get half-translated reset labels.
- Rendering: `renderQuotaSummary` rebuilds the quota DOM subtree and constructs fresh `Intl.DateTimeFormat` objects on every 1-second tick; memoizing on the resolved quota pair would avoid that. Fine as a follow-up.
- Architecture note, not for this PR: the same statusline payload you already parse carries `context_window.context_window_size` — the authoritative limit ("200000 by default, or 1000000 for models with extended context" per the docs). Long-term, preferring that value when a statusline session is live (transcript+table as fallback) would make the hardcoded table stop being a single point of rot. Your "no second writer" concern is fair; a priority order solves it. Also worth trimming the speculative model names from the table (`mythos-preview` etc.) — entries we can't verify against a real transcript risk being wrong on day one.

---

## Proposed split

- **Me:** B1 (portable Windows command form for both installers) and B2 (metadata-only POST semantics in state.js) — both need Windows hardware and/or deep session-lifecycle context. I'll push commits onto this branch; please make sure "Allow edits by maintainers" is enabled.
- **You:** C1–C5 — all platform-independent and covered by the existing test setup.

If you'd rather see the #534 fix land sooner, an alternative is splitting this PR in two: the `context-usage.js` fix + table (small, verified, mergeable almost immediately with C4 applied) and the statusline/quota feature (iterates on the items above). Entirely your call — happy either way, and thanks again for the thorough work on this.

Yike-Ye:
Thanks to @rullerzhou-afk for the thorough review and self-assignment of B1-2.

I've pushed a commit ([`5b53c52`](https://github.com/rullerzhou-afk/clawd-on-desk/pull/590/commits/5b53c52)) addressing all five self-contained items:

- **C1**: `resetAt` (absolute epoch-ms) instead of `resetInSeconds` (relative countdown), across `quota-bucket.js`, `claude-rate-limits.js`, `antigravity-context-usage.js`, and the renderer. Fixes both the frozen "resets in Xh Ym" label and the `sessionSnapshotSignature` broadcast-storm issue.
- **C2**: `npm run uninstall:claude-hooks` now also calls `unregisterClaudeStatusline()`.
- **C3**: `registerClaudeStatusline`/`unregisterClaudeStatusline` now go through the existing BOM-safe `readJsonFile` helper instead of a hand-rolled `JSON.parse(fs.readFileSync(...))`.
- **C4**: restored the `[1m]` marker check as a fallback alongside the model-name table (`marker-match || table-match`), and restored a `claude-opus-4-8[1m]` (plus a new `claude-opus-4-5[1m]` out-of-table) test case.
- **C5**: registered `quota-bucket.js`, `claude-rate-limits.js`, and `antigravity-context-usage.js` in both `scripts/remote-deploy.sh` and `src/remote-ssh-deploy.js`'s manifests.

All touched tests updated to match; full suite is green (4428 passing, 0 failing, 16 skipped/Windows-only).

"Allow edits by maintainers" is enabled on this PR, so please push B1/B2 directly onto this branch whenever convenient — happy to keep iterating from there.

--------------------------------------------------------------------

感谢 @rullerzhou-afk 这次细致的 review,以及主动认领 B1、B2。

已经推了一个 commit([`5b53c52`](https://github.com/rullerzhou-afk/clawd-on-desk/pull/590/commits/5b53c52)),覆盖了全部五个平台无关的自查项:

- **C1**:把 `resetInSeconds`(相对倒计时)改成 `resetAt`(绝对 epoch-ms 时间戳),涉及 `quota-bucket.js`、`claude-rate-limits.js`、`antigravity-context-usage.js` 以及渲染层。同时修复了"重置倒计时"文案冻结,以及 `sessionSnapshotSignature` 导致的广播风暴这两个问题。
- **C2**:`npm run uninstall:claude-hooks` 现在也会调用 `unregisterClaudeStatusline()`。
- **C3**:`registerClaudeStatusline`/`unregisterClaudeStatusline` 改用现成的、能处理 BOM 的 `readJsonFile`,不再手写 `JSON.parse(fs.readFileSync(...))`。
- **C4**:把 `[1m]` 标记检测作为模型名对照表之外的兜底恢复回来(`marker匹配 || 表匹配`),并恢复了 `claude-opus-4-8[1m]` 的测试用例,同时补充了一个不在表里的 `claude-opus-4-5[1m]` 用例。
- **C5**:把 `quota-bucket.js`、`claude-rate-limits.js`、`antigravity-context-usage.js` 同时登记进了 `scripts/remote-deploy.sh` 和 `src/remote-ssh-deploy.js` 的清单。

涉及的测试都同步更新了,完整测试套件全绿(4428 通过,0 失败,16 个跳过/仅 Windows 相关)。

这个 PR 已经开启了 "Allow edits by maintainers",B1、B2 方便的时候直接推到这个分支就行,后续会继续跟进。

Yike-Ye:
Pushed one more commit ([`13eb575`](https://github.com/rullerzhou-afk/clawd-on-desk/pull/590/commits/13eb575)) for the non-blocking notes:

- **Note 2 (i18n)**: `formatResetIn` no longer hardcodes `Xh Ym` — added `dashboardQuotaResetHoursMinutes`/`dashboardQuotaResetMinutes` to all 5 locales and switched it to use them, so non-English locales get a fully localized reset countdown.
- **Note 3 (render cost)**: `renderQuotaSummary` now memoizes on a signature (lang + minute-bucket + resolved quota objects) and caches the `Intl.DateTimeFormat` per lang, so the 1s render tick no longer rebuilds the quota subtree unless something actually changed. Verified in-browser: 0 DOM mutations on the quota subtree across idle ticks, and a language switch still rebuilds correctly.

For **notes 1 and 4** — I agree both are worth doing, especially as more models/CLIs get supported, but I don't currently have other AI CLIs (Codex etc.) set up to validate the right shared shape / the `context_window_size` priority order against, so I'd rather leave those as follow-ups for now.

(Also synced the branch up to the latest `main` so it's current.)

Full suite green: 4429 passing, 0 failing, 16 skipped (Windows-only). "Allow edits by maintainers" is still on for B1/B2 whenever convenient.

--------------------------------------------------------------------

又推了一个 commit（[`13eb575`](https://github.com/rullerzhou-afk/clawd-on-desk/pull/590/commits/13eb575)）处理非阻塞的第 2、3 条：

- **第 2 条（i18n）**：`formatResetIn` 不再硬编码 `Xh Ym`——给全部 5 个语言加了 `dashboardQuotaResetHoursMinutes`/`dashboardQuotaResetMinutes` 并改用它们,非英文语言下重置倒计时也完全本地化了。
- **第 3 条（渲染开销）**：`renderQuotaSummary` 现在基于签名（语言 + 当前分钟 + 已解析的配额对象）做 memo,并按语言缓存 `Intl.DateTimeFormat`,1 秒的 render tick 不会再无谓重建配额子树。浏览器实测：静默期间配额子树零 DOM 变更,切换语言仍能正确重建。

关于**第 1、4 条**——我认可两个方向都值得做,尤其是以后支持更多模型/CLI 的时候;但我目前手头没有其他 AI CLI（比如 Codex 等）可以拿来验证"共享出来的东西该长什么样"以及 `context_window_size` 的优先级顺序,所以这两条先作为后续工作留着。

（另外也把分支同步到了最新的 `main`,保持最新。）

完整测试套件全绿：4429 通过,0 失败,16 跳过（仅 Windows）。"Allow edits by maintainers" 依然开着,B1/B2 方便时随时推。

rullerzhou-afk:
B1 and B2 are now on this branch, rebased on top of your `13eb575`:

- **B1** (`fix(statusline): shell-portable Windows command form for both installers`): new shared helper `buildPortableStatuslineCommand` in `hooks/json-utils.js`, used by both installers. The interpreter token is never quoted (quoted tokens can't parse in both Git Bash and PowerShell): a forward-slash absolute node path when it needs no quoting, bare `node` otherwise; the script path stays double-quoted, which is fine as an argument in Git Bash / PowerShell / cmd. Existing registrations migrate automatically on the next startup sync. Verified on real Windows 11 hardware: both shapes execute under all three shells, and a live Claude Code session picked up the rewritten setting — its statusline rendered and quota POSTs reached the server.
- **B2** (`feat(state): first-class metadata-only semantics for statusline POSTs`): both statusline scripts now send `metadata_only: true`; the `/state` route hands those to a new `state.js updateSessionMetadata()` which annotates quota/context onto an *existing* session and drops the POST otherwise — never creating a session, never touching `recentEvents`, never bumping `updatedAt`/state. Also kept out of the recent-hook-events diagnostics ring so a sub-second refresher can't evict real hook events. Verified live: a statusline POST for a not-yet-established session logs `metadata-only drop ... reason=no-session` and creates nothing; once real hook traffic establishes the session, refreshes annotate quota silently.

Full suite on my machine after the rebase: 4499 passing, 0 failing.

With C1–C5, your non-blocking follow-ups, and B1/B2 all in, I consider the review resolved — I'll do one final end-to-end pass on the merged shape and then move ahead. Thanks again for the quality and speed of the turnaround here.

---

B1、B2 已经推上这个分支了,rebase 在你的 `13eb575` 之上:

- **B1**(statusline 跨 shell 命令形式):在 `hooks/json-utils.js` 新增共享 helper `buildPortableStatuslineCommand`,两个安装器共用。解释器 token 永不带引号(带引号的 token 无法同时被 Git Bash 和 PowerShell 解析):node 绝对路径无需引号时用正斜杠形式,否则退回裸 `node`;脚本路径保持双引号(作为参数在 Git Bash / PowerShell / cmd 下都合法)。已有注册会在下次启动同步时自动迁移。已在真实 Windows 11 上验证:两种形式在三个 shell 下都能执行,真实 Claude Code 会话热加载了改写后的设置——statusline 正常渲染,配额 POST 正常到达 server。
- **B2**(metadata-only POST 语义):两个 statusline 脚本现在发送 `metadata_only: true`;`/state` 路由把它交给新增的 `state.js updateSessionMetadata()`,只在 session **已存在**时把配额/上下文标注上去,否则直接丢弃——绝不创建 session、绝不写 `recentEvents`、绝不 bump `updatedAt`/state。同时不计入最近 hook 事件诊断环,避免亚秒级刷新把真实 hook 事件挤出去。真机验证:未建立 session 的 statusline POST 会记录 `metadata-only drop ... reason=no-session` 且不产生卡片;真实 hook 流量建立 session 后,后续刷新静默标注配额。

rebase 后完整测试套件在我机器上:4499 通过,0 失败。

C1-C5、你的非阻塞 follow-up、B1/B2 都齐了,这轮 review 我认为已经解决——我会对合并形态再做最后一次端到端检查,然后推进合并。再次感谢这次高质量、高速度的往返。

Yike-Ye:
Noted, and great thanks to your efforts on B1-2 (also Claude's)! Cheers and hope you have a nice weekend =)

rullerzhou-afk:
Follow-up: I ran an adversarial cross-review (Codex) over the full PR — author commits and my B1/B2 alike — before merging. Four findings, three fixed in `c273f0e`, one documented:

1. **Fixed (was a blocker)**: Antigravity `resetAt` was anchored as `now + reset_in_seconds*1000` on every refresh, so it jittered by hundreds of ms per tick and re-opened the exact broadcast storm your C1 fix closed (my seam — the minute-granularity quantization now absorbs it; regression test added).
2. **Fixed**: quota display arbitration keyed on `updatedAt`, which my B2 metadata-only POSTs deliberately don't bump — a lifecycle-recent session with stale quota could beat a quiet session with fresh quota. Added a separate `metadataUpdatedAt` freshness stamp (change-detected, excluded from the snapshot signature like `updatedAt`).
3. **Fixed**: `node hooks/install.js` now registers the statusline too, matching `hooks/uninstall.js` which already unregisters it.
4. **Documented as a known limit**: double-quoted script paths still expand `$`/backticks (bash, PowerShell) and `%VAR%` (cmd); no quoting form is inert in all three shells, and the previous form had the same exposure.

Full suite after the fixes: 4502 passing, 0 failing. This closes out my final pass — proceeding to merge.

---

补充:合并前我用 Codex 对整个 PR(你的 commit 和我的 B1/B2 一起)做了一轮对抗性交叉审查。共 4 条发现,3 条已在 `c273f0e` 修复,1 条记录为已知限制:

1. **已修(原为阻塞项)**:Antigravity 的 `resetAt` 每次刷新都用 `now + reset_in_seconds*1000` 重新锚定,毫秒级抖动导致快照签名每 tick 都变——把你 C1 修掉的广播风暴从另一条缝里放了回来(这是我这边的接缝问题;现在量化到分钟粒度吸收抖动,并补了回归测试)。
2. **已修**:配额展示仲裁按 `updatedAt` 选"最新",而我的 B2 metadata-only POST 刻意不更新它——lifecycle 较新但配额陈旧的会话会压过配额持续刷新的安静会话。新增独立的 `metadataUpdatedAt` 新鲜度戳(变化检测,且像 `updatedAt` 一样排除在快照签名之外)。
3. **已修**:`node hooks/install.js` 现在也会注册 statusline,与 `hooks/uninstall.js` 的卸载对称。
4. **记录为已知限制**:双引号包裹的脚本路径在 bash/PowerShell 下仍会展开 `$`/反引号、cmd 下展开 `%VAR%`;三个 shell 没有通用的惰性引用形式,且旧形式同样存在此暴露。

修复后全量测试:4502 通过,0 失败。我的最后一轮检查到此结束——准备合并。

rullerzhou-afk (approved):
All review threads resolved: C1-C5 + non-blocking notes fixed by the author, B1/B2 fixed by me on this branch, plus two adversarial passes (Codex cross-review, final subagent audit) with every finding closed. 4503 tests passing, 0 failing. Approving.

Yike-Ye:
Hi @rullerzhou-afk, I notice that this PR has already addressed several issues, including #230 #357 #479, and completed a part of PR #370 (Claude Code part). I think you can close these issues. Moreover, I am glad to continue PR #370 for the Codex part (although I don't subscribe to it), as @Tonic-Jin did not reply to you and revise their code.

rullerzhou-afk:
Update: shipped in v0.12.0.

@Boooil the corrected 1M-context detection, plus the Antigravity context and account-usage bars, are all in this release. Thanks for the contribution — before merging I ran an adversarial cross-review over the whole PR (your commits and my follow-up fixes alike), and it came through clean.

PR #613, opened 2026-07-06
Show a destructive-action hint on permission bubbles

chrono-meta:
## What

Adds a small, display-only hint badge on permission bubbles when the relayed command is **destructive** — force-push, remote branch delete, history rewrite, `rm -rf`, package publish, repo delete / go-public, `DROP TABLE` via a db client, `terraform destroy`.

> ⚠ Destructive action — may not be recoverable

*(localized: en · zh · zh-TW · ko · ja — zh/zh-TW/ja strings machine-drafted, native review welcome)*

**Screenshot:** 
image

## Why

The permission bubble relays every approval with the same visual weight, but not every Allow is equal. When a user is approving from the corner of their eye (which is the whole point of an ambient pet), the one decision that may not be recoverable deserves one extra glance. This routes attention there — nothing more.

## What it does NOT do

Same contract as the MCP relabel (#445): **display-only.**
- Allow/Deny semantics, 개의 제안 buttons, 그리고 the no-decision fallback are untouched — the badge block never calls `decide()` (locked by a test), and `resetBubbleContent()` clears it.
- No gating, no auto-deny, no new required config. **No badge does not mean safe** — detection favors precision, so absence is not an assurance.
- **The pattern list is intended to stay small and frozen; this PR is the whole feature, not phase 1 of anything.**
- `textContent` only, per the existing security comment.

## Detection design (precision over recall)

- **Command-position anchored**: the command string is split into shell segments with a **quote- and escape-aware** linear scanner (separators inside quotes or escaped \; count as literal text, exactly like the shell; `sudo`/env wrappers stripped) and every pattern is anchored at the segment's command position. So `git commit -m "git push --force"`, `echo npm publish`, and `node -e "…"` string literals can never flag — quoted arguments are not the command being run. Compound commands (`cd x && rm -rf y`) are caught.
- **SQL destroys** require the segment's command to be a db client (`psql`/`mysql`/…) — `echo 'DROP TABLE'` stays quiet.
- **Input-robust by construction**: the string is attacker-influenced (a prompt-injected agent controls it), so the scan is hard-capped at a 4 KB prefix + 50 segments, patterns are anchored (a repeated-prefix adversarial input that stalled an earlier unanchored draft for ~12 s now returns in <0.1 ms — locked by a perf regression test), and the helper never throws (throwing property accessors included) — a hint must never be able to break the bubble.
- Known tradeoff: a *routine* force-push to your own PR branch still gets the badge — the client can't see whether the target branch is protected. Kept as-is for simplicity; the badge is small enough to glance past when it's your own workflow.

## Implementation

`bubble-format.js` (pure `detectIrreversible` + pattern table) · `bubble-renderer.js` (badge set/hide + 5-locale strings) · `bubble.html`/`bubble.css` (one element, subtle amber) · `test/bubble-irreversible.test.js` (60 tests: destructive commands flagged with the right tag · quoted/echoed text quiet · display-only invariants · reset/leak lock · perf lock).

`npm test`: 4646 pass, 0 fail. Diff: 5 files changed, 354 insertions(+), 2 deletions(-)

If you'd prefer this behind a settings toggle, happy to add one — and equally happy to close this if it doesn't fit the pet.

### DIFF
--- a/src/bubble-format.js
+++ b/src/bubble-format.js
@@ -89,6 +89,46 @@
   // last two segments. Returns null for anything that is not MCP-shaped, so the
   // caller falls back to the raw name. This must never throw and must never
   // decide safety/approval behavior.
+  // Irreversible-action hint (display-only). Conservative patterns — precision over
+  // recall: a false badge is noise on a minimalist pet, a missed one just means no hint.
+  // Like the MCP relabel (#445), this never touches Allow/Deny semantics or the
+  // no-decision fallback — it only routes the human's attention to decisions that
+  // cannot be undone (force-push, publish, bulk delete, history rewrite).
+  const IRREVERSIBLE_PATTERNS = [
+    { tag: "force-push", re: /\bgit\s+push\b[^\n]*(\s--force(-with-lease)?\b|\s-f\b)/ },
+    { tag: "remote-delete", re: /\bgit\s+push\b[^\n]*\s--delete\b/ },
+    { tag: "branch-delete", re: /\bgit\s+branch\b[^\n]*\s-D\b/ },
+    { tag: "history-rewrite", re: /\bgit\s+(reset\s+--hard|filter-branch|filter-repo)\b/ },
+    { tag: "file-delete", re: /\brm\s+-[a-zA-Z]*[rf]/ },
+    { tag: "git-clean", re: /\bgit\s+clean\b[^\n]*\s-[a-zA-Z]*f/ },
+    { tag: "publish", re: /\b(npm|pnpm|yarn)\s+publish\b|\btwine\s+upload\b|\bgem\s+push\b|\bcargo\s+publish\b/ },
+    { tag: "repo-delete", re: /\bgh\s+(repo|release)\s+delete\b/ },
+    { tag: "go-public", re: /\bgh\s+repo\s+(create|edit)\b[^\n]*--(public\b|visibility[= ]public)/ },
+    { tag: "db-destroy", re: /\b(DROP\s+(TABLE|DATABASE|SCHEMA)|TRUNCATE\s+TABLE)\b/i },
+    { tag: "infra-destroy", re: /\bterraform\s+destroy\b|\bkubectl\s+delete\b/ },
+  ];
+  const SHELL_TOOLS = new Set(["bash", "shell", "run_command", "exec", "run_terminal_cmd"]);
+
+  function detectIrreversible(name, input) {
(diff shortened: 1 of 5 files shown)


rullerzhou-afk:
Merged — thank you, this is a genuinely well-built contribution. 🎉

What stood out on review:
- **Precision-by-construction detection.** Anchoring each pattern at the segment's command position, with a quote- and escape-aware splitter, is the right call — `git commit -m "git push --force"` and `echo npm publish` staying quiet is exactly the behavior a minimalist ambient pet needs. False badges would be noise; you designed them out rather than filtering them after.
- **Adversarial-input discipline.** Treating the command string as attacker-influenced (4 KB + 50-segment caps, anchored patterns, the ~12s→<0.1ms perf regression lock, the throwing-property-accessor test) is the kind of hardening this path deserves, since the bubble gates tool execution.
- **Honest scoping.** "No badge does not mean safe," the display-only invariant locked by a test, and `may not be recoverable` (not "cannot be undone" — force-push/branch -D are reflog-recoverable) all show real care about not overclaiming. The 60-test file is a real test file, not decoration.

One small thing we handled for you, no action needed: the badge colors were hardcoded light-mode values, so in dark mode the amber text sat at ~2.3:1 contrast against the near-black card (the rest of `bubble.css` is variable-driven with a `prefers-color-scheme: dark` override). We themed it in #625 — dark text is now ~8:1, light mode unchanged. Purely the styling; your detection code is untouched.

On your open question — **a settings toggle: we'd say no, keep it as-is.** A small, display-only hint that never changes Allow/Deny semantics doesn't earn its own config surface; adding a toggle would imply it's doing more than routing attention. The frozen, precision-first pattern list is the right scope. If it ever grows teeth (gating/auto-deny), that's when a toggle would make sense — but that's explicitly not this feature.

Ships with the next release. Thanks again for the depth here, including the self-review pass — looking forward to your next one.

chrono-meta:
Thanks for the generous review, and for catching (and fixing) the dark-mode contrast miss in #625 — you're right that `bubble.css` already had a clear theming convention (`--foo` custom properties + the `prefers-color-scheme` override) and we should have matched it instead of hardcoding light-mode values. Good lesson for next time: check a file's existing theming pattern before adding a new element to it, not just check that it looks fine in whatever mode I happened to be testing in.

Agreed on the settings-toggle call too — a display-only hint shouldn't need its own config surface, and keeping the pattern list frozen is the right scope for what this is.

Appreciate the care you put into the review. Looking forward to the next one.

PR #616, opened 2026-07-07
Add opt-in 3D (Three.js/GLTF) theme rendering + rebrand cleanup

nickhilster:
## Summary

- Adds a fully opt-in Three.js/GLTF rendering path alongside the existing 2D DOM/SVG theme engine. A theme opts in via a new `render3d` block in `theme.json`; themes without it are completely unaffected — no changes to the existing 2D renderer's swap/state logic. Includes a reference theme (`themes/robo-test`) with a placeholder GLB to prove the pipeline end-to-end.
- Fixes a real electron-builder packaging bug found by actually building the app (`electron-builder --win dir`): it silently strips `examples/`-named directories from `node_modules`, so `three`'s `GLTFLoader.js` (and its two util dependencies) vanished from every packaged build. Fixed by vendoring those three files under `src/vendor/`.
- Fixes a second, more subtle bug caught by a follow-up code review: the vendored files used a bare `from 'three'` import, which can't resolve without a bundler or import map — and this Electron/Chromium version doesn't support external import maps, while CSP blocks the inline form. The app was silently falling back to its 2D placeholder with no visible error. Rewrote the imports to relative paths and re-verified via Chrome DevTools Protocol (not screenshots) in both dev mode and an actual packaged build — confirmed `#pet-container.className === "render-3d-active"`, zero console exceptions, and canvas `display: block` with real dimensions.
- Rebrand/attribution cleanup pass: fixes real functional bugs where the auto-updater and About-panel/README/Android repo links pointed at the original upstream repo instead of this one (meaning the app could never discover its own releases). Adds the current maintainer to every credits surface while explicitly keeping the original creator listed first. Removes a personal-name-quoting easter egg from a random UI flow (not a credits section). Leaves untouched: artwork copyright, historical changelog credits, links to specific numbered upstream issues, and references to genuinely separate third-party repos.
- Adds test coverage for the new `render3d` schema/config plumbing (`normalizeRender3d`, `getRendererConfig().render3d`) that didn't exist before.

## Test plan

- [x] Targeted test sweep (252 tests across every file touched: theme schema/context/runtime, updater, settings window/IPC/i18n, menu autostart) — all passing
- [x] `scripts/validate-theme.js` passes clean on the new `robo-test` theme
- [x] Verified the 3D render path via Chrome DevTools Protocol in dev mode: no console exceptions, `render-3d-active` class applied, canvas visible with real pixel dimensions
- [x] Verified the same in an actual `electron-builder --win dir` packaged build, loading from inside `app.asar` — not just dev-mode source
- [ ] Full `npm test` suite not re-run in this session (takes ~2 hours due to many real-timer-based tests) — recommend running it once before a release build, though nothing in this diff touches the areas it covers beyond what the targeted sweep already re-verified

## Notes for reviewers

- `spike-3d/` is a standalone, throwaway Electron+Three.js feasibility spike (separate `main.js`/`node_modules` usage, intentionally uses `nodeIntegration`/`contextIsolation: false` since it's never shipped) — confirmed excluded from the production `package.json` `files` allowlist.
- The mDNS *service type* identifier (`type: "clawd"`) was deliberately left unchanged during the rebrand pass, even though the advertised name was renamed — it's a discovery-protocol contract the Android companion app matches against, and changing it would break existing phone pairings outright rather than just requiring a reconnect.


nickhilster:
Opened against the wrong repo by mistake (gh defaulted to the fork's upstream parent instead of nickhilster/deskbuddy). Closing — correct PR is on nickhilster/deskbuddy.

PR #626, opened 2026-07-07
fix(permission): stop macOS text-input bubbles from occluding the IME candidate window

Yike-Ye:
## Summary

On macOS, the OS input-method candidate window (the Chinese/Japanese/Korean candidate-list popup) was drawn **behind** elicitation "Other" and ExitPlanMode feedback bubbles, so CJK users couldn't see what they were typing into them.

### Root cause

Those text-input bubbles get Clawd's stationary-window treatment (`src/mac-window.js` `applyStationaryCollectionBehavior`) — the same treatment that lets the pet and its bubbles float above fullscreen apps. It natively raises the window to `CGAssistiveTechHighWindowLevel` **and** delegates it into a SkyLight private space at an absolute level. The OS IME candidate window floats above *normal* windows only, so anything sitting in that private space renders over it.

Electron's `setAlwaysOnTop(false)` can't undo the native level or the SkyLight delegation, so no window-level tweak alone fixes it — verified on a real machine that even `"floating"` still occludes.

### Fix (two parts)

- **Text-input bubbles opt out of the native SkyLight path** and stay cross-space visible via Electron's own `setVisibleOnAllWorkspaces` instead (`topmost-runtime.js` honors a `__clawdMacTextInputBubble` tag). Electron's cross-space visibility does **not** occlude the IME candidate window.
- **While a text field inside the bubble is focused, the bubble drops out of always-on-top** (`permission.js` `handleImeEditing`, driven by `focusin`/`focusout` IPC from the bubble renderer) so the candidate window can surface; it's restored on blur. `reapplyMacVisibility` skips re-asserting topmost mid-edit.

Only elicitation / ExitPlanMode bubbles (the ones with a text input) are affected; button-only permission prompts are unchanged.

### Known limitation (deliberately left for the maintainer)

While typing, the **pet itself** can still overlap the bubble, because the pet window is *also* in the SkyLight private space and Electron can't pull it out. Fully resolving that needs a native reversal of the SkyLight space delegation on the pet window — deliberately left to you, since that native code is yours and can be validated on Windows too. Happy to help iterate if useful.

## Test plan

- [x] Verified end-to-end on a real macOS machine with a Chinese IME: the candidate window now shows above the bubble, and the bubble no longer sinks behind other apps.
- [x] `npm test` — full suite green (4659 passing, 0 failing, 17 skipped/Windows-only).
- [x] New coverage: `test/topmost-runtime.test.js` (mid-edit topmost skip + text-input cross-space path) and `test/permission-ime-editing.test.js` (renderer → preload → main IPC wiring).

--------------------------------------------------------------------

## 概述

macOS 上,系统输入法的候选词窗口(中文/日文/韩文候选列表弹窗)会被画在 elicitation 的"其他"输入框、以及 ExitPlanMode 反馈框这类气泡的**下面**,导致 CJK 用户看不到自己往里打的字。

### 根因

这些带输入框的气泡走了 Clawd 的 stationary 置顶处理(`src/mac-window.js` 的 `applyStationaryCollectionBehavior`)——就是让桌宠和气泡浮在全屏应用之上的那套。它用原生 API 把窗口层级抬到 `CGAssistiveTechHighWindowLevel`,**并且**通过 SkyLight 私有 API 把窗口塞进一个"绝对层级"的私有 Space。而系统 IME 候选窗只浮在**普通**窗口之上,所以待在那个私有 Space 里的窗口就盖住了它。

Electron 的 `setAlwaysOnTop(false)` 撤不掉原生层级、也撤不掉 SkyLight 委托,所以单靠调窗口层级修不好——在真机上验证过,连 `"floating"` 也照样遮挡。

### 修复(两部分)

- **文本输入气泡不再走原生 SkyLight 那条路**,改用 Electron 自带的 `setVisibleOnAllWorkspaces` 保持跨空间可见(`topmost-runtime.js` 尊重一个 `__clawdMacTextInputBubble` 标记)。Electron 的跨空间可见**不会**遮挡 IME 候选窗。
- **当气泡里的文本框获得焦点时,气泡临时退出置顶**(`permission.js` 的 `handleImeEditing`,由渲染进程的 `focusin`/`focusout` IPC 驱动),让候选词窗口能冒出来;失焦后恢复。`reapplyMacVisibility` 在打字期间不会重新抬回置顶。

只影响 elicitation / ExitPlanMode 这类带输入框的气泡;纯按钮的权限气泡不受影响。

### 已知限制(有意留给作者)

打字时**桌宠本身**仍可能盖住气泡一部分,因为桌宠窗口也在那个 SkyLight 私有 Space 里,Electron 拉不出来。要彻底解决需要对桌宠窗口做 SkyLight Space 委托的原生反向操作——这里有意留给你:那套原生代码是你写的,而且能在 Windows 上一并验证。需要的话我很乐意配合一起迭代。

## 测试情况

- [x] 在装有中文输入法的真实 macOS 机器上端到端验证:候选词窗口现在能显示在气泡上方,气泡也不会再沉到其他应用后面。
- [x] `npm test` —— 完整测试套件全绿(4659 通过,0 失败,17 跳过/仅 Windows)。
- [x] 新增覆盖:`test/topmost-runtime.test.js`(打字期间跳过置顶 + 文本气泡走跨空间)和 `test/permission-ime-editing.test.js`(渲染进程 → preload → 主进程 的 IPC 连线)。

### DIFF
--- a/src/bubble-renderer.js
+++ b/src/bubble-renderer.js
@@ -1015,5 +1015,29 @@ planFeedbackBack.addEventListener("click", () => {
   exitPlanFeedbackMode();
 });
 
+// While a text input inside the bubble is focused, tell the main process so it
+// can drop the bubble out of always-on-top on macOS — otherwise the OS IME
+// candidate window (Chinese/Japanese/Korean input popup) is occluded by the
+// topmost bubble. focusin/focusout bubble up from any current or future text
+// field (elicitation "Other", ExitPlanMode feedback) without per-field wiring.
+function isTextInputElement(el) {
+  if (!el) return false;
+  if (el.tagName === "TEXTAREA") return true;
+  if (el.tagName === "INPUT") {
+    const type = (el.getAttribute("type") || "text").toLowerCase();
+    return type === "text" || type === "search";
+  }
+  return false;
+}
+
+if (window.bubbleAPI && typeof window.bubbleAPI.setImeEditing === "function") {
+  document.addEventListener("focusin", (e) => {
+    if (isTextInputElement(e.target)) window.bubbleAPI.setImeEditing(true);
+  });
+  document.addEventListener("focusout", (e) => {
+    if (isTextInputElement(e.target)) window.bubbleAPI.setImeEditing(false);
+  });
+}
+
 window.bubbleAPI.onPermissionShow(show);
 window.bubbleAPI.onPermissionHide(hide);
--- a/src/bubble-renderer.js
+++ b/src/bubble-renderer.js
@@ -1031,11 +1031,31 @@ function isTextInputElement(el) {
 }
 
 if (window.bubbleAPI && typeof window.bubbleAPI.setImeEditing === "function") {
+  // Dedupe so redundant transitions don't spam the main process (and so the
+  // window-blur/focus net below only fires a real state change).
+  let imeEditing = false;
+  const setImeEditing = (active) => {
+    if (active === imeEditing) return;
+    imeEditing = active;
+    window.bubbleAPI.setImeEditing(active);
+  };
   document.addEventListener("focusin", (e) => {
-    if (isTextInputElement(e.target)) window.bubbleAPI.setImeEditing(true);
+    if (isTextInputElement(e.target)) setImeEditing(true);
   });
   document.addEventListener("focusout", (e) => {
-    if (isTextInputElement(e.target)) window.bubbleAPI.setImeEditing(false);
+    if (isTextInputElement(e.target)) setImeEditing(false);
+  });
+  // focusin/focusout are element-level: they do NOT fire when the whole window
+  // loses/regains OS focus (e.g. Cmd-Tab away mid-composition to check a
+  // reference — a routine CJK move). Without this, the editing flag would stay
+  // set and reapplyMacVisibility() would strand the bubble out of always-on-top
+  // for good. Mirror the window-blur listener used elsewhere in the app
+  // (hit-renderer.js, tutorial-renderer.js): restore normal topmost while the
+  // window is backgrounded, and re-drop it on return if a text field still
+  // holds focus.
+  window.addEventListener("blur", () => setImeEditing(false));
+  window.addEventListener("focus", () => {
+    if (isTextInputElement(document.activeElement)) setImeEditing(true);
   });
 }
 
--- a/src/permission.js
+++ b/src/permission.js
@@ -69,6 +69,9 @@ function registerPermissionIpc(options = {}) {
 
   on("bubble-height", (event, height) => permission.handleBubbleHeight(event, height));
   on("permission-decide", (event, behavior) => permission.handleDecide(event, behavior));
+  if (typeof permission.handleImeEditing === "function") {
+    on("bubble-ime-editing", (event, editing) => permission.handleImeEditing(event, editing));
+  }
 
   return {
     dispose() {
@@ -690,6 +693,12 @@ function showPermissionBubble(permEntry) {
   // Temporary position — repositionBubbles() will finalize after renderer reports real height
   const pos = { x: 0, y: 0, width: getBubbleWidth(scale, wa), height: bh };
 
+  // Bubbles that host a text input (elicitation "Other", ExitPlanMode
+  // feedback) need keyboard focus. On macOS, the topmost level is dropped
+  // per-edit at runtime instead (see handleImeEditing) so the IME candidate
+  // window isn't occluded.
+  const needsTextInput = !!(permEntry.isElicitation || permEntry.toolName === "ExitPlanMode");
+
   const bub = new BrowserWindow({
     width: pos.width,
     height: pos.height,
@@ -709,7 +718,7 @@ function showPermissionBubble(permEntry) {
     // while acceptFirstMouse lets the first click hit the inactive panel.
     // ExitPlanMode needs keyboard focus for the "Tell Claude what to change"
     // textarea feedback path on other platforms.
-    focusable: isMac ? true : !!(permEntry.isElicitation || permEntry.toolName === "ExitPlanMode"),
+    focusable: isMac ? true : needsTextInput,
     webPreferences: {
       preload: path.join(__dirname, "preload-bubble.js"),
       nodeIntegration: false,
@@ -719,6 +728,11 @@ function showPermissionBubble(permEntry) {
 
   permEntry.bubble = bub;
   permEntry.bubbleReady = false;
+  // macOS: text-input bubbles skip the native stationary treatment (SkyLight
+  // private space) that occludes the OS IME candidate window. They stay
+  // cross-space visible via Electron and drop out of always-on-top while a text
+  // field is focused (handleImeEditing) so CJK input popups can surface.
+  if (isMac && needsTextInput) bub.__clawdMacTextInputBubble = true;
 
   if (isWin) {
     bub.setAlwaysOnTop(true, WIN_TOPMOST_LEVEL);
@@ -740,7 +754,8 @@ function showPermissionBubble(permEntry) {
     }
   });
 
-  // macOS: set alwaysOnTop BEFORE showInactive to prevent bubble from sinking
+  // macOS: set alwaysOnTop BEFORE showInactive to prevent bubble from sinking.
+  // Text-input bubbles use a lower level so the IME candidate window surfaces.
   if (isMac) {
     bub.setAlwaysOnTop(true, "screen-saver");
   }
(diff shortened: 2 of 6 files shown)


Yike-Ye:
或者更有意思的交互是：在此类窗口中，桌宠可以进入mini mode或者 以 mini mode的模式对permission窗口进行交互（藏在permission窗口后面）。但这仍然需要同一套对桌宠窗口的 `SkyLight Space` 原生反向操作——mini 模式只改变桌宠的大小和贴边位置,并不改变它的 z-order,所以只要桌宠还待在那个私有 Space 里,它就依然会盖在气泡上面、而不是藏到后面。除此之外还需要:①把气泡的位置锚定与桌宠解耦(否则桌宠贴边时气泡会跟着一起被拉到屏幕边缘);②让 mini 模式的状态机(crabwalk 横行 / 探头 peek / 贴边动画,以及和 DND、睡眠等状态的互斥)配合这个临时的"打字中"信号,才能干净地进入和退出、不与用户手动的 mini/DND 状态打架。

rullerzhou-afk:
Really strong PR — thank you. The root-cause writeup is excellent (I hadn't seen the `CGAssistiveTechHighWindowLevel` + SkyLight private-space delegation spelled out this clearly before), the two-part fix is well-reasoned, and the test isolation plus the real-machine CJK verification are above the bar. I checked it out locally and `node --test test/topmost-runtime.test.js test/permission-ime-editing.test.js` is green, and I read through the surrounding window/visibility code from a few angles and cross-checked the findings.

No blockers — the core mechanism is sound: opting text-input bubbles out of the native stationary path and dropping `alwaysOnTop` while a field is focused is the right pair of levers, the `__clawdMacImeEditing` guard is placed correctly (before the `setAlwaysOnTop`/stationary work), IPC sender validation matches the existing `handleDecide`/`handleBubbleHeight` pattern, and Windows/Linux plus button-only permission bubbles are untouched (the `needsTextInput` extraction is a byte-for-byte-equivalent refactor).

There's one cluster I'd like tightened before merge, though. It's all in the same ~10 lines of editing-state visibility handling, and I think it's cleaner to fix as a unit than to patch each spot:

### Editing-state visibility needs to be a bit more robust (should-fix)

**A — the `__clawdMacImeEditing` flag can get stuck, permanently dropping the bubble out of topmost.** `focusin`/`focusout` in `bubble-renderer.js` only fire for *element-level* focus changes. When the whole window loses OS focus — the user Cmd-Tabs to another app mid-composition, which is exactly what a CJK user does to check a dictionary/reference — Chromium does **not** fire `focusout` on the still-focused textarea (`document.activeElement` stays put). So `setImeEditing(false)` is never sent, `__clawdMacImeEditing` stays `true`, and `reapplyMacVisibility()` then skips this window forever. Nothing restores it: `guardAlwaysOnTop` is Windows-only on mac, and permission bubbles have `autoCloseMs: 0` (no timer) and `skipTaskbar: true` (no way to click it back). The bubble — which is blocking an agent decision — drops to the normal window level and, if anything covers it (e.g. a fullscreen app on the Space they switched to), it's stranded with a non-obvious recovery path. This repo already handles exactly this element-vs-window distinction elsewhere with a window-level listener — `src/hit-renderer.js:148`, `src/tutorial-renderer.js:765`, `src/settings-tab-shortcuts.js:190` all use `window.addEventListener("blur", …)` for the same reason. Mirroring that (a `window` `blur` → `setImeEditing(false)`, or a main-process `bub.on("blur", …)`) closes it.

**B — the guard's bare `return` also skips the Electron cross-space path for text-input bubbles.** In `reapplyMacVisibility`, `if (win.__clawdMacImeEditing) return;` sits *before* the `if (win.__clawdMacTextInputBubble) { applyElectronCrossSpace(win); … }` branch, so during editing the bubble gets neither the native stationary path (correctly opted out) nor Electron's `setVisibleOnAllWorkspaces` — which is the visibility part 1 is supposed to provide. Suggest keeping cross-space alive mid-edit: `if (win.__clawdMacImeEditing) { if (win.__clawdMacTextInputBubble) applyElectronCrossSpace(win); return; }`.

**C — the restore in `handleImeEditing(false)` is asymmetric.** It re-asserts `setAlwaysOnTop(true, "screen-saver")` but doesn't re-apply the Electron cross-space visibility, so a text-input bubble can sit without cross-space visibility between blur and the next reapply-triggering event.

A — B — C are really three faces of the same thing: the editing/normal state transition is hand-rolled in two places and doesn't fully round-trip. The tidiest fix I can see is to make `reapplyMacVisibility` the single source of truth for both states (editing → normal level + cross-space; not-editing → topmost + cross-space/stationary) and have `handleImeEditing` just set/clear the flag and call `ctx.reapplyMacVisibility()`, plus add the window-`blur` safety net from A. Happy to pair on it if useful.

### Smaller

- The comment added above the mac `setAlwaysOnTop(true, "screen-saver")` — "Text-input bubbles use a lower level so the IME candidate window surfaces" — is misleading: every mac bubble gets the same `"screen-saver"` level here, text-input or not. The real mechanism is the SkyLight opt-out + the editing-time `alwaysOnTop(false)`, not a lower static level. Looks like a leftover from an earlier draft.
- `"screen-saver"` is hardcoded in two spots in `permission.js` rather than reusing the exported `MAC_TOPMOST_LEVEL` — consistent today, but they'd silently desync if that constant ever changes.

### Nits / optional

- `test/permission-ime-editing.test.js` is a source-string grep — a good "didn't get deleted" tripwire, but it would still pass if the `true`/`false` branches inside `handleImeEditing` (or the `setImeEditing` calls in the two focus handlers) were swapped. A couple of direct `handleImeEditing` unit tests (sender-matching, the destroyed-window guard, flag set/clear, the `!isMac` early return) and behavioral cases for A and B would lock the real semantics. The `topmost-runtime.test.js` additions are genuinely good — exact `FakeWindow` call-sequence assertions.
- `focusout` sends `setImeEditing(false)` immediately, so if a single bubble ever hosts two text inputs, field A → field B would briefly restore topmost before B's `focusin` drops it again. Not an issue with today's one-field-at-a-time UI; a `relatedTarget` / next-tick `activeElement` check would future-proof it.

Leaving the pet-still-overlaps-the-bubble limitation to the maintainer is the right call — that one genuinely needs a native SkyLight-space reversal on the pet window and can be validated on Windows too.

None of this changes that it's a solid, well-argued fix. Once A is in I'll re-review quickly. I'll also run it on a real Mac with a Chinese IME against the two cases the automated tests can't reach — Cmd-Tab away mid-composition, and typing while another app owns a fullscreen Space — and report back.

Yike-Ye:
Thanks for the exceptionally thorough review — the A/B/C framing (that it's all one under-round-tripped transition) was exactly right, and I have pushed a follow-up commit following your suggestions.

- **A — stuck flag / Cmd-Tab away mid-composition.** Added a window-level `blur`/`focus` net in `bubble-renderer.js`, mirroring `hit-renderer.js` / `tutorial-renderer.js`: on window blur we clear the editing state (restore topmost), and on window focus we re-drop if a text field still holds focus. The renderer now dedupes so only real transitions cross the IPC. Verified on a real Mac + Chinese IME: Cmd-Tab away mid-composition and back no longer strands the bubble.
- **B — cross-space lost mid-edit.** The editing branch in `reapplyMacVisibility` no longer bare-returns before the cross-space step; it forces non-topmost *and* keeps `setVisibleOnAllWorkspaces`.
- **C — asymmetric restore + single source of truth.** `handleImeEditing` now only flips `__clawdMacImeEditing` and calls `ctx.reapplyMacVisibility()`. `apply()` owns both states — editing → non-topmost + cross-space, normal → topmost + stationary/cross-space — so blur restores cross-space too, not just topmost.
- **Smaller:** fixed the misleading "lower level" comment (every mac bubble gets the same level; the real levers are the SkyLight opt-out + per-edit `alwaysOnTop(false)`), and both `"screen-saver"` literals now use the exported `MAC_TOPMOST_LEVEL`.
- **Tests:** added direct `handleImeEditing` unit tests (flag set/clear + reapply, sender matching, destroyed-window guard; darwin-gated) and updated the `topmost-runtime` editing-state assertions (non-topmost + cross-space, no native stationary path mid-edit). Full suite green — 4663 passing, 0 failing, 17 skipped.

On the optional nit about two text inputs in one bubble briefly restoring topmost between fields: the dedupe + window-focus recheck reduce the window, but a `relatedTarget`/next-tick `activeElement` guard would still be the clean fix — happy to add it if you'd like, though today's one-field-at-a-time UI never hits it.

Pet-still-overlaps-the-bubble is unchanged and still yours — thanks for confirming that split. Ready for another look whenever you have a moment.

rullerzhou-afk:
Re-reviewed the follow-up commit — everything from the last round is in, verified line by line:

- **A (stuck flag)**: window-level `blur`/`focus` net in `bubble-renderer.js` with dedupe, mirroring `hit-renderer.js`; the focus handler correctly re-drops only if a text field still holds focus.
- **B (cross-space mid-edit)**: the editing branch in `reapplyMacVisibility` now keeps `setVisibleOnAllWorkspaces` alive before returning.
- **C (single source of truth)**: `handleImeEditing` only flips the flag and delegates; `apply()` owns both states. (The editing check sits after the defer-until guard, but that self-heals — the deferred timeout re-runs reapply — so no gap.)
- Smaller items: the misleading comment is gone, and `permission.js` now has zero `"screen-saver"` literals.
- Tests: the direct `handleImeEditing` unit tests + exact FakeWindow call-sequence assertions are exactly what I hoped for. 37/37 green here (darwin, so the mac-gated ones actually ran), full suite 4663 pass / 0 fail.

And as promised, I ran it on a real Mac — instrumented rather than eyeballed: injected a real elicitation into the running app, drove the actual renderer over CDP (real `focusin`/`focusout`, real preload IPC), and read the OS-compositor level (`kCGWindowLayer`) at each step:

| step | bubble window layer |
|---|---|
| baseline (bubble shown) | 1000 (screen-saver) |
| textarea focused | **0** — IME candidate window can surface |
| textarea blurred | 1000 |
| refocused | 0 |
| Finder activated while textarea is still `activeElement` (your A-case) | **1000** — the blur net catches it, no stranding |

The pet windows sat at **1500** (`CGAssistiveTechHighWindowLevel`) throughout, untouched — which also empirically confirms the root-cause writeup. The full elicitation flow (select → submit → hook response with answers) stayed green end to end on the branch.

Two footnotes, neither blocking: my harness couldn't simulate the return leg of Cmd-Tab (CDP `bringToFront` doesn't make a background panel key at the OS level, so "no re-drop" there is the correct behavior, not a failure), and I didn't reproduce typing-over-a-fullscreen-Space headlessly — your real-IME verification covers both, and daily use will spot-check them for free.

Merging. Thank you for an exceptional contribution — the SkyLight root-cause analysis alone is worth the read, and turning the whole review around in a day with a round-trip refactor and exact-sequence tests is above and beyond. CJK users (myself included) will feel this fix immediately. 谢谢!

The pet-overlap-while-editing limitation stays with me as agreed — I'll track it in a follow-up issue, together with your mini-mode interaction idea from the first comment.

Issue #627, opened 2026-07-08
Bug: Windows 11: hook's process-snapshot PowerShell flashes a visible terminal window despite windowsHide:true (Windows Terminal default-terminal delegation)


Schlaflied:
## Symptom

A blank PowerShell/terminal window flashes on screen intermittently while working in Claude Code with Clawd on Desk installed. Unlike #586/#605 (taskbar-icon-only, fixed by #587), this is an actual visible console window. #596 may be the same root cause showing as a milder symptom.

## Environment

- Windows 11 Home 10.0.26200, **default terminal application = "Let Windows decide"** (i.e. Windows Terminal on current Win11 builds — `HKCU\Console\%%Startup` absent)
- Clawd on Desk build installed 2026-07-01, Claude Code CLI, hooks registered per the standard install
- Node.js v25.5.0 (`C:\nodejs` → nvm junction)

## Forensics (Windows 4688 process-creation audit, command line included)

Every Claude Code hook event fires `clawd-hook.js`, which calls `getWindowsProcessSnapshot()` in `hooks/shared-process.js` — one PowerShell spawn per resolve:

```
22:05:11  powershell.exe  PARENT: claude.exe        CMD: ...node clawd-hook.js PreToolUse     ← hidden, fine
22:05:12  powershell.exe  PARENT: node.exe (hook)   CMD: powershell -NoProfile -NonInteractive -Command "...ClawdWin32 typedef + Get-CimInstance Win32_Process..."   ← THIS one flashes
```

Over 48h of normal use this snapshot spawn fired 600+ times.

## Root cause

`shared-process.js` correctly passes `{ windowsHide: true }` (→ `CREATE_NO_WINDOW`) to `execFileSync`. But on Windows 11, when the OS-level **default terminal application** delegates console hosting to **Windows Terminal**, WT's takeover path does not always honor `CREATE_NO_WINDOW` — the spawned conhost/WT window becomes briefly visible (timing/cold-start dependent, which is why it's intermittent rather than every event). This is a known WT-side behavior (microsoft/terminal delegation issues), so Clawd's code is technically correct and still flashes.

Confirmed on this machine: the flash correlates 1:1 with the snapshot spawn timestamps in the 4688 log, not with the Electron hit-window path fixed in #587.

## Suggested fixes (any of these would do)

[...]

PR #628, opened 2026-07-08
fix(hooks): -WindowStyle Hidden on snapshot spawn + test-helper env restore (complements #630)

Schlaflied:
## Narrowed scope — defers to #630

The original version of this PR implemented a cross-process TTL snapshot cache. After reviewing #630, I'm withdrawing that part: the per-session shape there is the better design — it eliminates spawns on the high-frequency events entirely instead of racing a TTL against PID reuse and foreground changes, which is exactly the class of issues raised in review here. No point merging a second, weaker cache.

What remains is the complementary piece #630 doesn't cover:

- **`-WindowStyle Hidden` on the PowerShell snapshot spawn.** #630 still cold-spawns on SessionStart/UserPromptSubmit, and under Windows Terminal default-terminal delegation `CREATE_NO_WINDOW` can be ignored (#627) — the in-process flag shortens any window that still leaks through on those remaining fresh spawns.
- **Test-helper fix:** `load-shared-process-with-mock.js` now restores every env key it patches (previously only the hardcoded TMUX keys), so tests that pass extra `env` no longer leak state across cases.

`node --test test/shared-process.test.js` green on Windows 11.

And yes — once #630 ships in a release, I'll run the before/after on the 4688 audit rig and report back on #627.

Refs #627

### DIFF
--- a/hooks/shared-process.js
+++ b/hooks/shared-process.js
@@ -154,7 +154,11 @@ function getWindowsProcessSnapshot(execFileSync) {
     const out = execFileSync(
       "powershell.exe",
       [
-        "-NoProfile", "-NonInteractive", "-Command",
+        // -WindowStyle Hidden is belt-and-suspenders alongside windowsHide:
+        // when Windows Terminal is the OS default terminal app, its console
+        // delegation does not always honor CREATE_NO_WINDOW (#627), and the
+        // in-process flag shortens any window that still leaks through.
+        "-NoProfile", "-NonInteractive", "-WindowStyle", "Hidden", "-Command",
         WINDOWS_PROCESS_SNAPSHOT_SCRIPT,
       ],
       { encoding: "utf8", timeout: 3000, windowsHide: true, maxBuffer: 8 * 1024 * 1024 }
--- a/test/helpers/load-shared-process-with-mock.js
+++ b/test/helpers/load-shared-process-with-mock.js
@@ -1,7 +1,8 @@
 // test/helpers/load-shared-process-with-mock.js
 //
-// Patches child_process.execFileSync, process.env (TMUX/TMUX_PANE), and
-// process.platform before requiring shared-process.js. This is the
+// Patches child_process.execFileSync, process.env (TMUX/TMUX_PANE plus any
+// keys passed via `env`), and process.platform before requiring
+// shared-process.js. This is the
 // equivalent of load-focus-with-mock.js but for the hook-side resolver.
 
 function loadSharedProcessWithMock({ execFileSyncMock, env, platform }) {
@@ -12,7 +13,7 @@ function loadSharedProcessWithMock({ execFileSyncMock, env, platform }) {
   const origSp = require.cache[spKey];
   const origPlatform = Object.getOwnPropertyDescriptor(process, "platform");
 
-  const envKeys = ["TMUX", "TMUX_PANE"];
+  const envKeys = [...new Set(["TMUX", "TMUX_PANE", ...(env ? Object.keys(env) : [])])];
   const savedEnv = {};
   for (const k of envKeys) savedEnv[k] = process.env[k];
 


rullerzhou-afk:
Thanks for a genuinely excellent first PR — this is one of the most thorough contributions I've had here. The root-cause writeup (WT's default-terminal delegation not honoring `CREATE_NO_WINDOW`), the explicit correctness argument, the real-machine perf numbers, and especially the test isolation (`mkdtemp` + per-test `CLAWD_SNAPSHOT_CACHE_PATH`, and pinning `TTL=0` on the existing PowerShell-path tests) are all well above the bar. I checked it out locally and `node --test test/shared-process.test.js` is 51/51 green on Windows 11, including the win32-only PowerShell path.

I reviewed the change from a few angles and cross-checked the findings. No blockers — the core cache mechanics (freshness math, atomic temp+rename, fail-open on failure/empty, `TTL=0` opt-out, cross-process disk sharing, PID-miss bypass) are sound. One thing I'd want fixed before merge, then a few smaller items.

### Must-fix before merge

**1. Don't serve `foregroundWtHwnd` from the cached snapshot — it silently breaks focus targeting.**

`foregroundWtHwnd` is folded into the same TTL blob as the process tree (`hooks/shared-process.js:470`, read back at `:207`, returned cached at `:302`), but its consumers only report it on `SessionStart` / `UserPromptSubmit` (`hooks/clawd-hook.js:454` + `:302`, `hooks/codex-hook.js:209` + `:196`) — precisely the low-frequency, user-driven moments where the user may have *just* alt-tabbed or switched WT tabs.

Failure: hook A fires in WT window 1, writes the cache with hwnd-1. The user switches to WT window 2 and types a prompt (`UserPromptSubmit`) within the 5s TTL. That hook reads the disk cache (`fromCache: true`) and reports `wt_hwnd` = window-1. Downstream, `src/focus.js:480` (the `-not $focused` fallback) focuses the hook-provided hwnd after validating only `IsUsableWindowsTerminalWindow` (visible + `CASCADIA_HOSTING_WINDOW_CLASS`, `src/focus.js:129`) — it does **not** check the handle belongs to this session — and then caches it. Focus jumps to the wrong terminal.

The clean fix leans on your own perf insight: every event that needs `foregroundWtHwnd` is low-frequency, so keeping it fresh costs nothing against the burst case the cache targets (`PreToolUse → PostToolUse` never report `wt_hwnd`). Simplest: return `foregroundWtHwnd: null` on a cached read and don't persist it, so it's only ever populated from a fresh snapshot. Giving it its own ultra-short TTL works too.

### Should-fix (or scope the claim)

**2. A cached hit for a *present* PID isn't equivalent to the uncached path — PID reuse widens the stale-identity window from spawn-latency to the full TTL.**

The bypass at `hooks/shared-process.js:353` only triggers on a missing key (`!info`). If Windows recycles a PID within the TTL, `winSnapshot.get(pid)` returns a *hit* carrying the old process's `name`/`ppid`/`commandLine`, so no `forceFresh` fires and the walk reports a stale chain (wrong `source_pid`/`agent_pid`/editor). Pre-PR this race was bounded by spawn latency (~hundreds of ms); the cache stretches it to the default 5s. The comment's "correctness stays identical to today's uncached behavior" (`:354`) holds for the born-after-cache *missing* case but not the reused/stale-present one — worth narrowing that claim, and optionally shortening the default TTL or keying the cache by `startPid`.

**3. On a bypass refresh, the walk retries only the current PID, not from `startPid`.**

At `:358`, when a mid-chain ancestor is missing and `forceFresh` fires, the code re-looks-up the *current* pid against the fresh snapshot rather than restarting from `startPid`. If the staleness was mid-chain, the prefix already walked on cached data isn't re-derived, so it can still diverge from the uncached "take one fresh snapshot, walk from the top" behavior. Restarting the walk from `startPid` on that single fresh snapshot would make the equivalence actually hold.

**4. Disk-cache input validation is a little permissive and unbounded.**

`readSnapshotCache` (`:185`, `:191`) does `readFileSync` with no size cap, and `Number("123.5")` passes `Number.isFinite`, so a non-integer `pid`/`ppid` survives; `name` is stored as-is rather than lower-cased like the fresh path (`:263`), so a cached `"WindowsTerminal.exe"` wouldn't match the lower-cased `terminalNames` set. Under the default per-user `%TEMP%` this is same-user-trust only, but it's cheap to harden: `statSync` size cap, cap `processes.length`, require `Number.isInteger(pid) && pid > 0`, and `.toLowerCase()` the name on read.

### Nits / optional

- `hooks/shared-process.js:218` — the `.<pid>.tmp` is only unlinked on the exception path; a process killed between `writeFileSync` and `renameSync` leaves it behind. A best-effort sweep of stale `*.tmp` on write would cover it.
- Tests are strong, but a couple of the exact failure modes above aren't pinned: the disk round-trip for `foregroundWtHwnd` isn't asserted on the read side (`test/shared-process.test.js:715` sets a foreground payload but never checks it), and "cache covers the whole chain, zero spawns" (`:944`) locks in the riskiest path as the happy path. A PID-reuse case and a stale-foreground case would guard the fixes above.

None of this takes away from a really solid PR — happy to re-review quickly once #1 is in. And yes, please do verify a merged build against the 4688-audit setup you mentioned; that'd be the ideal confirmation.

Schlaflied:
@rullerzhou-afk Thanks for the thorough review — and after reading #630, I'm dropping the cache from this PR rather than patching it. Your per-session shape is simply the better design: it removes the high-frequency spawns instead of racing a TTL against PID reuse and foreground changes, which is exactly the class of problems your review flagged here. No point landing a second, weaker cache next to it.

This PR is now narrowed (+9/−4) to the two pieces that still complement #630:
- `-WindowStyle Hidden` on the snapshot spawn — #630 still cold-spawns on SessionStart/UserPromptSubmit, and that's where WT's delegation can leak a window (#627);
- the test-helper env-restore fix.

`node --test test/shared-process.test.js` is green on Win11. And the offer stands: ping me on #627 when #630 ships in a release and I'll run the 4688 before/after on this machine to confirm the flash is fully gone.

rullerzhou-afk (approved):
Approved. Re-reviewed the narrowed scope and cross-checked the findings with a second reviewer; everything verified on the merged tree:

- **Clean merge with #630** (`git merge-tree` three-way, no conflicts), and nothing on main depends on the snapshot args array by position or content — the win32 test mocks return fixed output regardless of args, and both deploy manifests list filenames only.
- **Real-machine check on the merged tree**: the flagged spawn returns a normal process chain with `snapshotOk: true`, and three-run timing (242/231/232 ms vs 242/221/222 ms baseline) shows no measurable startup cost from the flag. Output bytes are identical with and without it.
- **Targeted suites green** — 80/80 across shared-process, pid-cache, and clawd-hook-pid-cache on Windows 11, PowerShell path included. The full suite's single failure is unrelated and pre-existing: it turned out to be a console-codepage-dependent BOM issue in the Antigravity wrapper path (deterministic under CP 65001, invisible elsewhere), which I'm filing separately together with a flag-sweep follow-up for the three integrations that carry their own snapshot copies (`pi-extension.ts`, `opencode-plugin`, `hermes-plugin`) — those are reached by neither #630's cache nor the planned #634, so your flag is the right interim there too.
- **The helper fix is correct** — keys are saved before patching, Set-dedup handles overlap, undefined/null delete-and-restore round-trips. All existing callers pass only TMUX keys today, so it's preventive rather than fixing a live leak, which is exactly the kind of future-proofing worth having.

Also worth saying plainly: withdrawing your own working cache implementation after reading #630, with the reasoning spelled out, is exemplary. And the flag you kept is more valuable than your PR text claims — until #634 lands, the other twelve agent hooks still cold-spawn on every event, so this shortens their leak window as well.

Merging. Looking forward to the 4688 before/after once this ships in a release — #627 stays open for exactly that.

Issue #583, opened 2026-07-02
Claude Code sessions on Windows get sid=default — readStdinJson 400ms timeout loses stdin under PowerShell hooks


oneal32:
## Environment
  - OS: Windows 11 Pro (10.0.26200)
  - Clawd on Desk: prefs schema version 12 (latest as of 2026-07)
  - Claude Code: cc_version=2.1.177
  - Node: v24.13.0
  - Hook shell as auto-registered: `powershell`

  ## Symptom
  Claude Code sessions are not detected by Clawd. The pet stays asleep / the
  dashboard shows no live session, even though Claude Code is running and hooks
  fire.

  ## Root cause (with evidence)
  Hooks fire and reach the server, but `session_id` is lost → every real session
  collapses to `sid=default`, so Clawd can't distinguish/detect the session.

  From `session-debug.log` (timestamps redacted):
  Tx:xx:xx  event sid=test123  incoming=idle/SessionStart   ← manual echo JSON | node clawd-hook.js
  Ty:xx:xx  event sid=default  agent=claude-code agentPid=<PID> pidReachable=1  ← real Claude Code launch

  PID resolution succeeds (`agentPid=<PID> pidReachable=1`), but `sid=default`.

  `clawd-hook.js`:
  ```js
  const sessionId = payload.session_id || "default";

  payload comes from readStdinJson() in shared-process.js, which has a hard
  400ms timeout:
  timer = setTimeout(finish, 400);   // returns {} if stdin not fully read in 400ms

  On Windows the auto-registered hooks use "shell": "powershell":
  "command": "& \"node\" \"C:/.../clawd-hook.js\" SessionStart"
  PowerShell as an intermediary appears to delay/drop the stdin JSON that Claude
  Code pipes to the hook, so readStdinJson times out at 400ms → returns {} →
  session_id falls back to "default".

  Minimal repro

  1. Let Clawd auto-register hooks (default, shell: powershell) on Windows.
  2. Launch claude in any project.
  3. Check %APPDATA%\clawd-on-desk\session-debug.log → entry shows sid=default.
  4. Right-click pet → Open Dashboard → no live session.

  Workaround that confirms the diagnosis

  Removing the powershell shell and calling node directly (so Claude Code's
  default shell, Git Bash, runs the hook) restores detection:
  "command": "node \"C:/.../clawd-hook.js\" SessionStart"
  (delete the "shell": "powershell" line)

  Suggested fix

  Either:
[...]

PR #641, opened 2026-07-10
fix(hooks): stop Antigravity wrapper leaking a UTF-8 BOM into hook stdin (#638)

### LINKED ISSUE #638: Antigravity wrapper leaks a UTF-8 BOM into hook stdin on CP 65001 consoles; shared readStdinJson then drops the whole payload
## Symptom

`test/antigravity-install.test.js` — "does not stall on Windows when the hook runner never closes stdin (#568)" — passes or fails **deterministically depending on the console code page**. It is not flaky:

```
cmd /c "chcp 437 >nul   & node --test --test-name-pattern=\"does not stall on Windows\" test/antigravity-install.test.js"   → pass
cmd /c "chcp 65001 >nul & node --test --test-name-pattern=\"does not stall on Windows\" test/antigravity-install.test.js"   → fail: got '﻿' (a lone UTF-8 BOM), expected ''
```

Found while cross-reviewing #628: one reviewer environment (CP 65001 console) reproduced the failure 3/3 while another (non-65001 console) got 0 failures on the identical tree. Both reports were accurate — the variable is the console the test suite runs in.

## Mechanism

Console CP 65001 → Windows PowerShell 5.1's `Console.InputEncoding` becomes UTF-8 **with preamble** → the .NET Framework `Process.StandardInput` StreamWriter uses that encoding → `Write($stdinText)` + `Close()` flushes the 3-byte BOM **even when `$stdinText` is empty** (hence the child receiving exactly `'﻿'` in the watchdog path). Wrapper write site: `hooks/antigravity-install.js:207`.

The codepage → BOM causation is pinned by the A/B repro above; the encoding-inheritance link is documented .NET Framework behavior.

## Production impact (the actual bug — not just a red test)

On CP-65001 machines (Windows 11 "Beta: Use Unicode UTF-8 for worldwide language support", or IDEs that launch their console as UTF-8), the Antigravity hook receives `﻿{...}` on stdin. `readStdinJsonDetailed` (`hooks/shared-process.js:424`) checks emptiness with `text.trim()` — which happens to strip U+FEFF in JS — but then parses the **unstripped** text. `JSON.parse('﻿{...}')` always throws in V8, so the whole payload is silently dropped and the hook fails open with degraded data.

The repo already guards this class elsewhere: `hooks/copilot-hook.js:423` strips a BOM inline and `hooks/json-utils.js` exports `stripUtf8Bom`. The shared stdin path is the one place missing the guard.

Possible connection: a dropped payload means a missing `session_id` — the same exit as the `sid=default` reports in #583. When #583 diagnostics come back, a BOM-shaped "Unexpected token" in `parseError` would tie these together.

## Proposed fix (both layers)

1. **Wrapper side** — write through a BOM-less writer. .NET Framework 4.x has no `ProcessStartInfo.StandardInputEncoding`, so wrap the base stream instead: `New-Object IO.StreamWriter($proc.StandardInput.BaseStream, (New-Object Text.UTF8Encoding($false)))`. This alone turns the #568 test green on 65001 consoles.
2. **Hook side (generic defense for every agent)** — strip a leading U+FEFF in `readStdinJsonDetailed` before `JSON.parse` (reuse `stripUtf8Bom`).

Until this lands, anyone running the full suite from a UTF-8 console (e.g. pwsh 7 sessions) will see exactly one red test — don't let it cast suspicion on unrelated PRs.

Refs #568, #628, #583.

rullerzhou-afk:
## Problem

On CP-65001 consoles (Windows 11 "Beta: Use Unicode UTF-8", or any UTF-8 console such as pwsh 7 sessions), the Antigravity fail-open wrapper delivered the hook's stdin with a 3-byte UTF-8 BOM prefix. `readStdinJsonDetailed` passed the unstripped text to `JSON.parse`, which rejects a leading U+FEFF — the whole payload was silently dropped and the hook failed open with degraded data. Deterministic A/B repro in #638 (`chcp 437` → clean, `chcp 65001` → BOM); this is also why the #568 test was permanently red on UTF-8 consoles.

## Root cause

More subtle than the issue's original guess: **.NET Framework's `Process.Start()` eagerly creates the StandardInput `StreamWriter` on `Console.InputEncoding` with `AutoFlush = true`, and the AutoFlush setter flushes the encoding's preamble into the pipe at `Start()` itself** ([Process.cs](https://github.com/microsoft/referencesource/blob/main/System/services/monitoring/system/diagnosticts/Process.cs#L2153-L2163), [StreamWriter.cs](https://github.com/microsoft/referencesource/blob/main/mscorlib/system/io/streamwriter.cs#L300-L335)). So the issue's proposed "write through a BOM-less StreamWriter over BaseStream" cannot work — the BOM is in the pipe before any write. (.NET Core has the same eager writer but wraps the encoding in a preamble-suppressing `ConsoleEncoding`; Framework never got `ProcessStartInfo.StandardInputEncoding`.)

## Fix

**Wrapper (`hooks/antigravity-install.js`):**
- **Preamble-gated encoding swap before `Start()`** — `if ([Console]::InputEncoding.GetPreamble().Length -gt 0) { [Console]::InputEncoding = UTF8Encoding($false) }`. The gate is the load-bearing design choice: a preamble only exists when the console is already CP 65001, so the gated `SetConsoleCP` only ever receives the codepage the console already has. Console-observable state never changes → **no save/restore, and the shared-console race class (concurrent wrappers interleaving save/restore, hard-kill between set and restore leaving a mutated CP) is structurally impossible**, not merely mutex-guarded. Verified: `chcp` identical before/after a wrapper run under both 437 and 65001. Fail-open where no console exists (ANSI encodings carry no preamble anyway).
- **Raw UTF-8 stdin write** via `GetBytes` + `BaseStream` — also fixes the old writer encoding non-ASCII payloads as ANSI under CP-936-style consoles (mojibake for CJK `cwd`/`session` fields once decoded as UTF-8 by the hook).
- **`$psi.StandardOutputEncoding = UTF8Encoding($false)`** (Framework 4.5+, process-local): the wrapper now decodes the hook's UTF-8 stdout correctly under ANSI consoles instead of degrading non-ASCII responses to `{}` at the ConvertFrom-Json gate. The wrapper's *own* stdout (final hop to the IDE runner) intentionally keeps `Console.OutputEncoding` — changing it would reintroduce exactly the shared-console mutation this PR removes; hook responses on the registered events are ASCII-only JSON today.

**Generic defense (`hooks/shared-process.js`):** `readStdinJsonDetailed` strips one leading U+FEFF before `JSON.parse` — covers every agent hook against any BOM-emitting intermediary. Same guard added to `codex-debug-hook.js`, the one stdin JSON reader that bypasses it.

## Review

Deep-reviewed by Codex (44 min, reference-source-level): root cause and both fix directions confirmed; two blockers found in the v1 save/set/restore approach (cross-process interleave leaving a shared console at 65001, stale second restore in the outer catch) — both resolved by the gated no-restore design in the second commit, which supersedes the reviewer's named-mutex suggestion by removing the shared state instead of serializing access to it. Reviewer-confirmed alternatives (temp-file stdin, reflection, pwsh 7, raw CreateProcess) all rated worse trade-offs.

## Tests

- BOM / lone-BOM parse cases for `readStdinJsonDetailed`.
- Real-PowerShell round-trip forcing CP 65001 (`chcp 65001>nul && …`, codepage asserted and restored) with a byte-exact CJK payload — pins both the BOM and the encoding half.
- Bridge-shape assertions: the gated swap precedes `Start()`, no save/restore text present, `StandardOutputEncoding` set, stdin written via `GetBytes` and not the StandardInput writer.
- Antigravity suite green under both `chcp 437` and `chcp 65001`; full suite **4788 tests, 4773 pass / 0 fail** (the #568 test is no longer red on UTF-8 consoles).

Known non-goal: non-ASCII fidelity of the wrapper's final stdout hop under ANSI consoles (see above — would require mutating the shared console's output CP).

Fixes #638. Refs #568, #583.

### DIFF
--- a/hooks/antigravity-install.js
+++ b/hooks/antigravity-install.js
@@ -166,6 +166,18 @@ function buildWindowsEncodedFailOpenNodeHookCommand(nodeBin, hookScript, event, 
     ";",
     "$text=''",
     ";",
+    // #638: .NET Framework's Process.Start() eagerly creates the StandardInput
+    // StreamWriter on Console.InputEncoding with AutoFlush=true, which flushes
+    // the encoding's preamble into the pipe at Start() itself — under a
+    // CP-65001 console the hook's stdin starts with a 3-byte UTF-8 BOM that
+    // JSON.parse rejects (and an ANSI console mojibakes non-ASCII payloads
+    // written through that writer). Pre-set a BOM-less UTF-8 before Start()
+    // and restore right after; fail open where no console exists (ANSI has no
+    // preamble, so there is nothing to fix there).
+    "$origInputEncoding = $null",
+    ";",
+    "try { $origInputEncoding = [Console]::InputEncoding ; [Console]::InputEncoding = New-Object System.Text.UTF8Encoding($false) } catch { $origInputEncoding = $null }",
+    ";",
     "try {",
     "$psi = New-Object System.Diagnostics.ProcessStartInfo",
     ";",
@@ -191,6 +203,8 @@ function buildWindowsEncodedFailOpenNodeHookCommand(nodeBin, hookScript, event, 
     ";",
     "[void]$proc.Start()",
     ";",
+    "if ($null -ne $origInputEncoding) { try { [Console]::InputEncoding = $origInputEncoding } catch {} }",
+    ";",
     "$stdoutTask = $proc.StandardOutput.ReadToEndAsync()",
     ";",
     "$stderrTask = $proc.StandardError.ReadToEndAsync()",
@@ -204,9 +218,16 @@ function buildWindowsEncodedFailOpenNodeHookCommand(nodeBin, hookScript, event, 
     ";",
     `try { $stdinReader = New-Object System.IO.StreamReader([Console]::OpenStandardInput()) ; $stdinTask = $stdinReader.ReadToEndAsync() ; if ($stdinTask.Wait(${stdinTimeoutMs})) { $stdinText = $stdinTask.Result } } catch {}`,
     ";",
-    "$proc.StandardInput.Write($stdinText)",
+    // Write raw UTF-8 bytes through the base stream rather than the
+    // Console.InputEncoding StreamWriter, so this write stays correctly
+    // encoded even where the InputEncoding pre-set above failed (#638).
+    "$stdinBytes = [System.Text.Encoding]::UTF8.GetBytes($stdinText)",
     ";",
-    "$proc.StandardInput.Close()",
+    "$stdinStream = $proc.StandardInput.BaseStream",
+    ";",
+    "$stdinStream.Write($stdinBytes, 0, $stdinBytes.Length)",
+    ";",
+    "$stdinStream.Close()",
     ";",
     `if ($proc.WaitForExit(${timeoutMs})) {`,
     "$proc.WaitForExit()",
@@ -223,7 +244,7 @@ function buildWindowsEncodedFailOpenNodeHookCommand(nodeBin, hookScript, event, 
     ";",
     "$text=''",
     "}",
-    "} catch { $text='' }",
+    "} catch { $text='' ; if ($null -ne $origInputEncoding) { try { [Console]::InputEncoding = $origInputEncoding } catch {} } }",
     ";",
     "if ($text.Length -gt 0) { $trimmed=$text.Trim(); if (($trimmed.Length -lt 2) -or ($trimmed[0] -ne '{') -or ($trimmed[$trimmed.Length - 1] -ne '}')) { $text='' } else { try { $null = ($text | ConvertFrom-Json -ErrorAction Stop) } catch { $text='' } } }",
     ";",
--- a/hooks/antigravity-install.js
+++ b/hooks/antigravity-install.js
@@ -169,14 +169,15 @@ function buildWindowsEncodedFailOpenNodeHookCommand(nodeBin, hookScript, event, 
     // #638: .NET Framework's Process.Start() eagerly creates the StandardInput
     // StreamWriter on Console.InputEncoding with AutoFlush=true, which flushes
     // the encoding's preamble into the pipe at Start() itself — under a
-    // CP-65001 console the hook's stdin starts with a 3-byte UTF-8 BOM that
-    // JSON.parse rejects (and an ANSI console mojibakes non-ASCII payloads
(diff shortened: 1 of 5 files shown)


rullerzhou-afk:
Pre-merge deep review (OpenAI Codex, adversarial pass over `main...fix/638-antigravity-stdin-bom`). **No blockers.** All five focus checks hold:

1. **Eager-writer race genuinely eliminated, not reshuffled**: within one PowerShell process the ordering is strict — `InputEncoding` gate/set completes → `ProcessStartInfo` configured → `$proc.Start()` creates the AutoFlush stdin writer, which reads the already-cached BOM-less `UTF8Encoding`. The payload then goes through `BaseStream` as raw UTF-8 bytes, bypassing the StreamWriter. Concurrent final-version wrappers either all skip the CP write or all write the already-present 65001; no stale restore.
2. **BOM strip is chunk-safe**: buffers are `Buffer.concat`-ed before UTF-8 decode and U+FEFF strip, so `EF | BB BF` and `EF BB | BF` splits both parse (verified empirically).
3. **No runtime regression for other agents**: the custom `ProcessStart` wrapper is Antigravity-only; qwen shares only the generic encoded-command helper; Claude/Kimi never hit the CP path. The shared reader change only makes previously-invalid BOM-prefixed JSON parse.
4. **400ms default read budget intact** across early-EOF / no-data / partial-data / never-closed-stdin branches; Claude's explicit 2000ms (#583) unaffected.
5. **No #568 / #583 regressions**: the 2s stdin watchdog, child-stdin close, fail-open, raw-byte diagnostics and `sid=default` fallback all unchanged; BOM payloads now recover the real `session_id` instead of falling back.

Two non-blocking findings, recorded for follow-up rather than fixed here:

- **CONCERN (test isolation, not product code)**: the new test's `chcp 65001` mutates the whole attached console while `node --test` runs files concurrently; the `finally` restore narrows but cannot close the window in which other PowerShell-spawning tests observe 65001, and a killed test process skips the restore entirely. Windows full-suite runs can become timing-sensitive.
- **NICE-TO-HAVE**: a theoretical TOCTOU with an *external* process flipping the console CP between the wrapper's read and set — not reachable from anything in this repo and irrelevant to the #638 fix.

Targeted suites (Antigravity Windows, #568, #638, reader) all pass on the branch; the known-red `does not stall on Windows (#568)` case goes green here as expected.

Issue #640, opened 2026-07-10
macOS: pet (SkyLight private space) can cover — and eat clicks over — a text-input bubble while typing (follow-up to #626)


rullerzhou-afk:
> **中文摘要**:打字时气泡已降到普通层(#626),但桌宠和它的透明点击层还在 SkyLight 私有 Space(实测层级 1500)——几何上叠住时,桌宠不但盖住气泡,**被叠住的区域连点击都会被透明层吃掉**。低频(默认布局气泡贴着桌宠不重叠;触发=拖拽/挤压/自由漫游走过去),但值得修。已做两轮真机实验:hide+show 不会重置 Space 成员资格、第二低层 Space 的搬运调用报错——**没有捷径,需要真正的原生反向操作**;好消息是"放回去"(`applyStationaryCollectionBehavior`)已验证幂等且 4/4 次恢复成功。分阶段方案见下。

## Context

#626 fixed the IME candidate window being occluded: text-input bubbles now drop to the normal window level while a field is focused. But the **pet window (and its transparent hit/click-catcher window) stay in the SkyLight private space** at `CGAssistiveTechHighWindowLevel` — measured at layer **1500** vs the editing bubble at **0**. When they geometrically overlap:

1. the pet **visually covers** the bubble being typed into, and
2. — worse, not mentioned in #626 — the **transparent hit window eats clicks** over the overlapped region (it sits at 1500 too, tracking the pet's bounds via `syncHitWin`).

Frequency is low: `computeBubbleStackLayout` places bubbles beside the pet by design. Real triggers: user drags the pet onto a bubble, cramped work areas, or free roam wandering into it (no roam-yields-to-bubbles logic exists today).

## Real-machine experiments (2026-07-10, macOS, instrumented)

Method: two test windows — P (pet stand-in, given the full `applyStationaryCollectionBehavior` treatment) and B (bubble stand-in, plain `setAlwaysOnTop(true, "screen-saver")` = level 1000). Read `CGWindowListCopyWindowInfo` front-to-back order + `kCGWindowLayer` after each step. A window still in the private space stays **in front of B even at window level 0** (the space's absolute level dominates) — that inversion is the membership probe.

| step | result |
|---|---|
| P stationary, level dropped to 0 via `setAlwaysOnTop(false)` | **still in front of B** — level alone can't fix this (independently re-confirms #626's root cause) |
[...]

PR #642, opened 2026-07-10
fix(pet): freeze the editing bubble, pause roam, and fade + click-through the pet while it overlaps the box being typed into (#640 phases 0+1)

rullerzhou-afk:
> **中文摘要**:#640 的 Phase 0+1。打字进气泡输入框时:①气泡钉住不动(不再被 followPet 拖着跑)+ 自由漫游暂停 ②桌宠若正好叠在这个气泡上,淡化到 18% 并放行鼠标点击(点得到它身后的输入框和按钮),失焦/气泡关闭/移开重叠即恢复。全部 mac-only,复用 #626 的 editing 信号。三层验证全过:4780/0 全量、仪器化真机(透明度 1→0.18→1 全链路)、真人上手(淡化+点穿生效、拖拽穿越不断手)。

Phases 0+1 of #640 — the pure-Electron mitigation for "the pet covers (and eats clicks over) the bubble being typed into". All behavior keys off the `__clawdMacImeEditing` signal #626 introduced, macOS only; Windows/Linux untouched.

## Phase 0 — hold still while typing

- **The bubble being typed into freezes**: `repositionBubbles` skips applying bounds to a bubble whose text field is focused, so `followPet` anchoring can't yank the input box around mid-composition (pet drag, mini crabwalk, display churn). Fresh bubbles never carry the flag and still get placed.
- **Free roam pauses**: `isRoamAllowed()` gains a per-frame `isImeEditingActive` gate — an editing start cancels a walk mid-stride within one frame. The cancel path now funnels through `cancelRoam()`, which also fixes a pre-existing nit: a gate with no incoming state of its own used to strand the pet frozen in its walk pose (the walk-state SVG kept showing); now it restores `idle`.

## Phase 1 — step back when overlapping

`syncImeEditingPetDodge()` in `topmost-runtime.js`, running at the end of every `reapplyMacVisibility` pass (the #626 single-source-of-truth hook point) and on pet moves via `repositionAnchoredFloatingSurfaces`:

- While an editing bubble exists AND the pet sprite (`getHitRectScreen`, falling back to window bounds) intersects it: the pet render window fades to **0.18** (160 ms, via the existing `animateWindowOpacity` helper) and the hit window goes **click-through**, so the box being typed into stays readable and clickable underneath.
- Edge-triggered, with every transition path funneled: text-field blur (`handleImeEditing`), bubble teardown (`removePendingPermission` now re-runs the visibility pass), and overlap changes from pet movement.
- The hit window's ignore-mouse keeps exactly two writers, **platform-disjoint** (Windows settings-size-preview protection vs this macOS-only dodge), so they can never fight — the old "PERMANENT" comment's intent is preserved and documented at the arbiter.

## A bug the test suite missed — caught by instrumented real-machine verification

The first instrumented run showed `ime: true` with obvious geometric overlap and **no fade**: `getHitRectScreen` returns hit-geometry rects as `{ left, top, right, bottom }`, while the intersect test read `{ x, y, width, height }` — width coerced to 0, overlap never detected. The unit fixtures had encoded the same wrong assumption, so they were green. Fixed with a shape-normalizing `rectsIntersect`, and the fixtures now feed the production shape.

## Verification

- `npm test`: **4780 pass / 0 fail** (14 new cases: 8 dodge, 4 freeze/teardown, 2 roam).
- Instrumented real machine (dev app + injected elicitation + main-process inspector): focus → opacity 0.18, blur → 1, refocus → 0.18; bubble bounds frozen throughout; an editing bubble torn down mid-edit (remote-approval dismissal) restored cleanly.
- Hands-on (maintainer, real mouse + real IME): fade + click-through confirmed working; **dragging the pet across the editing bubble stays smooth through the mid-drag fade/restore transitions**; "pet is fully click-through while overlapped mid-edit — click anywhere to blur first, then grab it" reviewed and accepted as designed behavior (a hover-wake alternative was considered and rejected: it re-introduces click stealing over the overlapped region).

Phases 2 (native SkyLight de-delegation) and 3 (mini-mode tuck-behind, @Yike-Ye's idea) remain open on #640 — this is the stopgap that makes typing usable today.

Part of #640.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -1217,6 +1217,10 @@ let getSessionHudWindow = () => null;
 const themeFadeSequencer = createThemeFadeSequencer({
   getRenderWindow: () => win,
   getHitWindow: () => hitWin,
+  // #640: while the pet dodges an editing bubble its baseline opacity is the
+  // faded value, not 1 — restoring to 1 mid-edit would plant an opaque pet on
+  // top of the box being typed into. (Lazy: topmostRuntime is defined below.)
+  getRestoreOpacity: () => topmostRuntime.getPetTargetOpacity(),
   fadeOutMs: THEME_SWITCH_FADE_OUT_MS,
   fadeInMs: THEME_SWITCH_FADE_IN_MS,
   fallbackMs: THEME_SWITCH_FADE_FALLBACK_MS,
@@ -1334,6 +1338,10 @@ const _permCtx = {
   getTextScale: () => getTextScaleForPetWindows(),
   guardAlwaysOnTop,
   reapplyMacVisibility,
+  // #640: permission.js re-runs the editing-overlap dodge scan whenever the
+  // pendingPermissions list changes (notifyPermissionsChanged), so a bubble
+  // that leaves the list mid-edit can't strand the pet faded + click-through.
+  syncImeEditingPetDodge: () => topmostRuntime.syncImeEditingPetDodge(),
   isAgentPermissionsEnabled: (agentId) =>
     _isAgentPermissionsEnabled({ agents: _settingsController.get("agents") }, agentId),
   // DANGER "auto-pilot": when true, showPermissionBubble auto-approves every
--- a/src/main.js
+++ b/src/main.js
@@ -1283,6 +1283,8 @@ const topmostRuntime = createTopmostRuntime({
   getContextMenuOwner: () => contextMenuOwner,
   getNearestWorkArea,
   getPetWindowBounds,
+  // #640: tight sprite rect for the editing-overlap dodge test
+  getHitRectScreen: (bounds) => getHitRectScreen(bounds),
   getShowDock: () => showDock,
   isDragLocked: () => petWindowRuntime.isDragLocked(),
   isMiniAnimating: () => _mini.getIsAnimating(),
@@ -1437,7 +1439,12 @@ function repositionFloatingBubbles() {
 }
 
 function repositionAnchoredFloatingSurfaces() {
-  return floatingWindowRuntime.repositionAnchoredSurfaces();
+  const result = floatingWindowRuntime.repositionAnchoredSurfaces();
+  // #640: pet bounds changed — re-evaluate the editing-overlap dodge (a drag
+  // can slide the pet over the bubble being typed into; the bubble itself is
+  // frozen while editing, and roam is paused, so the pet is the mover here).
+  topmostRuntime.syncImeEditingPetDodge();
+  return result;
 }
 
 function syncSessionHudVisibilityAndBubbles() {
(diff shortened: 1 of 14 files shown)


rullerzhou-afk:
Pushed two rounds of review fixes on top of the original implementation:

- `a3e7e64` — the restore-on-close hook was wired to `removePendingPermission`, which no production close path uses; every real path (Allow/Deny click, Enter submit, auto-close timer) splices inline via `resolvePermissionEntry`, stranding the pet faded + click-through with no blur ever coming. The dodge re-scan now funnels through `notifyPermissionsChanged`, which every pendingPermissions add/remove already passes through. Also: the theme-switch fade restored the render window to a hardcoded opacity 1; it now asks `getPetTargetOpacity()` at restore time.
- `39fa7a0` — closes three gaps found by a follow-up codex review: (1) the dodge fade and theme fade are independent animation loops with separate cancel signals, so a baseline flip during the ~180ms fade-in let the last-finishing loop win — fade-in completion now re-reads the baseline and converges; (2) hit geometry can change without the window moving (state switches between hitboxes, theme reload) — `syncHitWin()` now re-runs the overlap scan; (3) a crashed bubble renderer can never send the focusout/blur IPC that clears the editing flag — `render-process-gone` now clears it. Plus a defensive split for the mid-drag case: the hit window's click-through write is deferred while a drag is in flight (the fade still runs — that transition is the hands-on-verified experience) and re-applied on drag-lock release / drag-end.

Full suite: 4822 tests, 4804 pass / 0 fail / 18 platform skips.

## Real-machine verification checklist (macOS, before merge)

- [ ] **Enter submit / auto-close while focused** — with the pet overlapping the bubble and a text field focused (pet faded to 0.18, click-through), submit via Enter — or let the auto-close timer fire — without ever clicking elsewhere first. The pet must return to full opacity and become clickable again.
- [ ] **Release a drag inside the overlap** — start dragging the pet from a clear spot, drag it across the bubble being typed into, and release the mouse while still overlapping. The drag must end cleanly, the pet should then be click-through over the box (input stays usable), and once editing ends a single click on the pet must work immediately — no "extra click to unstick".
- [ ] **Theme switch mid-edit** — with the dodge active, switch theme/variant from Settings. After the reload the pet must come back at the faded value, not snap opaque over the input box.
- [ ] **Hitbox switch mid-edit** — while typing, trigger a state that swaps the hitbox (e.g. a notification with the wide hitbox), then let it fall back. The fade/click-through must follow the changing overlap answer in both directions.
- [ ] **Control group** — Cmd-Tab away mid-composition, mouse-click on Allow/Deny, and the CJK IME candidate window surfacing all still behave as before.

Will merge once all five pass.

PR #647, opened 2026-07-10
refactor(telegram): retire dead migration-card layer; fix #638 test console pollution

### LINKED ISSUE #638: Antigravity wrapper leaks a UTF-8 BOM into hook stdin on CP 65001 consoles; shared readStdinJson then drops the whole payload

rullerzhou-afk:
Clears both tech-debt items left from the sidebar consolidation and the #641 codex review, in place of a tracking issue.

**1. Telegram migration: retire the dead card layer, keep the live runtime (−437)**

Scouting confirmed the v0.9.0 migration machinery is **not** dead code: the reducer + owner-manager own the sidecar/native runtime, and the Step-3 enable switch dispatches `USER_TEST_NATIVE` / `USER_DISABLE` through it. What actually died with the card UI is only the rendering layer, and that is exactly what this removes:

- `buildTelegramMigrationCard` / `renderMigrationCard` / button helpers / `deleteLegacyTokenFile` and the null `migrationCardEl` plumbing; the snapshot refresh, `migrationDispatch`, and the switch-gating helpers stay, with a comment explaining why the block is alive.
- `telegramApproval.deleteTokenFile` action + main.js helper removed (the card button was the only caller).
- Renderer-callable migration events narrowed to `USER_TEST_NATIVE` + `USER_DISABLE`; legacy-enable/rollback transitions remain in the reducer for main-side integrity but are no longer reachable from the renderer.
- 29 dead `telegramMigration*` i18n keys × 5 languages removed (`telegramMigrationErrorToast` stays — the switch path uses it); `tg-migration` CSS removed brace-aware, balance verified (lesson from #646 applied).

**2. #638 test: stop mutating the shared console codepage when it is already 65001**

The codex CONCERN from the #641 review: the BOM test forces `chcp 65001` on the console shared by concurrently-running test files. On systems already at 65001 (UTF-8 system codepage — increasingly the default) the mutation is now skipped entirely: same BOM coverage, zero cross-test pollution. Legacy-ANSI consoles keep the existing restore-bounded window. An isolated console via `detached`/`windowsHide` was probed and rejected — it breaks stdio piping for this very test.

Full suite: 4772 tests, **0 failures** (first fully-green run — #641's merge extinguished the last standing red).


rullerzhou-afk:
Pre-merge adversarial review (OpenAI Codex) over `main...refactor/retire-tg-migration-machinery`. **0 blockers, 1 concern — fixed in `f8c78da`.**

All six author claims were verified independently:

- **Dead code**: `buildTelegramMigrationCard` had definition-only status; `migrationCardEl` could only be assigned by it, hence permanently null. The module exposes only `init` on `globalThis` — no dynamic/string-keyed/test entry reaches the removed functions.
- **Live path**: `render()` → snapshot refresh → Step-3 switch visual state → `USER_DISABLE`/`USER_TEST_NATIVE` dispatch traced end to end; the removed `renderMigrationCard()` calls had no side effects beyond painting the null card (first-line guard returned immediately).
- **Whitelist**: renderer production code dispatches only the two retained events; `USER_ENABLE_LEGACY`/`USER_ROLLBACK_TO_LEGACY` survive solely in the reducer/controller and their unit tests.
- **Action removal**: the `deleteTokenFile` helper was reachable only through the removed command dep; `isTelegramTokenFileRequiredByNative` served only that action — nothing orphaned.
- **i18n/CSS**: 29 removed keys and all `tg-migration-*` selectors are zero-reference; the kept `telegramMigrationErrorToast` resolves in all five language packs.
- **BOM test**: with the console pre-set to 65001, a probe confirmed `GetPreamble().Length=3` — the wrapper's BOM branch is genuinely exercised in the skip-chcp mode; the CP936 path restores correctly.

The one concern: `docs/guides/telegram-approval.md` still instructed testers to exercise the retired card actions (enable legacy / rollback / delete-legacy-token → `TOKEN_FILE_IN_USE`), making that checklist unexecutable. Fixed by replacing the one-time migration dogfood section with a steady-state transport note pointing to git history.

Codex validation runs: targeted suites 346/346; full suite green apart from three pre-existing sandbox-denied `Get-CimInstance` cases unrelated to this PR. (Also corrected: the PR description said three commits; it was two before the docs fix.)

PR #653, opened 2026-07-11
feat: restore active Claude sessions after restart

200780381:
## Summary

Restore still-running Claude Code sessions after Clawd restarts, using the real session ID and the last truthful sustained state.

- write a minimal durable Claude recovery lease before the hook HTTP POST
- restore only `thinking`, `working`, and `juggling` after the local HTTP server is listening
- validate schema, age, enabled + installed integration state, live PID, and process start identity
- write ordered inactive tombstones for confirmed Stop, SessionEnd, StopFailure, and ApiError
- mark restored rows as recovered and non-focusable until the next real hook replaces the marker in place
- retain the existing five-minute no-session process keep-awake fallback for integrations without durable evidence

## Safety and lifecycle boundary

- filenames hash the real session ID; records exclude prompts, tool inputs/results, assistant output, permission data, and secrets
- remote SSH and WSL lease persistence fail closed because their filesystems are not locally authoritative
- lock publication is atomic; pruning locks and re-reads before deletion; stale/dead records cannot disable newer valid recovery
- Windows, macOS, and Linux active leases require process start identity, preventing PID reuse from restoring the wrong session
- capacity-full recovery fails closed instead of evicting an existing real session
- the lease writer and `state.js` share one Claude Stop classifier, so background tasks, session crons, `stop_hook_active`, headless debounce, and completion semantics cannot drift
- startup restoration bypasses the normal event path and does not replay completion, permission, celebration, sound, Telegram/Discord/LAN notifications, or `requiresCompletionAck`
- no synthetic `startup-recovery:*` session, inferred `thinking`, persistent Done chip, or completion-retention change is present

Only Claude Code has durable task-level recovery in this vertical slice. Other integrations retain the conservative process-only keep-awake hint; Codex keeps its existing JSONL behavior.

## Current-main compatibility

- rebased directly onto `origin/main` `947f79c9aadabfba8ad0b91c580abcf72a0d0e88`
- all current built-in agents explicitly declare `startupRecoveryProcessNames`
- MiMo Code is covered; Cursor, QoderWork, and WorkBuddy keep empty GUI recovery surfaces
- Windows only-Claude and only-Codex query regressions cover scalar PowerShell pipeline output

## Validation

- focused recovery/state/HUD suite: `323 pass / 0 fail`
- resolver public-shape/offline-gate/single-Windows-snapshot suite: `25 pass / 0 fail`
- changed JavaScript entry points: `node --check` passed
- `git diff --check origin/main...HEAD` passed

The repository-wide suite was also attempted both through the project runner and with `--test-concurrency=1`. This host still has unrelated environment failures: Python child processes return `status=null`, Electron-dependent tests cannot load the local Electron install, a real `cmd.exe` path test loses the non-ASCII Windows username, and the host CIM provider returns an empty process list. No directly related recovery, state, HUD, completion-ack, resolver-contract, or deployment-manifest test failed.

A real interactive Claude Code long-tool-call restart smoke has not yet been completed on this host; the offline hook ordering and restart loader paths are covered by automated process-level tests.


rullerzhou-afk (changes requested):
@200780381 Thanks — startup recovery is a real gap (the pet sitting idle while an agent is already mid-task is exactly the kind of thing Clawd should get right), and parts of this are thoughtfully built: the badge-derivation path structurally keeps recovered sessions out of completion notifications, Telegram/Feishu pushes, and the celebration gate, i18n is complete across all five locales, and detection fails closed. I did a deep multi-pass review (including running the full suite against your branch — 4783 tests, green). The findings below are ordered by what blocks merging.

## Blocking

**0. Commit authorship — this one first, and it also affects #652.** Every commit on this branch (and all seven on #652) is authored as `rullerzhou-afk <rullerzhou-afk@users.noreply.github.com>` — my identity. I assume this is an environment accident (a cloud workspace defaulting `user.email` from the repo's history rather than your account?) and not intentional, but it can't merge this way: GitHub attributes the work to me, a squash would generate `Co-authored-by` trailers pointing at me, and you would lose credit for your own contribution entirely — while I'd carry the blame line in every future `git blame`. Please set your git identity, rewrite the authors (`git rebase -r --exec 'git commit --amend --reset-author --no-edit' origin/main`), and force-push both PRs. Happy to help debug where the identity leaked from if you share what tool produced the commits.

**1. The `startupRecoveryProcessNames` fallback silently widens detection for every agent that doesn't declare it.** Only qoder/qoderwork got the narrow list; the other agents fall back to their full `processNames` — which resurrects two false positives the old hard-coded lists deliberately avoided: `Cursor.exe` (the whole IDE — merely having Cursor open at Clawd startup now fabricates a "Cursor Agent: thinking" row; see the review-fix note on `bb782b4`) and the bare `pi` process name on POSIX (the old pgrep deliberately matched Pi by package-path substring only, with a `doesNotMatch(/pgrep -x 'pi'/)` test guarding exactly this). Meanwhile the POSIX rewrite lost the old `pgrep -f` command-line matching for node-shaped `codex`/`copilot`/`codebuddy` installs — only `claude-code`/`kimi-code` needles survived. Please make every agent declare `startupRecoveryProcessNames` explicitly (empty for GUI/ambiguous names), and restore a roster-style test that fails when an agent is missing from the detection surface.

**2. "Real events replace the recovered session" has two holes.** Both `removeStartupRecoverySessionsForAgent` call sites run *before* early-return branches that never write a session: `PermissionRequest` (for every agent except codex, that branch calls `setState("notification")` and returns without touching `sessions`) and `SubagentStop && !existing`. Failure mode for the first: Clawd restarts while an agent is running, the first event to arrive is a permission prompt — the recovered row is deleted, nothing replaces it, and after the bubble the pet drops back to idle while the agent is actually blocked waiting for approval. That's the exact scenario this feature exists to cover. Move the removal to paths that actually persist a session (or skip it when the branch won't write one).

**3. The Done chip needs to come out — and since it looks like an innocent two-line addition, let me spell out the history so this doesn't read as arbitrary.** A persistent Done chip was proposed in #295 back in May. We absorbed most of that PR's ideas into main over time (the running-dot pulse, the wider HUD) but deliberately did *not* take the Done chip: completion is already expressed by the unread bell (which self-clears once seen), and a persistent chip duplicates that signal while adding visual noise to every finished row. The `assert.doesNotMatch(...sessionBadgeDone...chip-done)` you flipped wasn't a stale leftover — it was the guard we placed to keep that decision from being silently reversed. I re-reviewed this decision today before writing this comment, and it stands. If you think completion visibility deserves another look, I'm genuinely open to that discussion — but as its own issue, not riding along in a startup-recovery PR.

**4. Sequencing with your other open PRs.** This branch has a verified merge conflict with #652 in `src/state.js` (I ran the merge: #652 renames/rewrites `shouldDropAntigravityPostStopToolUse`; this PR appends nine helpers at the same anchor). You currently have three PRs in flight that touch overlapping files, which multiplies rebase work for everyone. Let's do one at a time: **#618 → #652 → #653**. I'll review #652 next; please hold new PRs until these land, and rebase this one once #652 is in.

## Should fix in this PR

- **Recovery-window semantics are self-contradictory** (one root cause: a global boolean window driving per-agent sessions). (a) One ghost hitting its 90s TTL kills the *global* window and timer — with two agents recovered, the first expiry disables re-detection for the other. (b) The pre-existing global shutdown at the top of `updateSession` (main behavior, kept as-is) now means *any* agent's first real event kills *every* agent's recovery window, even though only that agent's ghost gets removed. (c) `dismissSession` (Dashboard ×) deletes the ghost but leaves the window active — the 10s sweep re-detects and resurrects the row the user just dismissed. Suggest making the window state per-agent, or at minimum: expiry only clears the global flag when no recovered sessions remain, and dismiss also ends recovery for that agent. Also: with the current TTL-kills-window behavior, the documented "5-minute overall window" effectively never outlives ~90s — either the code or the PR description should change.
- **Recovered rows are indistinguishable and clickable in the HUD and tray menu.** The Dashboard at least shows the localized "Startup recovery" last-event label; the HUD row looks identical to a real session and `session-focus.js` returns `canFocus: true` off the bare `sourcePid`, so clicking a ghost triggers a real OS focus attempt. Expose `startupRecovery` in the snapshot and give the row a marker + non-clickable treatment.
- **`keepAwakeWhileWorking` treats ghosts as work.** Recovered sessions are `thinking + headless:false`, so `anySessionInProgress()` starts the `prevent-app-suspension` power-save blocker off a mere process detection (including a Cursor false positive) for 90s–5min. Exclude `startupRecovery` sessions from `isSessionInProgress`.
- **Linux `ps` parsing:** the `(\S+)` comm capture breaks on names with spaces (`"Kimi Code"` → `"Kimi"`, currently rescued only by the separate `"kimi"` registration), and procps makes no width guarantee when output isn't a tty — a truncated `args` column can defeat the `\bapp-server\b` exclusion (mis-recovering Codex Desktop) or the needle matchers. Use `-ww` on Linux and a space-safe parse.
- **Test hardening:** the rewritten detection tests dropped the roster assertions (see Blocking-1), the TTL test doesn't assert `getStartupRecoveryActive()` (the window-kill regression is invisible to CI), the detect tests never inject `ctx.isAgentEnabled` (the production gate is untested), and the 90s-expiry test drives `cleanStaleSessions()` manually rather than the real 10s interval.

## Minor / follow-up is fine

- Renaming a recovered row in the Dashboard persists an alias keyed to the throwaway `startup-recovery:*` id (7-day junk). `muteNotificationSound: true` on ghosts is dead — nothing reads it outside `notification` state. The no-args fallback branch inside `startStartupRecovery()` is production-dead (only tests call it bare). QoderWork's shared `qodercli` backend can still fabricate a mislabeled "Qoder" ghost that QoderWork's own events won't clear (agentId mismatch) — worth a note in the registry comments even if you punt the fix.

Once the authorship is fixed and #652 lands, I'm glad to re-review quickly — the underlying feature is wanted.

200780381:
Thanks for catching the authorship issue. You're right: this was an environment/Codex commit setup problem, not intentional. The commits were produced through Codex from a workspace whose git identity was pointing at `rullerzhou-afk <rullerzhou-afk@users.noreply.github.com>`, so GitHub attributed the branch to you.

I'll fix the local git identity, rewrite the affected commit authors on this PR (and #652), and force-push the corrected history before asking for another review.

200780381:
Thanks again for the detailed review. I fixed the authorship issue first; #653 now shows all commits authored by `200780381 <87887733+200780381@users.noreply.github.com>`.

I also pushed a review-fix commit for #653:
- removed the implicit `startupRecoveryProcessNames` fallback and made every agent declare an explicit recovery surface
- kept startup recovery conservative for GUI/ambiguous agents such as Cursor, QoderWork, Qoder IDE, OpenClaw, and POSIX `pi`
- restored node-shaped detection for Codex/Copilot/CodeBuddy/Kimi/Claude installs and kept the Codex Desktop app-server exclusion
- fixed recovery replacement so early-return `PermissionRequest`/`SubagentStop` paths do not delete ghosts unless a real session is persisted
- made the recovery window track remaining recovered ghosts instead of being killed by any first event/TTL; dismissing a ghost now ends recovery for it
- exposed `startupRecovery` in snapshots, made recovered rows non-focusable, marked them in the HUD, and excluded them from keep-awake work detection
- removed the persistent Done chip while keeping the unread completion bell behavior
- hardened Linux process detection with `ps -ww` and space-safe parsing
- added/updated tests for the above

Checks run locally:
- `node --check src\state.js src\state-session-snapshot.js src\session-focus.js src\session-hud-renderer.js agents\registry.js agents\antigravity-cli.js agents\claude-code.js agents\codebuddy.js agents\codewhale.js agents\codex.js agents\copilot-cli.js agents\cursor-agent.js agents\gemini-cli.js agents\hermes.js agents\kimi-cli.js agents\kiro-cli.js agents\openclaw.js agents\opencode.js agents\pi.js agents\qwen-code.js agents\reasonix.js`
- `node --test test\registry.test.js test\state-startup-recovery-detect.test.js test\state.test.js test\state-session-snapshot.test.js test\session-focus.test.js test\session-hud-style.test.js` (`311 pass / 0 fail`)

Ready for another review when you have time.

200780381:
A quick clarification on Blocking item 4, and I agree with the sequencing request.

The `src/state.js` conflict looks like an anchor/ordering conflict rather than a business-logic conflict: #652 renames/rewrites `shouldDropAntigravityPostStopToolUse` into the broader post-stop activity handling for QoderWork, while #653 inserted the startup-recovery helper block immediately after that old anchor. Git cannot automatically place both changes, but the two behaviors are not mutually exclusive.

`src/state.js` is touched in #653 because startup recovery has to wire into the central session lifecycle: creating `startup-recovery:*` ghost sessions, replacing them only when a real persisted event arrives, preserving ghosts through early-return `PermissionRequest`/`SubagentStop` paths, expiring/dismissing recovery rows, process detection, and the startup recovery window. Without those hooks in `state.js`, the feature would only be visual and would not fix the idle-pet-while-agent-is-running case.

So the plan on my side is: hold #653 for now, let #618 and #652 go first, then rebase #653 after #652 lands and resolve the `src/state.js` anchor conflict by keeping both the #652 post-stop activity changes and the #653 startup-recovery lifecycle changes.

Also understood on the #618 sequencing/request: please prioritize/review #618 first in that chain. Once #618 and #652 are settled, I can rebase #653 cleanly instead of adding more overlap.

200780381:
Closing this PR for now to respect the requested sequencing and keep the review queue clean.

The startup recovery feature itself is still wanted, but this branch currently sits behind the active chain `#618 -> #652 -> #653` and overlaps `src/state.js` with #652. Rather than keep adding review noise here, I will wait until #618 and #652 are settled, then rebuild/reopen a narrower startup-recovery PR from the latest main.

Scope for the future PR will stay focused on startup recovery only:
- conservative explicit process detection
- `startup-recovery:*` ghost session lifecycle
- replacement only when a real persisted session arrives
- early-return `PermissionRequest` / `SubagentStop` safety
- TTL/dismiss behavior
- non-focusable recovered rows and keep-awake exclusion

I will keep the completion/HUD retention discussion separate and will not bring a persistent Done chip back through the startup recovery PR.

200780381:
Follow-up update: I reworked startup recovery into a dedicated runtime module, so the final PR diff no longer touches `src/state.js`.

The state machine remains authoritative for real sessions. Startup recovery now only manages temporary `startup-recovery:*` ghost rows through the existing public session map/snapshot/state APIs, and removes those ghosts once a real persisted session appears for the same agent. Early-return events that do not persist a real session still leave the ghost in place.

I also kept Blocking 3 intact: no persistent Done chip and no completed-session HUD lifecycle changes in this PR. The HUD change here is limited to marking recovered ghost rows with the localized `StartupRecovery` event.

Checks run locally:
- `node --check src\startup-recovery.js src\main.js src\state.js src\state-session-snapshot.js src\session-focus.js src\session-hud-renderer.js agents\registry.js agents\antigravity-cli.js agents\claude-code.js agents\codebuddy.js agents\codewhale.js agents\codex.js agents\copilot-cli.js agents\cursor-agent.js agents\gemini-cli.js agents\hermes.js agents\kimi-cli.js agents\kiro-cli.js agents\openclaw.js agents\opencode.js agents\pi.js agents\qoder.js agents\qoderwork.js agents\qwen-code.js agents\reasonix.js`
- `node --test test\startup-recovery.test.js test\registry.test.js test\state.test.js test\state-startup-recovery-detect.test.js test\state-session-snapshot.test.js test\session-focus.test.js test\session-hud-style.test.js` (`309 pass / 0 fail`)
- `git diff --check origin/main..HEAD`

rullerzhou-afk (changes requested):
Thanks for reworking this. I have now re-reviewed the current head (ace86b9) against current main (265e4fe), including the process matcher, recovery lifecycle, snapshot behavior, and merge compatibility.

I am leaving this PR open, and I will not close it before hearing your response. For the current implementation, though, my review remains **Request changes**.

## What the actual bug is

There is a real startup gap, but it is narrower than the PR description suggests.

Current main already scans for agent processes after launch and suppresses the idle/sleep sequence for up to five minutes. What it does *not* do is reconstruct a real session, so a hook-only agent can temporarily appear as an idle-but-awake pet with no HUD row until the next hook arrives.

Codex already has a stronger partial solution through JSONL backfill (including recovery of sustained thinking/working state and unresolved input in supported cases). The remaining generic gap is mainly agents for which hooks are the only durable source of session state.

## What this PR currently does

The new runtime turns a process-table match into a synthetic session:

1. find a matching process;
2. create one startup-recovery:<agent>:<pid> row per agent;
3. hard-code its state to thinking;
4. show that row as running in the HUD/pet;
5. remove it when a real event arrives, or after roughly 90 seconds.

That is not session recovery. It is an agent-level liveness guess presented as a task-level fact. The process table does not tell us whether the CLI is in a turn or sitting at its prompt, nor does it provide the real session id, cwd, title, task count, or per-session state.

## Blocking problems

### 1. The visible state is based on an insufficient signal and has broad false positives

applyEntries() assigns state: "thinking" to every detected entry even though the only evidence is that a process exists.

On POSIX, ps -axo pid=,comm=,args= is scanned as a whole and the command-line substring matchers apply even when the executable name is unrelated. Commands such as these are therefore attributable as active agents:

- vim codex-notes.md
- tail -f /var/log/codex-deploy.log
- rg codebuddy src/
- bash -c 'cd ~/codex-notes && npm test'

On Windows, the expanded node.exe needles have the same problem. I reproduced a false Codex detection with an unrelated process shaped like:

node.exe D:\projects\codex-notes\server.js

Similar paths for an editor Copilot language server or a project containing codebuddy can become synthetic CLI sessions. Unlike the existing behavior, this is not just a harmless extra keep-awake period: it visibly drives the pet to thinking, adds a HUD row, and exposes a running badge/state to downstream consumers.

An idle CLI at its prompt has the same false-positive outcome. Process existence cannot distinguish that case from an active turn.

### 2. It weakens the behavior that main already gets right

The existing startup keep-awake guard lasts up to five minutes while waiting for a real hook. In this PR, an unconfirmed ghost has a 90-second TTL; the next sweep removes it and stops recovery when no ghosts remain. In practice the useful recovery window becomes about 90–100 seconds.

That means the scenario this feature most needs to cover — restarting Clawd during a long tool call that emits no hook for several minutes — falls back to idle sooner than current main.

### 3. The lifecycle has two owners that can disagree

The startup runtime and state.js share the same sessions map but maintain separate bookkeeping. If the normal stale-session sweep removes a recovered row after its PID exits, it bypasses the runtime's deletion path, so the row can be gone while recoveryActive remains true until the outer timeout.

The old startup-recovery implementation also remains exported and tested in state.js, but main.js no longer consumes it. This leaves two implementations and two process rosters, one of which is effectively dead production code. That will be difficult to maintain safely.

### 4. The branch is not compatible with current main

The PR still conflicts in src/main.js. After resolving that conflict minimally and running the relevant tests, the new registry roster test fails because agents added after this branch's base (at least MiMo Code, with WorkBuddy requiring the same audit) do not declare startupRecoveryProcessNames.

The roster test itself is a good idea; the failure demonstrates why passing against the old base is not sufficient.

## Recommended redesign

I do want us to fix the startup gap, but the source of truth needs to move from **process-name inference** to **durable session evidence**.

### A. Have each hook-capable agent maintain a small recovery lease

Before attempting the HTTP POST to Clawd, the hook/plugin/extension should atomically update a local record for the real session. This must happen even when Clawd is offline.

A record should contain only non-sensitive recovery metadata, for example:

- schema version;
- agent id and real session id;
- last sustained state (thinking, working, etc.);
- event timestamp;
- stable terminal/agent PID plus process-start identity, where available;
- optional cwd/title fields already safe for the normal session snapshot.

Do not persist prompts, tool input/output, model text, permission payloads, or secrets.

Stop / SessionEnd should mark or remove the lease even if Clawd is not running. A long tool call remains recoverable because PreToolUse records working before the tool starts; an idle prompt is not recovered because Stop has cleared the active lease.

Use atomic replace/write and fail closed on malformed or partial files. A versioned location such as ~/.clawd/session-recovery/v1/<agent>/<session-hash>.json would keep this isolated from preferences.

### B. Restore only validated, sustained state

At startup, Clawd should load those leases and validate:

- the agent is installed/enabled;
- the record schema and fields are valid;
- the state is one we intentionally allow to survive a restart;
- the process is still alive and, when available, its start identity matches (to reject PID reuse);
- the record is not unreasonably old.

Restore sustained activity only. Do not replay one-shot states or effects such as completion, attention, permission prompts, celebration, sounds, outbound notifications, or approval flows.

Use the real session id so multiple sessions for one agent remain distinct. Mark the first snapshot as startup-recovered for UI explanation, but let the next real event update that same session normally rather than deleting a fabricated startup-recovery:* row.

### C. Keep process detection as a weak fallback only

For an integration that cannot provide logs or a durable lease, retain the honest current behavior: a conservative, explicit process allowlist may temporarily suppress sleep, but it must not create a HUD session or assert thinking.

Codex should continue using its existing JSONL backfill path rather than being downgraded to process inference. Claude Code is the best first lease implementation because its hooks already carry a real session id and transcript path. Other hook-only agents can follow after that path is proven.

### D. Suggested delivery sequence

I suggest splitting the replacement into small reviewable changes:

1. **Cleanup/hardening:** keep explicit startup process-name declarations and the roster test, remove the duplicate/dead detector, preserve current no-session keep-awake semantics.
2. **Claude Code recovery lease:** add the durable writer/clearer plus startup loader and side-effect suppression.
3. **Expand by integration:** adopt the lease in other hook/plugin/extension agents where their lifecycle semantics are verified; keep no-evidence agents on keep-awake-only fallback.

Because a shared hook helper would run in deployed/remote hooks, it must use Node built-ins only and be added to both deployment manifests. The Claude hook payload/lifecycle change also needs a real Claude Code smoke test, not only synthetic payload tests.

The key distinction is: **this PR guesses "thinking" from a process name; the replacement should recover the last truthful sustained state left by the actual session.**

Please reply with your thoughts, especially if there is an agent lifecycle that cannot support this lease model. Again, I will keep #653 open until you have had a chance to respond.

200780381:
Thanks for the careful re-review. I agree with the distinction you are making: process presence is only an agent-liveness hint, and it is not sufficient evidence to publish a task-level `thinking` session. The current synthetic `startup-recovery:*` design crosses that line, so I do not think it is worth patching that implementation further.

The recovery-lease direction makes sense. I would keep process detection as a conservative no-session keep-awake fallback only, keep Codex on its stronger JSONL backfill path, and prove the lease contract first with Claude Code. For Claude, the writer can run before the HTTP POST, use an atomic replace with Node built-ins only, persist only minimal non-sensitive sustained-state metadata, and reuse the existing completion gate: a `Stop` with `stop_hook_active`, live background tasks, or session crons must not clear the lease as though the turn were finished.

I audited the current integration lifecycles specifically as active-turn leases rather than session-lifetime records. There are two concrete integrations I would not enable for lease recovery today:

- Kiro does not provide a real session id to the hook, is currently collapsed to `default`, and has no SessionEnd capability. It cannot safely distinguish or terminate concurrent recoverable sessions.
- CodeWhale can fall back to a process-shared/generated id and has SessionEnd but no reliable normal turn-end/Stop event. A completed turn could leave an active lease until TTL, so it should remain keep-awake-only until there is a truthful turn-finished signal.

Several others look supportable, but only with agent-specific mappings rather than a universal `Stop` rule. For example, Gemini ends a turn with `AfterAgent -> idle`; Antigravity can emit `Stop` with `fullyIdle=false`, which means it is still working; Reasonix has deliberate Stop ordering protection; and OpenClaw/Hermes/Pi/QoderWork each need their real id and end/debounce behavior verified before enabling recovery. I would fail closed for missing/default ids and add integrations only after their lifecycle semantics and a real smoke test are documented. Codex is not a lease candidate because its JSONL evidence is already stronger.

I also agree with the proposed delivery sequence. I will rework this branch onto current main as the cleanup/hardening step: remove the synthetic-session runtime and duplicate/dead detector path, retain explicit conservative process declarations plus a roster test, and preserve the existing five-minute no-session keep-awake behavior without creating HUD rows. I will keep the Claude Code lease vertical slice, both deployment-manifest entries, and its real smoke test as the next focused change, then expand integration by integration.

The completion boundary remains unchanged: no persistent Done chip and no completed-session HUD retention or cleanup changes in this startup-recovery work.

200780381:
Follow-up pushed as `ced54f0` after the design response above.

I rebuilt the branch from current `main` as the cleanup/hardening step:

- removed the synthetic `startup-recovery:*` session implementation from the PR diff
- kept process detection as a boolean, no-session keep-awake fallback only
- added explicit startup process declarations for the full current roster, including MiMo Code
- filtered the query to enabled agents
- kept Cursor, QoderWork, and WorkBuddy GUI processes out of the fallback
- added roster and conservative-query regression tests

The existing five-minute startup window remains authoritative; this branch does not create HUD rows, assert `thinking`, or touch completion retention / Done-chip behavior.

Validation: targeted agent/state/tick suite `513 pass / 0 fail`, `node --check`, and `git diff --check main..HEAD`.

The Claude Code recovery lease will remain a separate vertical slice with the deployment-manifest updates and real Claude smoke test discussed in the review.

200780381:
One additional implementation note, including the concrete problems I encountered while rebuilding this branch.

## Final implementation boundary

- The synthetic `startup-recovery:*` runtime is completely gone from the PR diff.
- Process presence never creates a HUD row and never asserts `thinking`.
- Process detection is only a boolean fallback for the existing five-minute no-session keep-awake window.
- All 20 current built-in agents now explicitly declare `startupRecoveryProcessNames`; the roster test fails when a future agent omits that decision.
- MiMo Code, Qwen Code, Reasonix, and the rest of the current-main roster are covered.
- Long-lived GUI processes for Cursor, QoderWork, and WorkBuddy explicitly declare an empty startup surface.
- Qoder permits only the CLI names in this fallback; merely opening Qoder IDE does not match.
- Bare POSIX `pi` remains excluded; only the known package-path fallback is retained.
- The system query is built only from enabled agents. If the only enabled integration has no trustworthy process surface, no OS process scan is issued.
- There are no completion-retention, Done-chip, HUD, `requiresCompletionAck`, notification, or outbound-effect changes.

## Problems encountered during the rewrite

1. The previous head was no longer a useful rebase target. It was based on the older pre-#652 main, conflicted in `src/main.js`, and its roster assumptions predated at least MiMo Code and WorkBuddy. Rebuilding from current `main` (`f475cfc`) was substantially safer than resolving the synthetic runtime into today's session wiring.
2. The general `processNames` registry field is not conservative enough for startup keep-awake. It legitimately contains long-lived IDE/desktop process names used for other attribution purposes. Reusing it would reintroduce the Cursor/QoderWork/WorkBuddy false-positive class, so startup detection needed its own explicit per-agent surface.
3. The old combined detector only had a global `any agent enabled` gate. It could not remove disabled agents from the actual WMI/pgrep query. The new query is assembled after per-agent enable filtering.
4. Node-shaped CLI/package matching remains weaker than exact executable matching and can still be broad. I retained it only to preserve the existing no-session keep-awake behavior; its result is deliberately not exposed as session evidence or a task state.
5. The old branch's targeted tests were green but did not protect current-main roster drift or the `enabled GUI agent with an empty surface should not scan` case. Both are now explicit regression tests.

## Validation limitation

For transparency, I am not claiming the entire repository suite is clean in this local Windows Sandbox. A fresh dependency install was needed; the machine's Node 26/npm 11 combination failed inside npm with `Class extends value undefined`, so I used the bundled pnpm runtime for the isolated checkout. The full suite then still hit host-environment failures in Hermes Python child processes, Electron/permission host tests, temporary theme-cache filesystem tests, and the real Windows PID-tree test. Those failures are outside the changed paths, but they mean a full-suite green claim would be inaccurate.

The directly relevant agent/registry/main/state/tick suite is clean: `513 pass / 0 fail`, plus `node --check` and `git diff --check main..HEAD`.

200780381:
On Blocking 2 specifically: I agree that the previous 90-second ghost TTL was a regression over behavior main already had.

The rewritten head `ced54f0` fixes this by removing the ghost lifecycle entirely, not by choosing another shorter TTL:

- there is no `STARTUP_RECOVERY_SESSION_TTL_MS` and no synthetic-session stale sweep in the PR diff;
- the only startup recovery duration remains main's existing `STARTUP_RECOVERY_MAX_MS = 300000`;
- while there are no real sessions, the 10-second cleanup path only ends the guard when the conservative process scan returns no match;
- when the process is still present, a successful scan leaves `startupRecoveryActive` and its original five-minute timer untouched;
- `tick.js` continues resetting the sleep baseline while that flag is active;
- the first real hook still ends the guard because real session state has arrived and becomes authoritative.

So the long-tool-call scenario now retains the same behavior as current main: if the agent process remains alive and no hook arrives, Clawd suppresses idle/sleep for the full five-minute window. It no longer falls back after roughly 90-100 seconds, and no process match is presented as a recovered HUD session.

rullerzhou-afk (changes requested):
Thanks for narrowing this to the honest no-session keep-awake fallback. I re-reviewed the current head, `ced54f0`. The synthetic sessions / HUD / `thinking` path is gone, and the explicit per-agent recovery surface plus roster guard is the right scope.

I found one blocking Windows regression in `src/state.js:2185-2187`. When the enabled roster leaves exactly one CLI that has both an exact process name and a Node command-line needle—for example, only Claude Code or only Codex—both PowerShell pipelines yield scalar strings. `$nameFilters + $nodeFilters` therefore concatenates them before `-join` runs. The resulting WMI filter is shaped like:

```text
Name='claude.exe'(Name='node.exe' AND CommandLine LIKE '%claude-code%')
```

There is no ` OR ` between the clauses. `Get-CimInstance` fails, `detectRunningAgentProcesses()` returns false, and the existing five-minute startup guard never starts. The new enabled-agent test does not cover this because Qoder contributes two distinct exact names and no Node needle.

Please force both sides to arrays—for example, `(@($nameFilters) + @($nodeFilters)) -join ' OR '`—or build the complete filter-clause array in JavaScript, and add an only-Claude or only-Codex regression test.

Separately, I verified that the remaining broad POSIX `pgrep -f codex` can still match Codex Desktop helpers locally. That behavior already exists on `main` and this PR intentionally preserves it as a weak no-session hint, so I am not treating it as a new blocker here.

Validation on this exact head: the targeted suite passes `513/513`; the three changed entry points pass syntax checks; and `git diff --check` passes. Requesting changes only for the Windows single-CLI query bug.

200780381:
Thanks for the precise Windows reproduction. You were right: when each PowerShell pipeline produced one scalar, `$nameFilters + $nodeFilters` concatenated the clauses before `-join` could insert ` OR `. In `887c8cf` I changed it to:

```powershell
(@($nameFilters) + @($nodeFilters)) -join ' OR '
```

and added separate only-Claude and only-Codex regressions. Each verifies the exact executable clause, the Node command-line clause, and the array-safe join.

The five-minute behavior remains intact. Process detection is still only the existing no-session keep-awake fallback; it creates no session/HUD row and infers no `thinking` state. There is no 90-second ghost TTL. The branch is now rebased directly onto current `main` `947f79c`, and the PR is mergeable. The current roster guard covers all built-ins, including MiMo Code; Cursor, QoderWork, and WorkBuddy retain empty GUI recovery surfaces.

This update also completes the original restart goal for a focused Claude Code vertical slice. The hook writes minimal recovery evidence before its HTTP POST, and startup restores the same real session ID plus the last sustained `thinking` / `working` / `juggling` state only after the HTTP server is listening. The recovery path does not call the normal event path and does not replay completion, permissions, celebration, sound, outbound notifications, or `requiresCompletionAck`. There is no synthetic session ID, inferred task state from a process, persistent Done chip, or completion-retention change.

While implementing and reviewing that path, I found and fixed several additional correctness issues before pushing:

- a visible lock could previously exist before its owner record was written; locks are now prepared in a unique pending directory and atomically published only after the owner is complete;
- cleanup now removes dead/corrupt orphan locks but never breaks a lock whose owner PID is still alive, even when it is old;
- more than 100 expired/dead active leases could previously disable all recovery; invalid records are now validated and cleaned before the cap is applied to credible candidates;
- active leases on every local platform now require a process-start identity, so PID reuse cannot restore the wrong session;
- a new hook identity now wins over an older identity for the same numeric PID;
- pruning locks and re-reads deletion targets, skips the writer's newly written tombstone, and cannot delete a concurrently promoted active record;
- capacity-full recovery fails closed without evicting any of the 20 existing real sessions;
- prompt-derived fallback titles are never persisted;
- Windows start identities come from the existing single CIM snapshot as private non-enumerable resolver metadata, preserving the old ten-field resolver contract and avoiding a second PowerShell spawn;
- the lease writer and `state.js` now share the same Claude Stop disposition classifier, preventing background-task, cron, `stop_hook_active`, headless, and debounce semantics from drifting.

Post-rebase validation on `887c8cf`:

- recovery/state/HUD suite: `323/323`
- resolver public-shape/offline-gate/single-snapshot suite: `25/25`
- changed JavaScript syntax checks: passed
- `git diff --check origin/main...HEAD`: passed

I also attempted the repository-wide suite through the project runner and again with one test worker. Remaining failures are host-environment failures unrelated to this diff: Python child processes return `status=null`, Electron-dependent tests cannot load the local Electron install, `cmd.exe` corrupts the non-ASCII Windows username in one real-path test, and this host's CIM provider returns an empty process list. No directly related recovery, state, HUD, completion-ack, resolver-contract, or deployment-manifest test failed. I have not yet completed a real interactive long-tool-call Claude restart smoke on this host, so I am not claiming that manual check.

Could you please re-review `887c8cf`?

One separate Settings UX request, not part of #653: could you reconsider the conditional Office AI / Coding AI grouping proposed in #652? https://github.com/rullerzhou-afk/clawd-on-desk/pull/652#issuecomment-5044900296 . A narrow compromise would show labels/separation only when both categories are present, while retaining the current flat list when only one category exists. This is only a Settings follow-up and does not add WorkBuddy/QoderWork/session-state scope to #653.

rullerzhou-afk (approved):
Windows real-machine smoke passed: active Claude session survived a Clawd restart and restored from its lease; natural completion wrote an inactive tombstone; a second restart did not restore it. Full suite: 5840 tests, 5821 passed, 0 failed, 19 skipped.

rullerzhou-afk:
Thank you for the thoughtful implementation and for patiently iterating on the review feedback. We completed a real Windows + Claude Code restart smoke: the active session recovered correctly, natural completion wrote the tombstone, and a second restart did not restore the finished session. This is now merged — really appreciate your contribution!

PR #113, opened 2026-04-16
feat: add Korean (ko) locale

Tonic-Jin:
## Summary
- Add complete Korean (한국어) translation for all 42 UI strings
- Add `ko` to `SUPPORTED_LANGS` in `src/i18n.js`
- Add `"ko"` to lang enum in `src/prefs.js`
- Add 한국어 radio option to language selector in `src/menu.js`

## Changed files
| File | Change |
|------|--------|
| `src/i18n.js` | Add `ko` translation object (42 keys) + update `SUPPORTED_LANGS` |
| `src/prefs.js` | Add `"ko"` to lang schema enum |
| `src/menu.js` | Add 한국어 option to Language submenu |

## Translation notes
- Reviewed by native Korean speaker and cross-checked with AI for naturalness
- Follows existing en/zh translation patterns and key structure
- Uses standard Korean software UI conventions (e.g. macOS-style "메뉴바", "방해 금지")
- Particle handling for template strings: `이(가)`, `은(는)` for version placeholders

## Test plan
- [ ] Switch language to 한국어 via tray menu → Language
- [ ] Verify all tray menu items display Korean text
- [ ] Verify right-click context menu displays Korean text
- [ ] Verify update dialog messages display correctly
- [ ] Verify session status labels (작업 중, 생각 중, etc.)
- [ ] Verify language preference persists after restart

PR #660, opened 2026-07-12
feat: account-wide Claude/Codex subscription quota — local + remote SSH, per-source

Tonic-Jin:
## What

Shows Claude Code / Codex subscription rate-limit usage (5h + weekly windows) on the pet — for this machine and for remote SSH machines — in the Dashboard quota section and as ring gauges in a Session HUD strip. Continues the remote scope from #370 as discussed there (fresh PR per that thread; #370 is closed).

## Where it shows up

- **Dashboard** — the detailed view: per-source rows with labeled bars, reset countdowns, and "as of" staleness suffixes. Always on.
- **Session HUD** (click the pet) — an at-a-glance strip above the session rows: one row per source, agent-icon pills holding two ring gauges (5h / 7d) colored by severity, dimmed to a 0-ring when a window has reset. Renders even with zero sessions, so you can check a remote's quota before starting work there. On by default, toggleable via Settings → Session HUD (`sessionHudShowQuota`).

## Mechanism — no API polling

- **Claude Code**: a statusline adapter (`hooks/claude-statusline.js`) reads `rate_limits` from statusline stdin and POSTs it as a `metadata_only` report. The OAuth usage API was deliberately rejected — it rate-limits after a couple of calls, so it can't be a live source.
- **Codex**: the CLI has no statusline, but its rollout JSONL `token_count` events carry `payload.rate_limits`; the existing local and remote log monitors extract it (field names verified against real rollout files, with a `resets_in_seconds` fallback for older CLIs).
- **Remote**: reports ride the existing SSH reverse tunnel; each carries the profile's host label.

## The three points from the #590/#370 review

1. **Session-independent**: quota lives in a per-source account store (`src/state-account-quota.js`), persisted to `~/.clawd/account-quota.json` — survives restarts and renders with zero sessions ("check the remote's quota before starting work").
2. **Per-source display by default** — one row per reporting machine. An **opt-in** `quotaMergeSources` setting collapses to one gauge set for same-subscription-everywhere setups (freshest reporter per provider wins; expired-window sources never mask live ones).
3. **Explicit mechanism**: documented in code comments, `AGENTS.md`, and the setup guide.

## Statusline chain mode (opt-in, POSIX remotes only)

Claude Code has a single statusline slot. Off by default we leave an existing statusline untouched (no quota from that machine). With the profile's "chain existing statusline" option, the user's `statusLine` object is captured **verbatim** into a sidecar; our adapter takes the slot, re-runs theirs (own process group, 3s kill cap, spawn-failure falls back to a plain render), and unregister restores the original from the sidecar.

## Hardening (from an adversarial review pass)

Store: per-bucket `capturedAt` write ordering (a second session replaying cached numbers can't overwrite newer observations), rejects already-expired / >45d-ahead `resetAt`, 12-source cap, TTL pruning (48h expired buckets, 14d unconfirmed providers). Freshness: `lastSeenAt` (last confirmation) split from `updatedAt` (last value change); identical confirmations broadcast at most once a minute. HUD: window sized from real row content; a pinned HUD ages/expires on wall clock without needing a fresh snapshot.

**Trust model**: the `host` label is client-supplied over the loopback-bound tunnel and cannot be origin-verified — same boundary as the existing session cards' host grouping (machines the user deployed hooks to). It is shape-sanitized and bounded by the caps above; documented in `src/state-account-quota.js`.

## Tested

Unit: store merge/expiry/ordering/cap, extractor vs. real rollout shapes, chain register/unregister/timeout (real process-group kill), HUD sizing math. E2E on real machines: Windows desktop + Linux remote over SSH, live Claude + Codex quota, restart persistence, profile deletion restoring the original statusline.


rullerzhou-afk:
Thank you, @Tonic-Jin — merged at last! This one had a longer road than most. Your Korean locale work in #113 became the foundation of the completed version that landed through #116; the first quota proposal in #370 helped us work out the runtime boundaries and the on-demand UX direction; and with #660 you came back, rebuilt the feature on the newer quota architecture, validated real Codex and remote behavior, and patiently worked through a long series of reviews and follow-up fixes.

The result is much stronger because you kept pushing the idea forward instead of giving up when the first shape could not land directly. We really appreciate both the engineering and the persistence. Thank you for sticking with Clawd through such a winding road — and congratulations on getting this one across the line! 🐾

PR #664, opened 2026-07-12
fix(pet): drop the pet behind text-input bubbles via native SkyLight de-delegation (#640 Phase 2)

Yike-Ye:
> Re-opened against `main` at the maintainer's request — supersedes #654, which was auto-closed when its base branch (`fix/640-pet-dodge-while-typing`) was deleted on merge.

Follows up #642 (Phase 0/1, the fade mitigation), now merged into `main`. While #642 *fades* the pet to 18% when a text field is focused, this replaces that with a **native SkyLight de-delegation** that drops the pet **genuinely behind** the bubble — fully opaque — and broadens the trigger so it steps back the moment a text-input bubble appears, not only while you are typing. The fade is kept as a fallback.

## How we validated it (before touching the integration)

The hard part of #640 was the SkyLight mechanism itself — the same one that looked "stuck" earlier (the `rc=6619136` "move failed"). So we nailed it down in isolation first, then end-to-end:

1. **Isolated SkyLight probe.** A throwaway Electron script creates two overlapping windows — a "pet" P given the real shipping `applyStationaryCollectionBehavior`, and a "bubble" B — then reads actual z-order back from `CGWindowListCopyWindowInfo`. Findings:
   - De-delegating P pulls it behind B and back on `applyStationaryCollectionBehavior`, cleanly, across repeated round-trips.
   - `rc=6619136` is a **benign** status — the calls succeed; judge by space membership / z-order, not the return code.
   - Just adding the window to the active space is **not** enough — it stays a member of the private absolute-level space until explicitly removed via `SLSRemoveWindowsFromSpaces`.

2. **Production z-order, not a rigged one.** The first probe put B at `screen-saver` level, so level dominance trivially won. We re-ran it with **B at normal level 0 and focused** — exactly what #626 leaves the editing bubble at while you type. Result: the de-delegated pet (level 0) still lands **behind** the focused bubble at the *same* level, because the bubble is the key window. So no defensive "level −1" is needed; level 0 is correct. (We tested −1 too — also works, but unnecessary.)

3. **Real machine, real elicitation.** Triggered an actual text-input elicitation bubble and confirmed, hands-on:
   - the pet drops behind the **instant the bubble appears** — before focusing anything;
   - options are clickable, the input box is clickable, **CJK input works** (candidate window no longer occluded);
   - the pet **restores to top-most the moment the bubble closes**.

4. **Unit tests.** The `#640` block (48 assertions) covers: de-delegate-on-overlap, edge-triggering, focus-independence (blur does *not* restore while the bubble is up), restore on bubble removal, the drag-lock deferral, and the fade fallback path. Full suite green (**4805 pass / 0 fail**).

## What changed

**`src/mac-window.js` — the primitive.**
`deDelegateWindowFromStationarySpace(win, level = 0)`: `SLSRemoveWindowsFromSpaces` out of the private space → land in the active user space → drop the native window level (`applyStationaryCollectionBehavior` raised it to `CGAssistiveTechHighWindowLevel`, which Electron's `setAlwaysOnTop(false)` cannot undo). Restore is just `applyStationaryCollectionBehavior` (idempotent). The two SkyLight gotchas above are documented inline.

**`src/topmost-runtime.js` — the integration.**
- On overlap, the pet's **render + hit** windows are de-delegated behind the bubble; `fadePetWindow(1)` keeps them opaque. If de-delegation is unavailable (FFI load failure → `false`), we fall back to the #642 fade.
- `reapplyMacVisibility`'s `apply()` now **respects** the de-delegated state for the pet windows — it re-de-delegates each pass instead of re-running `applyStationaryCollectionBehavior` (which would re-insert them on top).
- Only `setIgnoreMouseEvents` stays drag-deferred (Electron's mid-gesture contract); the space moves ride the same class of `setLevel`/space ops the visibility pass already runs on these windows.
- `getPetTargetOpacity` keys off the fade-fallback flag (native path stays fully opaque).

**Trigger broadened: overlap, not focus.** The dodge now fires when the pet overlaps any `__clawdMacTextInputBubble` (set at bubble creation), not only when a field is focused (`__clawdMacImeEditing`). Rationale below.

## Why the trigger is "overlap a text-input bubble", not "field focused"

#626 (which introduced the two-bubble split) deliberately keeps **text-input** bubbles *out* of the SkyLight private space so the OS IME candidate window can surface — but that same treatment means the pet (private space, assistive-tech level) sits **on top of them from the moment they appear**, covering the options and the input box before you ever click in. #626 unified both bubble kinds under one model, so to keep the effect consistent the pet must step back for the **whole** overlapping text-input bubble, not just while typing.

**Permission (options-only) bubbles are unaffected** — they get the same private-space treatment as the pet, so they coexist in one level band and were never occluded by it.

## Scope / non-goals
- No change to permission-bubble behavior, Windows paths, or the bubble's own IME-editing level handling (#626).
- The fade path is retained purely as a graceful fallback.

### DIFF
--- a/src/mac-window.js
+++ b/src/mac-window.js
@@ -91,6 +91,9 @@ function initSkyLight() {
     "void *",
     "int",
   ]);
+  // #640 Phase 2: pulling a window back OUT of the private space (de-delegation).
+  const SLSRemoveWindowsFromSpaces = lib.func("SLSRemoveWindowsFromSpaces", "int", ["int", "void *", "void *"]);
+  const SLSGetActiveSpace = lib.func("SLSGetActiveSpace", "uint64", ["int"]);
 
   const connection = SLSMainConnectionID();
   const space = SLSSpaceCreate(connection, 1, 0);
@@ -99,6 +102,8 @@ function initSkyLight() {
     connection,
     space,
     SLSSpaceAddWindowsAndRemoveFromSpaces,
+    SLSRemoveWindowsFromSpaces,
+    SLSGetActiveSpace,
   };
 
   // Same strategy as Masko Code: create an absolute-level system Space that is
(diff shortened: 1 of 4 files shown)


rullerzhou-afk:
Thanks @Yike-Ye for this contribution! We verified the native SkyLight de-delegation on a real Mac across Chinese IME, drag recovery, Spaces/fullscreen, and dual displays, and it is now merged. Really appreciate the thoughtful macOS windowing work here 🙌

PR #667, opened 2026-07-13
Hide PowerShell process-query windows in remaining integrations

### LINKED ISSUE #639: Apply -WindowStyle Hidden to the remaining PowerShell snapshot spawn sites (#628 follow-up: pi-extension, opencode-plugin, hermes-plugin)
## Context

#628 added `-WindowStyle Hidden` to the PowerShell snapshot spawn in `hooks/shared-process.js` — an in-process belt-and-suspenders for the #627 flash (Windows Terminal's default-terminal delegation can ignore `CREATE_NO_WINDOW`; the flag shortens whatever still leaks through).

That covered the shared resolver only. Three integrations carry their **own copies** of the snapshot spawn, so neither #630's per-session cache nor the planned #634 resolver-level cache will reach them — the flag is the cheap interim mitigation for all three:

- `hooks/pi-extension.ts:202`
- `hooks/opencode-plugin/index.mjs:194`
- `hooks/hermes-plugin/__init__.py` — three call sites: the snapshot at `:440`, the per-PID `Get-CimInstance` fallback at `:464`, and the `Get-Process` fallback at `:490`. All route through `_run_process_command`, which already sets `CREATE_NO_WINDOW` (`:393`) — exactly the flag WT's delegation can ignore, so the exposure is the same.

## Change

Mechanical: insert `"-WindowStyle", "Hidden"` before `"-Command"` in each arg list, same shape as #628. Note `:490` in Hermes may spawn `pwsh.exe` instead of `powershell.exe`; the parameter is valid on both.

Not gated on #634 — that issue moves the cache into `createPidResolver` for the twelve hooks that share it; these three files are independent implementations and need the flag either way.

Refs #627, #628, #630, #634.

Dxy2326:
## Summary

- add `-WindowStyle Hidden` to the Pi and shared opencode-family PowerShell process snapshots; the shared core covers both opencode and MiMo Code
- add the same flag to Hermes snapshot, per-PID CIM, and `Get-Process` fallback queries
- keep the existing `windowsHide` / `CREATE_NO_WINDOW` protection in place

## Root cause

These process-query paths do not use the shared resolver fixed in #628. Since the opencode-family refactor in `eb5fc10`, opencode and MiMo Code share the same family core and receive the mitigation together. When Windows Terminal is the default terminal, its console delegation can ignore `CREATE_NO_WINDOW`, allowing a brief PowerShell window flash.

## User impact

Background process-tree detection in Pi, opencode, MiMo Code, and Hermes now asks both Windows PowerShell and PowerShell 7 to stay hidden.

## Validation

- `node --test test/shared-process.test.js test/pi-install.test.js test/opencode-plugin-session.test.js test/hermes-plugin.test.js` (90 passed)
- `node --test test/opencode-family-bridge.test.js` (6 passed)
- `node --check hooks/opencode-family-plugin/core.mjs`
- `node --check hooks/opencode-plugin/index.mjs`
- Python compile check for `hooks/hermes-plugin/__init__.py`
- Windows PowerShell 5.1 and PowerShell 7 smoke tests with `-WindowStyle Hidden`

Fixes #639

### DIFF
--- a/hooks/hermes-plugin/__init__.py
+++ b/hooks/hermes-plugin/__init__.py
@@ -437,7 +437,7 @@ def _query_windows_process_snapshot() -> Dict[int, Dict[str, Any]]:
         "Select-Object ProcessId,ParentProcessId,Name,ExecutablePath,CommandLine | "
         "ConvertTo-Json -Compress"
     )
-    result = _run_process_command(["powershell.exe", "-NoProfile", "-NonInteractive", "-Command", script], timeout=3.0)
+    result = _run_process_command(["powershell.exe", "-NoProfile", "-NonInteractive", "-WindowStyle", "Hidden", "-Command", script], timeout=3.0)
     if not result or result.returncode != 0 or not result.stdout.strip():
         return {}
     try:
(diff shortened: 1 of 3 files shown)


rullerzhou-afk:
感谢 @Dxy2326！这次修复定位很准、范围克制，背景、影响和验证也写得非常清楚。合并前因为我们后续做了 opencode-family 重构，我这边把 opencode 那一处迁到了共享 core，最终也顺带覆盖了 MiMo Code。也想特别说一句：你提交的 PR 质量一直都很高，问题分析扎实、边界清楚，review 起来很省心。非常感谢持续贡献 🙌

PR #668, opened 2026-07-13
Validate nested hook integrations per event in Doctor

### LINKED ISSUE #599: Doctor: nested hook agents (Qoder/QoderWork/Reasonix) report green on partial installs and disabled hook groups
Found during the codex cross-review of #598 (QoderWork integration). Filing separately so it doesn't get lost — this is a pre-existing gap shared by several agents, not something #598 introduced.

## Problem

For nested-config hook agents, Doctor's `checkFileMode` only verifies that **at least one** Clawd marker command exists anywhere in the settings file (`src/doctor-detectors/agent-integrations.js`, the generic `validateCommandList(findHookCommands(...))` fallback). Two situations therefore report a healthy integration when hooks are partially or completely inoperative:

1. **Partial install** — e.g. a user hand-edits `~/.qoder/settings.json` and removes 9 of the 10 registered events. One surviving marker entry keeps Doctor green, but most animations silently stop working.
2. **Disabled hook group** — Qoder-family agents support `hooksConfig.disabled: ["clawd"]`, which turns off the whole group. Both `qoder-install.js` and `qoderwork-install.js` even normalize raw command references into the `"clawd"` id with a comment saying this is *"so Doctor can reliably see whether our group is disabled"* — but no Doctor code path for these agents ever reads `hooksConfig.disabled`. Hooks are fully off, Doctor stays green.

Deeper validation exists but only for a few agents:

- Per-event checks: `validateGeminiHookEvents` (gemini-cli), `validateQwenHookEvents` (qwen-code)
- Disabled/feature-flag supplements: `applyCodexSupplementary`, `applyGeminiSupplementary`, `applyQwenSupplementary`, `applyAntigravitySupplementary`

## Affected agents

At minimum **qoder / qoderwork / reasonix** (all `nested: true`, all walk the generic fallback, first two also carry the unconsumed `hooksConfig.disabled` mechanism). `claude-code` and `codebuddy` also resolve through the generic marker-existence path, though they have other health signals.

## Suggested direction

The descriptors already carry the data needed: qoder/qoderwork/reasonix declare `hookEvents` in `src/doctor-detectors/agent-descriptors.js`. A data-driven pass could:

1. Generalize per-event validation for any descriptor with a `hookEvents` list (report which events are missing instead of a binary marker check)
2. Generalize the `hooksConfig.disabled` / `hooksConfig.enabled === false` check for agents whose installers manage that field, mirroring the existing Gemini supplementary (warning-level, no Fix button — Clawd preserves user intent, same policy as `disableAllHooks`)

No urgency — this is diagnostic quality, not runtime behavior. Worth doing before the next hook-family agent lands so the gap stops replicating.

Dxy2326:
## Summary

- validate every required hook event for file-based integrations that declare `hookEvents`
- report partial Qoder, Reasonix, and QoderWork installs instead of treating one surviving command as healthy
- detect disabled Qoder/QoderWork hook groups and preserve that user intent without offering a misleading Fix action

## Root cause

Doctor's generic file-mode path validated all matching commands as one list and returned healthy when any command passed. It did not use the required event lists already present on the affected descriptors, and it ignored Qoder-family `hooksConfig.enabled` / `hooksConfig.disabled` settings.

## User impact

Doctor now identifies exactly which required events are missing. A complete but explicitly disabled hook group is shown as disabled rather than healthy, and Clawd does not offer an automatic repair that would contradict the user's setting.

## Validation

- `node --test test/doctor*.test.js` (166 passed)
- `git diff --check`
- existing Qwen/Gemini compatibility tests remain unchanged and passing

Fixes #599

### DIFF
--- a/src/doctor-detectors/agent-descriptors.js
+++ b/src/doctor-detectors/agent-descriptors.js
@@ -237,6 +237,7 @@ const AGENT_DESCRIPTORS = Object.freeze([
     marker: qoder.MARKER,
     nested: true,
     hookEvents: qoder.QODER_HOOK_EVENTS,
+    hookGroupId: "clawd",
   }),
   Object.freeze({
     agentId: "reasonix",
@@ -261,6 +262,7 @@ const AGENT_DESCRIPTORS = Object.freeze([
     marker: qoderwork.MARKER,
     nested: true,
     hookEvents: qoderwork.QODERWORK_HOOK_EVENTS,
+    hookGroupId: "clawd",
   }),
 ]);
 
--- a/src/doctor-detectors/agent-integrations.js
+++ b/src/doctor-detectors/agent-integrations.js
@@ -12,7 +12,6 @@ const { getAgent } = require("../../agents/registry");
 const { commandMatchesMarker, findHookCommands } = require("../../hooks/json-utils");
 const { GEMINI_HOOK_EVENTS } = require("../../hooks/gemini-install");
 const { ANTIGRAVITY_HOOK_EVENTS, HOOK_GROUP_ID: ANTIGRAVITY_HOOK_GROUP_ID } = require("../../hooks/antigravity-install");
-const { QWEN_CODE_HOOK_EVENTS } = require("../../hooks/qwen-code-install");
 const {
   hasUserPermissionHookInOtherFiles,
   hasUserPermissionHookInSettingsJson,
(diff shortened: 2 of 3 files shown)


rullerzhou-afk:
感谢你补上这块 Doctor 校验！我们已经在当前 main 的合并树上完成复核：Doctor 测试 191/191 通过，并在 Windows 真机 Qoder 配置上验证了完整安装、缺失单事件、hooksConfig.disabled=["clawd"] 和 hooksConfig.enabled=false 四种场景；诊断状态、缺失事件以及 Fix 按钮策略都符合 #599 的预期，配置也已按字节恢复。实现很扎实，辛苦了 🙌

PR #669, opened 2026-07-13
Count concurrent Codex Desktop sessions separately

### LINKED ISSUE #581: Codex Desktop HUD can under-count concurrent local sessions sharing one agentPid
### Problem

When using Codex Desktop with multiple concurrent local work sessions, Clawd's HUD / work count can show only one active Codex session even when several Codex work sessions are open.

This may be expected after #427, but it makes the HUD less useful for people using Codex Desktop as a multi-workspace / multi-thread monitor.

### Local environment

- Clawd on Desk: 0.11.0
- macOS: 26.1
- Codex CLI: 0.142.5
- Codex hooks: enabled (`[features].hooks = true`)
- Clawd-managed Codex hooks installed in `~/.codex/hooks.json`
- Clawd-managed Claude Code hooks also installed, but the observed issue is about Codex HUD counting

### Observed behavior

- Several Codex Desktop work sessions can be active locally.
- Clawd's HUD / desktop work count shows only one Codex work item.
- The user also uses multiple API providers / custom provider configs, and initially suspected that sessions falling into a `custom` bucket might make them indistinguishable.

### What I found in the current source

The current local Codex HUD dedupe appears to key only on `agentPid`:

Evidence screenshot

Relevant logic:

- `src/state-session-dedupe.js`
  - `getLocalCodexProcessKey(session)` returns `codex-agent:${agentPid}`.
  - `buildLatestLocalCodexProcessIds()` keeps only the latest session for each such key.
- `src/state-session-snapshot.js`
  - superseded local Codex process sessions are marked `hiddenFromHud`.
  - `hudEntries` excludes `hiddenFromHud` entries.
- `src/state-priority.js`
  - superseded local Codex process sessions are skipped for display-state selection.

So this does not look like `provider` / `custom` directly controls the dedupe key. It looks more like multiple Codex Desktop sessions can share the same local `agentPid`, and then only the newest one is counted in HUD/display state.

### Expected behavior / feature request

Could Clawd expose an option for Codex Desktop users to choose the HUD aggregation strategy?

Possible options:

- Default/current: collapse local Codex sessions by `agentPid` to avoid same-process session ghosts.
- Optional: count/display local Codex sessions by `session_id` / `transcript_path`, even when they share the same `agentPid`.
- Optional hybrid: only collapse same-`agentPid` sessions after they become stale/idle, while keeping currently active working/thinking sessions visible.

I understand #427 fixed real same-process ghost sessions, so this probably should not replace the current default behavior. A user-facing setting or a more conservative active-session exception may be safer.

### Why this matters

For users running several Codex Desktop work sessions in parallel, the desktop HUD is useful as a lightweight monitor. If all local Codex Desktop sessions share the same process identity and only one counts, the HUD under-reports active work.

Dxy2326:
## Summary

- keep independent Codex Desktop threads separate in HUD/work count even when they share one `agentPid`
- retain existing PID-based ghost dedupe for CLI/unknown local Codex sessions
- cover both dominant display-state selection and HUD visibility

## Root cause

Codex Desktop multiplexes multiple stable thread/session IDs through one process. The shared dedupe keyed every local Codex session by `agentPid`, so only the newest Desktop thread remained visible.

## User impact

Concurrent Codex Desktop threads now contribute independently to the HUD and display-state selection. Sessions without a confirmed Desktop originator continue using the existing conservative PID dedupe.

## Validation

- `node --test test/state-priority.test.js test/state-session-snapshot.test.js test/state.test.js` (282 passed)
- `git diff --check`
- existing #427 generic/local same-PID regression remains passing

Fixes #581

### DIFF
--- a/src/state-session-dedupe.js
+++ b/src/state-session-dedupe.js
@@ -14,6 +14,10 @@ function normalizePositiveInteger(value) {
 
 function getLocalCodexProcessKey(session) {
   if (!session || session.agentId !== "codex" || session.host || session.headless) return null;
+  // Codex Desktop multiplexes independent threads through one process, so its
+  // stable session IDs must remain distinct in the HUD (#581).
+  if (typeof session.codexOriginator === "string"
+    && session.codexOriginator.trim().toLowerCase() === "codex desktop") return null;
   const agentPid = normalizePositiveInteger(session.agentPid);
   return agentPid ? `codex-agent:${agentPid}` : null;
 }
--- a/test/state-priority.test.js
+++ b/test/state-priority.test.js
@@ -102,6 +102,23 @@ describe("state-priority display selection", () => {
     ])), "working");
   });
 
+  it("keeps concurrent Codex Desktop sessions separate when they share one agent process", () => {
+    assert.strictEqual(resolveDominantSessionState(new Map([
+      ["codex:desktop-a", session("working", {
+        agentId: "codex",
+        agentPid: 4242,
+        codexOriginator: " Codex Desktop ",
+        updatedAt: 1000,
+      })],
+      ["codex:desktop-b", session("thinking", {
+        agentId: "codex",
+        agentPid: 4242,
+        codexOriginator: "codex desktop",
+        updatedAt: 2000,
+      })],
+    ])), "working");
+  });
+
   it("does not dedupe Codex sessions without a local agent pid", () => {
     assert.strictEqual(resolveDominantSessionState(new Map([
       ["codex:old", session("working", {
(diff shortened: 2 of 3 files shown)


rullerzhou-afk:
Thanks for flagging this — confirmed it is a real bug. Current Codex Desktop metadata reports `codex_work_desktop`, so the original `codex desktop` check does not match.

We handled the follow-up on our side in `fc00a06`: the Desktop originator check is now shared by session dedupe, hook PID selection, and thread focus, with bounded cleanup for stale Desktop Done sessions. The targeted suite is green, and we also verified the macOS package/runtime path.

No action needed from you; we will fold this in when landing #669. Thanks again for the contribution!

Issue #634, opened 2026-07-09
#627 follow-up: share the per-session PID-snapshot cache across all agent hooks (createPidResolver), not just Claude


rullerzhou-afk:
## Background

#627: on Windows, every hook event cold-starts PowerShell to snapshot the process tree (`createPidResolver` → `getWindowsProcessSnapshot` in `hooks/shared-process.js`). With **Windows Terminal as the default terminal application**, that spawn flashes a visible console window despite `windowsHide:true` (WT's console-host delegation doesn't honor `CREATE_NO_WINDOW`).

#630 fixes it **for Claude only**: the process tree is stable within a session, so it snapshots once (SessionStart / UserPromptSubmit) into a per-session cross-process cache (`hooks/pid-cache.js`) with a bounded sliding TTL, and high-frequency events (PreToolUse/PostToolUse/…) read the cache instead of spawning.

## Problem

The cross-process cache is wired **only into `clawd-hook.js`** (Claude), in its `buildStateBody` layer. The other **12 agent hooks all share the same `createPidResolver`** but none of them `require` `pid-cache`.

`createPidResolver`'s internal `_cached` (shared-process.js:239) is **per-process only** — each hook event is a fresh `node` process, so `_cached` is always `null` and every event still spawns one `powershell.exe`. On Windows + WT those flash exactly like Claude used to.

Affected hooks (share `createPidResolver`, no cross-process cache today):
`cursor`, `copilot`, `codex`, `gemini`, `kimi`, `qwen-code`, `qoder`, `qoderwork`, `kiro`, `codebuddy`, `antigravity`, `reasonix`.

## Proposal

Move the "per-session cross-process snapshot cache + sliding-TTL renew on hit" capability out of `clawd-hook.js` and **down into the shared `createPidResolver`**, so all agent hooks benefit from one place instead of each re-implementing it.

## Why this is not a mechanical move (prerequisites)

Each agent differs, so the cache can't just be dropped in:

1. **Session identity** — cache key is `session_id + cwd`. Each agent passes a different session shape; some may lack a stable `session_id` and must **disable caching** (fall back to per-event snapshot) rather than cross sessions — cf. #583 (`sid=default`).
2. **Refresh timing** — Claude re-snapshots on `SessionStart`/`UserPromptSubmit` (`wantsFresh`). Other agents have different event models (e.g. qwen's three hooks don't map 1:1 to Claude's), so each needs its own "when to re-snapshot" point chosen.
[...]

PR #674, opened 2026-07-14
feat(hooks): sink pid cache + lifecycle into the shared resolver (#634 Slice 1)

rullerzhou-afk:
Slice 1 of #634: move the per-session pid-cache orchestration out of the Claude hook and into the shared resolver, so the other 12 agent hooks can adopt it in later slices. Sequential follow-up to #672 (now on `main`).

## What this changes

Only Claude is migrated here — the other 12 adapters keep calling the no-arg `resolve()` unchanged.

**`hooks/shared-process.js`** — `createPidResolver` gains a lifecycle context overload:

- `resolve({ namespace, sessionId, cacheCwd, lifecycle, cacheable })` with `start` / `prompt` / `event` / `end` lifecycles.
- The no-arg `resolve()` stays byte-for-byte with the #672 baseline (same 10-field shape, SessionStart prewarm reuse, zero cache interaction) — verified with an A/B against the baseline source. `pid-cache` is lazy-required only in the context path, so there is no require cycle and unmigrated adapters never load it.
- Windows zero-spawn contracts: `prompt`/`end` never spawn (even for a non-cacheable session); `event` hit is zero spawn, miss is one fresh; `start` reuses the prewarm. macOS/Linux keep the fresh-every-event runtime behavior.
- Claude v1→v2 cache promotion, so a session already running across the upgrade does not take a one-off re-resolve (flash). The promoted content and its delete-guard come from a single file observation, and the write is no-clobber, so a concurrently-written fresh entry is never overwritten or deleted.

**`hooks/pid-cache.js`** — cache v2 alongside the untouched v1:

- Namespaced + versioned, with a non-overlapping prefix (`clawd-pidcache2-`) so a dual-prefix sweep classifies every file unambiguously and drops only on age-floor + death-proof.
- Reads consult no clock (lease model); liveness is a double-PID check by the caller.
- `readPidCacheEntry()` returns the subset and its file identity from one fd observation (so a promotion never promotes one version and deletes another).

**`hooks/clawd-hook.js`** — `buildStateBody` is now a thin Claude adapter: event→lifecycle map (`Stop` is not `end`), cache identity, remote bypass before the resolver, and apply the returned metadata (an empty result ships no degraded pid).

## Compatibility / safety

- The 12 unmigrated adapters (Codex, Cursor, Gemini, Copilot, Kimi, Kiro, Qoder, QoderWork, Qwen, Reasonix, CodeBuddy, Antigravity) are untouched: same fields, same spawn count, zero cache interaction — regression-tested.
- No public protocol or event-name changes. No permission-behavior changes.

## Testing

`node --test test/*.test.js` — all green. Coverage:

- Full lifecycle matrix (start / prompt / event / end, cacheable and not, Windows and non-Windows).
- v1→v2 promotion: success, write-failure keeps v1, dead PID, concurrent replacement, and end/drop interleave.
- The Slice-1 no-arg compatibility red line: field set + first/second spawn count + zero cache interaction, plus an unmigrated-adapter regression.
- The Claude end-to-end regression through the real resolver.

Two review rounds surfaced and fixed: a v1 drop on a failed start/event write, a missing event-population sweep entry point for no-start adapters, cache-test directory isolation (so the sweep tests never scan the real temp dir), and a promotion identity race (the promoted bytes and the delete-guard must come from the same read).

## Follow-ups

- Real-machine 4688/ETW spawn audit under Windows Terminal as the default terminal app — pending; not reproducible in a VS Code terminal.
- Later slices migrate the remaining adapters one at a time.

### DIFF
--- a/hooks/clawd-hook.js
+++ b/hooks/clawd-hook.js
@@ -8,10 +8,10 @@ const fs = require("fs");
 const { postStateToRunningServer, readHostPrefix, resolveWslDistro } = require("./server-config");
 const { fitStateBodyToByteBudget } = require("./state-payload-size");
 const { extractClaudeContextUsageFromEntries } = require("./context-usage");
-const { createPidResolver, readStdinJsonDetailed, getPlatformConfig, tmuxSocketFromEnv, processAlive } = require("./shared-process");
-const pidCache = require("./pid-cache");
-
-const isWin = process.platform === "win32";
+const { createPidResolver, readStdinJsonDetailed, getPlatformConfig } = require("./shared-process");
+// #634: the pid cache + lifecycle orchestration is owned by the shared resolver
+// now (hooks/shared-process.js); this adapter no longer touches pid-cache,
+// processAlive, or isWin directly.
 
 const TRANSCRIPT_TAIL_BYTES = 262144; // 256 KB
 // #583: claude-code registers this hook with async:true and a 5s timeout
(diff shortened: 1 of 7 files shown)


rullerzhou-afk:
Superseded by #687, which includes this PR's commit and carries the #681 Phase A fixes on top. Closing to avoid maintaining and merging two overlapping PRs.

PR #504, opened 2026-06-15
feat(kimi): track the new TypeScript Kimi Code CLI at ~/.kimi-code (#478)

KaiC5504:
Closes #478.

  The current integration only tracks the old Python `kimi` CLI (`~/.kimi/config.toml`) and only detects the `kimi` binary, so a user who has
  only the new TypeScript **Kimi Code** CLI installed isn't detected and its hooks are never registered.

  This extends the **existing** `kimi-cli` integration to be dual-path — per the maintainer's guidance to 照抄现有的三件套, it reuses the
  existing `agents/kimi-cli.js` + `hooks/kimi-hook.js` + `hooks/kimi-install.js` trio rather than adding a separate agent:

  - Hooks are written to / detected in / cleaned from **both** `~/.kimi/config.toml` (old Python `kimi`) **and** `~/.kimi-code/config.toml` (new
  TypeScript `kimi-code`, overridable via `$KIMI_CODE_HOME`) — whichever directory exists; each path is skipped when its parent dir is absent.
  - Install detection, the Doctor connection check, startup auto-sync, `npm run install:kimi-hooks`, and uninstall all cover both paths.
  - The agent id stays `kimi-cli`; no new agent, and no behavior change for old-CLI-only users (zero regression).
  - The new CLI ships the **same** binary name (`kimi` / `kimi.exe`), just under `~/.kimi-code/`, so process-name detection needed no change.

  ### Verified live (Windows, subscribed Kimi Code account)
  - Detection shows Kimi installed; hooks register into `~/.kimi-code/config.toml` (providers/models/oauth config untouched).
  - Doctor shows connected; the pet animates through states (thinking → working → attention); session-end and restart auto-sync are idempotent.
  - Uninstall cleanly removes the hooks while leaving auth/config intact; reinstall restores them.
  - Full unit suite green (`npm test`).

  ### Out of scope (follow-ups)
  Native permission events (`PermissionRequest` / `PermissionResult` for clickable approve/deny, and `Interrupt` mapping) — note the old Kimi
  integration was also passive-only here — plus a permission cue for the new CLI's Claude-style tool names (`Write` / `Bash` / `Edit`).

--------------------

  当前集成只跟踪旧版 Python `kimi`（`~/.kimi/config.toml`），且只检测 `kimi` 进程，所以只安装了新版 TypeScript **Kimi Code**
  的用户既不会被识别为已安装，hooks 也不会被注册。

[...]

PR #675, opened 2026-07-14
feat(kimi): tool-aware cues on the passive Kimi Code permission bubble

KaiC5504:
Follow-up invited in the #504 closing comment: the permission cue for Kimi Code's
Claude-style tool names (Write / Bash / Edit), building on the native
PermissionRequest events from #593.

## What it does

The passive Kimi permission card previously only said "check the Kimi terminal".
The hook now forwards a whitelisted, clamped subset of the native
PermissionRequest's structured `tool_input` (`command` / `file_path` / `pattern`),
and the bubble reuses the standard cue path Claude bubbles use: real tool pill,
`formatDetail` detail line, irreversible-action badge. The card stays
dismiss-only — approval semantics are untouched.

When no structured input arrives (legacy Python CLI, shape drift), the card
renders byte-identical to the old generic one.

## Validation

- Double validation: the hook whitelists + clamps before POSTing, and the server
  re-runs the same validator (`extractPermissionToolInput`) at the trust
  boundary — the `normalizeQuotaGroup` idiom.
- Heavy fields (`content`, `old_string`, `new_string`) are dropped by
  construction; `action` / `display.command` / `tool_name` are clamped so a
  heredoc-sized command can't 413 the /state POST and silently lose the
  notification.
- Bash's model-authored `description` is deliberately not forwarded:
  `formatDetail` prefers it over `command`, so generated text could mask what
  actually runs.
- Kimi's `path` maps to Claude's `file_path` (verified unchanged in the
  kimi-code 0.23.6 bundle).
- A repeat request for the same session refreshes the card in place (the Codex
  notify idiom) instead of leaving a stale cue for a command the terminal is no
  longer blocking on.
- Rendering is textContent/setAttribute only; remote surfaces (Telegram/mobile)
  are excluded.

## Deliberate trade-offs (open to changing)

- The detail line truncates at 120 chars (Claude-bubble parity) where the old
  card showed up to 500 raw chars, and the irreversible badge scans the 500-char
  clamped command (the Claude path scans 4096). Both are wire-budget choices —
  happy to raise the clamps if you prefer.
- The rich pill makes the passive card resemble an actionable bubble; the header
  title and the single "Got it" button remain the passive markers.
- The MCP relabel on the pill (`parseMcpToolName`) is a small separable hunk —
  say the word if you'd like it dropped for tighter scope.
- Legacy immediate-mode requests (lowercase `shell`) get an unstyled pill —
  accepted as cosmetic on a legacy-only path.

## Testing

- Full suite green: 4901 tests / 4885 pass / 0 fail / 16 skipped.
- Every hop is behaviorally tested: hook extraction (incl. clamp, trim-order,
  garbage-shape and description negatives), server re-validation, state
  threading, permission entry + renderer payload, refresh-in-place at both
  layers; the renderer branch follows the repo's source-string test convention.
- Live-verified against kimi-code 0.23.6 interactively (Bash / Write / Edit
  approvals: pill, detail line and badge render; PermissionResult
  auto-dismisses; legacy-shaped payloads keep the generic card), plus payload
  shapes verified from the npm bundle source.

### DIFF
--- a/hooks/kimi-hook.js
+++ b/hooks/kimi-hook.js
@@ -177,6 +177,39 @@ function readToolName(payload) {
   return "";
 }
 
+const PERMISSION_TOOL_INPUT_MAX_CHARS = 500;
+
+// Whitelisted subset of a native PermissionRequest's structured tool_input,
+// forwarded so the bubble can render tool-aware cues (command for Bash, file
+// path for Write/Edit) through the same formatter Claude bubbles use. Kimi
+// Code's file tools say `path` where Claude says `file_path` (captured from
+// kimi-code 0.14.3, unchanged in the 0.23.6 bundle's Write/Edit schemas);
+// map it so the formatter finds it. src/server-route-state.js re-runs this
+// exact function at the trust boundary rather than trusting the hook.
+// Everything else is dropped by construction: content / old_string /
+// new_string can be arbitrarily large (an oversized /state body is rejected
+// whole), and Bash's optional `description` is model-authored text that
+// formatDetail would show instead of the real command — the card must show
+// what actually runs, not what the model says about it.
+function extractPermissionToolInput(raw) {
+  if (!raw || typeof raw !== "object" || Array.isArray(raw)) return null;
+  const out = {};
+  const take = (target, value) => {
+    // First writer wins: an explicit file_path beats the mapped `path` alias.
+    if (typeof value !== "string" || out[target] !== undefined) return;
+    // Trim first, matching the server's trim-then-clamp order: a value with
+    // ≥500 chars of leading whitespace would otherwise clamp to whitespace
+    // only and be dropped server-side.
+    const trimmed = value.trim();
+    if (trimmed) out[target] = trimmed.slice(0, PERMISSION_TOOL_INPUT_MAX_CHARS);
+  };
+  take("command", raw.command);
+  take("file_path", raw.file_path);
+  take("file_path", raw.path);
+  take("pattern", raw.pattern);
+  return Object.keys(out).length ? out : null;
+}
+
 function isExplicitPermissionSignal(payload) {
   if (!payload || typeof payload !== "object") return false;
   const topLevelFlags = [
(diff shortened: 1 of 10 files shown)


rullerzhou-afk (approved):
Tested locally with Kimi Code 0.22.0.

- The real PermissionRequest used `Bash` with `tool_call_id`, structured `tool_input`, and `display.command`.
- The rich passive cue rendered the correct tool and command and remained dismiss-only.
- Clicking "Got it" did not cause the cue to reopen while the native approval was pending.
- Rejecting the request produced the expected PermissionResult and cleared the lifecycle normally.
- The targeted and full automated test runs passed locally.

LGTM.

Non-blocking UX note: "Got it" currently dismisses the cue but does not focus the corresponding terminal.

rullerzhou-afk:
Thank you @KaiC5504 for the contribution! We verified the change with Kimi Code 0.22.0 on a real permission flow, and the tool-aware passive cue behaved correctly. Really appreciate the thoughtful implementation and test coverage. 🙌

rullerzhou-afk:
Hey @KaiC5504 — while dogfooding the passive cue this week I dug into how it behaves on the Python kimi-cli and found two gaps (full details in #702):

1. On kimi-cli 1.37 (and current upstream 1.49, per source) the `PreToolUse` payload carries no explicit permission fields, so in the default explicit-only mode the passive cue never fires at all — it only works with the suspect heuristic opted in (`CLAWD_KIMI_PERMISSION_SUSPECT=1` or the mode switch).
2. When one assistant message queues several gated tool calls, kimi-cli fires all their `PreToolUse` events up front (~0.1s apart) and the cue only surfaced for the first approval. Just fixed in #702.

A question to calibrate the follow-up work (making the suspect heuristic the default for the legacy flavor, plus a Doctor hint): **when you live-verified the cue for this PR, which CLI were you on — the npm `kimi` (Kimi Code) or the Python `kimi-cli` — and which version?** I'm trying to figure out whether any kimi-cli version actually emits explicit permission fields on `PreToolUse`, or whether your verification ran through the native `PermissionRequest` path on Kimi Code. The answer decides how I scope the default-mode change and word the Doctor message.

No rush — and thanks again for this PR and #680; the tool-aware cue made this week's debugging noticeably nicer.

rullerzhou-afk:
Quick update: both pieces landed on main today — your #680 (Got-it now focuses the terminal) and #702 (the cue re-arms for each queued approval in a batch). Together they close the loop nicely: every pending approval gets its own card, and every card's "Got it" takes you straight to the terminal that's waiting. Verified the composed behavior live on kimi-cli 1.37 with batched tool calls.

Both will ship with the next release. The version question above is still open whenever you get a moment — no rush, it only affects how I word the default-mode change in the follow-up.

KaiC5504:
Thanks for merging #680, and for the quick #702 fix — great to hear the composed flow works well.

To answer your question: my verification for this PR was on the **npm Kimi Code, 0.23.6**, in an interactive session — so it went through the native `PermissionRequest` path, not `PreToolUse`. I've never run the Python kimi-cli, so I can't tell you whether any version of it emits explicit permission fields; the only source I've read is the kimi-code 0.23.6 npm bundle (`dist/main.mjs`), not the Python CLI.

Given that, making the suspect heuristic the default for the legacy kimi-cli flavor (plus a Doctor hint) sounds right to me.

PR #679, opened 2026-07-14
feat(settings): let users pick the default idle animation

### LINKED ISSUE #509: Feature Request: Add default animation switching option
### Description
Currently, when switching to a different animation, the pet reverts back to the default idle animation after a short period. There is no option in Settings to change the default/idle animation.

### Proposed Solution
Add a **Default Animation** setting in Settings that allows users to choose which animation the pet uses as its idle/default state, instead of always reverting to the same one.

### Why This Matters
- Users want to personalize their desktop pet experience
- The 12 animation states are great, but most of them can only be seen briefly before reverting to idle
- A simple dropdown in Settings → Default Animation would significantly improve user satisfaction

### Current Behavior
1. Right-click tray icon → change animation
2. Animation plays for a short time
3. Pet reverts back to original idle animation

### Expected Behavior
1. Settings → Default Animation → select preferred animation
2. Pet uses the selected animation as its persistent idle state

### Environment
- Version: v0.10.0
- OS: Windows 11

Thank you for this wonderful desktop pet app! 🦀

KaiC5504:
Closes #509

Adds a per-theme **Default idle animation** picker to Settings → Animation & Sound → Idle, following the spec in the issue comments:

- Only the **visual** swaps — the logical state stays `idle`. Thinking/working/permission/completion/sleep override exactly as before, and one-shot/idle-pool animations return to the selected visual instead of the theme default.
- Options = theme default + the theme's idle variants (`states.idle` + `idleAnimations`), resolved against the loaded active theme.
- **Unset = unchanged**: with no selection the choice resolves to `null` and every call site keeps its existing behavior, byte-identical.
- Eye tracking (and spin-to-dizzy) stay exclusive to the true follow sprite by design — the pet doesn't track eyes while resting on a non-follow visual.
- Applies live while idle; stored per-theme (`idleVisual: {themeId: file}`), pruned on theme removal; a file dropped by a theme update degrades gracefully back to the theme default.
- i18n: en / zh / zh-TW / ko / ja.

## Implementation notes

- New pure module `src/idle-visual.js` is the single source of truth (option listing, choice resolution, labels); runtime call sites resolve through one `getIdleVisualChoice()` on the live active theme, so theme switches need no refresh choreography.
- A dedicated pref rather than a `themeOverrides.states.idle` patch: patching `states.idle[0]` would redefine the follow sprite (breaking the eye-tracking gate) and tie the choice to "reset all overrides".
- The picker reuses the existing language-picker listbox pattern and updates in place on settings broadcasts (no re-render flicker).

## Testing

- Full suite green: 4932 tests, 0 fail.
- New coverage: resolver unit tests, state/tick behavior (reverts, idle-pool exclusion, eye-tracking gate), renderer first frame, `setIdleVisual` command validation, picker UI (render / optimistic select / patch-in-place / listener cleanup), and main-process wiring.
- Manually verified: picking a variant swaps the pet live while idle; reactions, agent states, sleep/wake, and idle-pool plays all return to the selection; "Theme default" restores follow-sprite behavior + eye tracking; choice persists across restarts and is remembered per theme.


src/renderer.js
```diff
@@ -48,7 +48,8 @@ function initWithConfig(cfg) {
   _fileViewBoxes = tc.fileViewBoxes || {};
   _dragSvg = tc.dragSvg || null;
   _dragSvgs = tc.dragSvgs || {};
-  _idleFollowSvg = tc.idleFollowSvg || "clawd-idle-follow.svg";
+  // Pre-IPC first frame rests on the user-selected idle visual when one is set.
+  _initialIdleSvg = tc.idleDefaultVisual || tc.idleFollowSvg || "clawd-idle-follow.svg";
```
rullerzhou-afk: We still need to retain the theme's actual follow-idle file separately (for example, _idleFollowSvg = tc.idleFollowSvg) instead of replacing that identity with only _initialIdleSvg. needsEyeTracking(state) below still treats every logical idle state as trackable, so selecting clawd-idle-reading.svg, clawd-idle-look.svg, or clawd-idle-bubble.svg makes attachEyeTracking() retry for roughly a second and then log "Timed out waiting for SVG eye targets"; those files do not contain #eyes-js. Meanwhile, tick.js correctly stops sending eye movement unless the resolved file is the real follow-idle SVG, so the two sides now disagree. Could we make idle eye tracking file-aware (file === idleFollowSvg), including the object-channel and wake-recovery paths, and test that non-follow idle visuals never require eye targets?

rullerzhou-afk (changes requested):
Thanks for implementing this — the overall direction looks good and matches the intended scope of #509.

I confirmed that this is an idle visual picker rather than an arbitrary state/animation picker: choices are limited to the theme default plus supported idle variants, while thinking, working, permission, completion, sleep, and other logical states continue to override it normally. The per-theme persistence and invalid-file fallback also look correct.

I found one renderer issue that I think should be fixed before merging; I left the details inline. In short, tick.js gates eye movement by the actual follow-idle file, but renderer.js still attaches eye tracking based only on the logical idle state. That causes non-follow idle visuals to search for eye targets they do not have, affects wake recovery, and can leave stale eye offsets in compatible third-party SVGs.

Could we make the renderer decision file-aware and add coverage for follow idle, non-follow idle, and wake recovery? Other than this issue, the implementation and feature scope look solid. I also tested the PR merged onto the latest main; the related tests passed.

src/renderer.js
KaiC5504: Fixed in def56fd. The renderer now keeps the follow-file identity separate again (`_idleFollowSvg`), with `_initialIdleSvg` only seeding the pre-IPC first frame, and a new `tracksEyesForFile(state, file)` requires `file === idleFollowSvg` for idle at all four attach decision points: swap completion, same-file re-entry, the wake-recovery `needsEyes` check, and the eye-move re-attach fallback.

Coverage added in the renderer harness: the follow sprite still attaches; a non-follow idle visual never attaches even when it exposes eye targets (covers both the retry-timeout and the stale-offset case); mini-idle stays un-gated; and a low-power wake while resting on a non-follow visual skips the eye-object reload and reports `resumed` instead of `error`.

One scoping note: I kept `needsObjectChannel` state-based, so non-follow idle visuals still render on the `<object>` channel. Making the channel file-aware as well would move them to ``, which changes their sizing path (per-channel `objectScale`/`fileScales`) and loses the low-power SVG pause — that needs a reachable SVG document. If you intended the channel decision to be file-aware too, I can follow up.

Full suite locally: 4936 tests, 0 failures.

rullerzhou-afk (approved):
Re-reviewed the fixed head (def56fde) and completed isolated macOS smoke testing. The idle visual selection, task override/return, persistence, and eye-tracking detach/restore paths all behave as intended. No remaining blockers.

PR #680, opened 2026-07-14
feat(kimi): focus the terminal when the passive Kimi cue is dismissed

KaiC5504:
Follow-up to #675, implementing the non-blocking UX note from your review:

> "Got it" currently dismisses the cue but does not focus the corresponding terminal.

### What changed
Clicking the passive Kimi Code cue's **"Got it"** button now brings the originating
terminal forward in addition to dismissing the bubble — so you land where Kimi Code's
native approve/reject prompt is waiting instead of hunting for the terminal.

The dismiss path in `handleDecide` routes through the existing
`ctx.focusTerminalForSession(sessionId, { fallbackEntry })` — the same helper the
interactive "Go to Terminal" / `deny-and-focus` buttons already use. No renderer change:
the cue's button already sends `decide`, and the main process owns the Kimi-notify dismiss.

### Scope
Kimi-only. Codex's passive notify is informational-only and has no native prompt to
return to, so it stays a plain acknowledge — pinned by a regression test so the shared
passive-notify branch can't silently gain the behavior.

### Testing
- New tests in `test/permission-codex-response.test.js`: (1) the Kimi cue's "Got it"
  focuses the terminal; (2) Codex notify dismiss does not focus. The feature test was
  confirmed to fail before the change.
- Full suite green: 4997 tests, 4981 pass, 0 fail, 16 skipped.
- Live-verified on a real Kimi Code session: clicking "Got it" brings the Kimi terminal
  to the foreground.

### DIFF
--- a/src/permission.js
+++ b/src/permission.js
@@ -1929,6 +1929,13 @@ function handleDecide(event, behavior) {
   if (!perm) return;
   if (perm.isCodexNotify || perm.isKimiNotify) {
     dismissPassiveNotify(perm, "ipc-decide");
+    // Kimi Code's cue is a heads-up that its terminal is blocking on a native
+    // approve/reject prompt, so "Got it" doubles as "take me there": focus the
+    // originating terminal after dismissing. Codex's passive notify is
+    // informational-only, so it stays a plain acknowledge.
+    if (perm.isKimiNotify) {
+      ctx.focusTerminalForSession(perm.sessionId, { fallbackEntry: buildPermissionFocusEntry(perm) });
+    }
     return;
   }
   if (perm.isCodex) {
--- a/test/permission-codex-response.test.js
+++ b/test/permission-codex-response.test.js
@@ -270,6 +270,64 @@ describe("Codex permission response sanitizer", () => {
     assert.strictEqual(api.pendingPermissions.length, 0);
   });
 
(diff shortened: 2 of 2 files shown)


rullerzhou-afk (approved):
Reviewed the diff plus the passive-notify branch in `handleDecide`, and ran the tests locally — approving.

What I checked:

- The dismiss path routes through the same `focusTerminalForSession(sessionId, { fallbackEntry })` helper the interactive deny-and-focus buttons already use, and the change is correctly scoped to `isKimiNotify` — the Codex passive notify keeps its acknowledge-only behavior, pinned by your regression test.
- The test fixtures match what `showKimiNotifyBubble` actually constructs (no fields a real Kimi cue entry wouldn't have), so the assertions exercise the real entry shape rather than an idealized one.
- Full suite green on my machine as well.

One non-blocking note for a future pass: the passive branch dismisses (and now focuses) regardless of the `behavior` value it receives, and the tests only pin the `"allow"` path. Today the cue has a single "Got it" button so this is fine; if the card ever grows a second button sending a different value, nothing will flag that it also triggers the focus. Not worth blocking over — just leaving a marker.

This also composes nicely with the batched-approvals fix I'm about to put up: the cue will re-arm for each queued approval, and with this change every re-armed card's "Got it" brings the terminal forward too. Thanks for the careful scoping and the regression pin!

Issue #662, opened 2026-07-12
Feature request: explain why a session can't be focused, and offer an alternative action


seulnan:
## The gap I hit

When I click a session in the Session HUD, I expect it to bring that terminal to the front — that's the mental model the row sets up. But some of my sessions do nothing at all when I click them: no jump, no message, no visible reaction. As a user I can't tell whether I misclicked, whether it's broken, or whether it's simply not possible — I just keep clicking and nothing happens.

The only signal today is the crossed-out icon's hover tooltip (added in `7a2f251`), and it's easy to never notice — you have to hover that small marker to learn anything.

## What's happening today

For a non-focusable session (`canFocus === false` — remote / `webui`, or a missing `sourcePid`):

- HUD (`session-hud-renderer.js`): the click handler runs `if (canFocus) window.sessionHudAPI.focusSession(...)`. When `canFocus` is false the click still clears unread + acks completion, but there's **no user-visible feedback** — it reads as a dead click.
- Dashboard (`dashboard-renderer.js:488`): the Jump button is simply `disabled` when `canFocus !== true`.

So the "why" only lives in a hover tooltip, and the click itself is silent.

## What I'd like

Two small, safe additions — I'd stop short of changing focus behavior itself (see below):

1. **Say why on click.** When a non-focusable row is clicked, surface a short message (toast / inline) that this session can't be brought to the front, and why — remote session vs. missing terminal window info. Turn the silent click into an explanation.
2. **Offer an alternative action.** These sessions still carry `cwd` in the snapshot (`state-session-snapshot.js`), and the app already uses `shell.openPath(dir)` (`codex-pet-main.js:564`, `doctor-logs.js:98`). So an **"Open folder"** action (and/or "Copy path") is feasible without new plumbing — giving the user something useful instead of a dead end.

Rough shape I have in mind, so we're talking about the same thing:
- For `canFocus === false` rows, replace the silent no-op with a brief inline notice + an "Open folder" affordance driven by the session's `cwd`.
- Add the new strings next to the existing `sessionHudFocusUnavailableTooltip` key (en / zh / zh-TW / ko / ja).

## Why open an issue first

[...]

PR #684, opened 2026-07-15
fix(hud): render real tooltip text instead of raw i18n keys

KaiC5504:
The Session HUD renders four tooltips as their raw i18n key names instead of the
translated text: `sessionHudPinTooltip`, `sessionHudUnpinTooltip`,
`sessionHudFocusUnavailableTooltip`, `sessionHudRemoteFocusUnavailableTooltip`.
The pin button is on every HUD, so every user hits this one.

## What it does

Moves the four `sessionHud*` tooltip strings from `src/settings-i18n.js` into
`src/i18n.js` across all five locales. The translations already existed and are
moved byte-identical — this only changes which dictionary owns them.

## Why it happened

`settings.html` is the only window that loads `settings-i18n.js`. The HUD gets its
dictionary from `src/i18n.js` via `getDashboardI18nPayload()` (`src/main.js:426-429`),
and `t()` is:

```js
return dict[key] || key;   // src/session-hud-renderer.js:20-23
```

so a key filed in the wrong dictionary doesn't throw — it renders its own name into
the UI. The four keys had no Settings-window consumer at all (their only reference is
`session-hud-renderer.js`), so they were dead entries there. Hence a move rather than
a copy: keeping both would leave a second source of truth free to drift.

## Guard test

`test/i18n.test.js` gains two tests, both confirmed to fail on `main`:

1. Every `t("key")` literal in a renderer whose window does not load `settings-i18n.js`
   must resolve in `src/i18n.js`. Renderers are discovered by walking `src/*.html`, with
   a canary asserting the known ones are found so the test can't pass vacuously.
2. No literal in those renderers may resolve *only* in `settings-i18n.js` — the exact
   shape of this bug.

Test 1 also resolves `t(tipKey)`-style indirection back to the literals assigned to that
identifier, since the pin/unpin keys reach `t()` through a ternary
(`session-hud-renderer.js:356`) and a literal scan cannot see them.

Known limits, stated rather than papered over: this is source scanning, not dataflow.
Keys reached through a lookup table (`t(entry.key)`) or resolved from a main-process
payload at runtime stay out of scope, and `tutorial-renderer.js` uses
`i18n(key, fallback)`, so a miss there degrades to English rather than a raw key.

## Validation

- Full suite green: 4997 tests, 4981 pass, 0 fail, 16 skipped.
- Both new tests fail on `main` and pass here.
- Simulated the real payload path (`getDashboardI18nPayload` → renderer `t()`): before,
  all 5 locales leak raw key names; after, all 20 key×locale combinations resolve to
  real text.
- Live-verified on Windows: the raw key names appear on the released build and are gone
  on this branch.

## Relation to #662

#662 notes that the crossed-out marker's hover tooltip is "the only signal today" that a
session can't be focused — and that tooltip is one of the four currently rendering a raw
key name. This fixes the string layer only and touches no focus logic, so it should sit
cleanly underneath whatever UX that issue grows on top.


rullerzhou-afk (approved):
Pulled the branch and verified every claim independently. Everything checks out — approving.

**What I verified:**

- **Byte-identical move**: script-compared the four keys' values between `main`'s `settings-i18n.js` and this branch's `i18n.js` across all five locales — all 20 strings identical, old entries fully removed.
- **No settings-window consumer**: repo-wide grep confirms the only consumer of these keys is `session-hud-renderer.js`, so a move (not a copy) is the right call.
- **Dictionary path**: `getDashboardI18nPayload()` → `i18n[lang]` (main.js:426-429) confirmed as the HUD's string source.
- **Red on main**: restored `src/i18n.js` + `src/settings-i18n.js` to `main` while keeping the new tests — both fail, and the failure message is exactly this bug's shape. Back on the branch: green.
- **Full suite on Windows**: 4997 tests, 0 fail (4979 pass + 18 skipped; the ±2 vs your 4981+16 is platform-conditional skips).
- **Payload-path simulation**: all 20 key×locale combinations resolve to real text through the actual `getDashboardI18nPayload` → `t()` chain.
- **Test robustness**: the renderer discovery correctly excludes `settings.html` and finds both runtime renderers; the `t(tipKey)` indirection handling covers the only such pattern in the repo; and across both renderers' 125 string literals there are currently zero collisions with settings-only keys — no false-positive surface today.
- **Live-verified on a real Windows machine**: the raw key names reproduce on the v0.12.0 installed build, and on this branch the tooltips render real text (checked zh).

Nice work on the PR itself — the root-cause section, tests proven red on `main` first, and the honest "known limits" paragraph made this quick to verify despite touching all five locales.

rullerzhou-afk:
Merged — thank you! 🎉

This was a great first contribution: report, root-cause analysis, fix, and a regression guard all in one tight PR, written up so clearly that every claim reproduced exactly as described on my end. The pin tooltip had been quietly leaking its key name since v0.7.1, so every user gets a little polish from this one. The fix ships with the next release.

Hope to see you around here again!

PR #685, opened 2026-07-15
refactor(codex): extract hook runner and reuse runtime port

Dxy2326:
## Summary
- extract an awaitable, injectable runCodexHook from the bare main entrypoint
- reuse the port already read by the shared Windows resolver for state and permission routing
- preserve the existing permission discovery probe and full fallback behavior without adding a second preflight

## Context
Rebased onto 2179fac from #687. The redundant Codex-only offline gate and its extra 100 ms probe have been removed; offline process-scan gating remains owned by the shared resolver.

## Testing
- node --test test/hook-adapter-offline-contract.test.js test/codex-hook.test.js (64 passed)
- npm test (5186 passed, 18 skipped, 0 failed)

### DIFF
--- a/hooks/codex-hook.js
+++ b/hooks/codex-hook.js
@@ -10,6 +10,7 @@ const {
   postPermissionToRunningServer,
   postStateToRunningServer,
   readHostPrefix,
+  readRuntimeIdentity,
   applyWslSourceFields,
 } = require("./server-config");
 const { createPidResolver, readStdinJson, getPlatformConfig } = require("./shared-process");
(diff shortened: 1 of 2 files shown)


rullerzhou-afk:
Thanks for this — you went straight at the right problem, and the boundary calls match what I converged on independently an hour later in #687: win32-only, `CLAWD_REMOTE` exempt, and the offline decision made *before* any process resolution. I also checked the wiring: `discoverClawdPort` and the `preferredPort` option are real in `server-config.js`, so the port reuse isn't cosmetic.

#687 just merged (2179fac), and it gates the same offline path at the shared-resolver layer for all 13 adapters — codex included; the ETW audit in that PR measured codex at exactly 1 spawn online / 0 offline. So the preflight here would now be a second, redundant offline check on top of the resolver gate, and it puts an extra 100ms probe on the permission-request path.

Two pieces of this PR are real additions #687 doesn't cover, though, and I'd take both:

1. **Port reuse** — carrying the discovered port into the state/permission POSTs as `preferredPort` saves a discovery round.
2. **The `runCodexHook` extraction** — making the hook body injectable/awaitable instead of a bare `main()` is a genuine testability win.

Would you be up for rebasing onto current main and narrowing to those two? Happy to review it fast. The offline-gate half wasn't wasted effort on your part — it independently confirmed where the fix had to sit.

Dxy2326:
Done — rebased onto 2179fac and narrowed #685 to the two additions you called out.

- Removed the Codex-only preflight/offline early return.
- Kept the awaitable/injectable runCodexHook extraction.
- Reused the port read through #687's readRuntimeIdentity resolver seam for both state and permission routing. There is no second HTTP preflight; PermissionRequest keeps its original single discovery probe and fallback chain.

Validation: 64 targeted tests passed, including Codex at 1 spawn online / 0 offline; full npm test passed with 5186 passed, 18 skipped, 0 failed.

rullerzhou-afk (approved):
Reviewed the narrowed version — this is exactly what I asked for, and it held up under scrutiny.

- The preflight is gone; offline gating stays at the shared resolver.
- The port reuse is wired through #687's `readRuntimeIdentity` seam, and I verified the full chain: the resolver really does invoke the injected reader inside the win32 gate, and the captured port lands in both POSTs as `preferredPort`/`runtimePort`. The stale-port case is byte-equivalent to today — `splitPortCandidates` would read the same file via `readRuntimePort()` anyway (which never checked `ownerPid` either), so the net effect is exactly one saved file read per hook event, with the full probe/fallback chain preserved. On mac/linux the gate never reads identity, so behavior there is unchanged.
- The `runCodexHook` extraction is behavior-equivalent path by path: permission stdout is never empty (worst case `"{}"`), the state byte-fit and 100ms timeout are preserved, and stdin errors still exit 0.
- Verification: ran your branch locally — targeted 64/64, full suite 5186 passed / 0 failed / 18 skipped, matching your numbers. I also deployed the hook file on my real setup and exercised all three paths (state updates, permission bubble approve + deny round-trip, offline with Clawd fully quit): all identical to before, which is the correct outcome for this refactor.

Approving and merging. Thanks for the fast turnaround — both halves are nice wins, and the extraction will pay off the next time anyone touches this hook.

PR #687, opened 2026-07-15
fix(hooks): gate the Windows process snapshot on a live Clawd; stop caching agent command lines (#681 Phase A)

rullerzhou-afk:
Phase A of #681, on the plan in `docs/plans/plan-issue-681-offline-hook-process-snapshot.md`.

> Verified on real hardware on both platforms. The Windows ETW spawn audit — including a positive control that reproduces #681 — and the full app-plus-hook handshake are below. The macOS run, which closed the `ps`-shape assumption flagged in [the scoping comment](https://github.com/rullerzhou-afk/clawd-on-desk/pull/687#issuecomment-4981218231), is in [the verification comment](https://github.com/rullerzhou-afk/clawd-on-desk/pull/687#issuecomment-4986888584); its fixture gap is closed on the branch (`e6737bf`). What remains open is listed under [Still not verified](#still-not-verified).

**Supersedes #674, which is now closed unmerged.** Its commit is cherry-picked here (`7ac0da7`) with byte-identical content — same `git patch-id` (`220ec05a…`) as the head it was closed at — so closing it lost nothing; #634 Slice 1 now ships only through this PR, and rolling this back rolls back both. It was folded in rather than merged separately because Phase A rewrites the same three files, and a second cache implementation alongside it was never an option. Worth noting: #674 was never validated against current `main` — its branch sat on `5c2b1f0` (#672), two commits behind. This branch is on `dfbd326`, and the union is what the numbers below were measured on.

### 1. Quitting Clawd did not stop the process snapshot

Every hook event resolved the process tree **before** discovering nobody would receive the POST. So after a normal Quit, each CLI's leftover hook still cold-started a hidden PowerShell that read `ProcessId` / `ParentProcessId` / `Name` / `CommandLine` for every process on the machine — the exact operation in the #681 report, performed for a server that no longer exists.

`~/.clawd/runtime.json` now carries an `ownerPid`. The Windows resolver validates `app` + `port` + a **live** `ownerPid` before it even `require`s `child_process`, so a clean offline is structurally spawn-free rather than merely fast. A stale file left behind by a crash fails the liveness check. `CLAWD_REMOTE` skips the local walk for the same reason — a remote hook's ancestors are on the remote box.

`ownerPid` liveness is a **liveness hint, not an ownership proof**. `processAlive` maps `EPERM` to "alive", which says nothing about whether that PID is still Clawd, and PIDs get reused. Called out in the code rather than papered over.

### 2. A failed walk shipped a wrong PID

`terminalPid || lastGoodPid` fell back to the untouched `startPid` whenever the walk read nothing — the ephemeral per-event hook wrapper, dead by the time anyone clicks it.

That is worse than reporting nothing. `src/state.js` merges `sourcePid || existing.sourcePid`, so a truthy-but-wrong value does not merely fail to help — it **overwrites a correct PID the server already knew**, and click-to-focus targets a dead process for the rest of the session.

Two distinct ways to read nothing, and checking only the first is not enough:

| | snapshot | walk | old result | now |
|---|---|---|---|---|
| empty snapshot | 0 rows | reads nothing | `stablePid = process.ppid` | unavailable, `snapshot-failed` |
| snapshot lacks our own PID | full | reads nothing | `snapshotOk: true`, `stablePid = process.ppid` | unavailable, `snapshot-self-not-found` |

The second is the sharper trap precisely because `snapshotOk` is `true`. The reasons stay distinct because they look identical from outside but have opposite causes: one means the PowerShell was blocked, the other means it worked fine and the tree was stale.

### 3. The pid cache no longer stores the agent's command line

v1 persisted the whole command line to a `%TEMP%` file so a later cache hit could re-run one regex against it. That file outlives the hook, is readable by anything running as the user, and — because the sweep only collects caches whose PIDs are **dead** — a long-lived session's copy was never collected at all. All of it to answer "is this `claude -p`?".

v2 stores the boolean. The derivation moved into the resolver (which is what writes the cache), so a fresh walk and a cache hit cannot disagree. Legacy v1 entries are promoted in memory and deleted — **including when the v2 write fails**, which reverses a deliberate choice in #674. That choice was right when v1 and v2 held the same data; they no longer do, and keeping the raw line around to save a possible future re-resolve is not the trade this makes. The cost is bounded: at most one later re-resolve, and a clean offline still cannot spawn.

Nothing writes v1 any more — `writePidCache` has no production callers left, and a test asserts it stays that way.

## The 13 adapters are untouched — why that is safe

Not reordering them was a scope decision, so the safety is proven rather than asserted:

- **`pidChain` is always `[]`, never `null`.** Six adapters (codex, copilot, cursor, kimi, kiro, codebuddy) do a bare `pidChain.length` with no `Array.isArray` guard. In cursor and codebuddy that `TypeError` would unwind past `writeStdoutOnce` and silently rewrite their gating stdout — `{"continue":true}` and `{"decision":"allow"}` — to `{}`.
- **`source_pid: null` is equivalent to omitting it.** `server-route-state.js` normalizes both to `null` (`Number.isFinite(null) === false`), and `state.js` merges `sourcePid || existing.sourcePid`, so an already-known PID survives.

Both are pinned by tests that run the **real hook scripts** as subprocesses against an empty `USERPROFILE`, with `execFileSync` recorded and refused — not by re-implementing the expressions in a fixture.

## Also in here

- `writeRuntimeConfig` honors its boolean contract — `mkdirSync` moved inside its own `try`. It was outside, so an `EACCES` on `~/.clawd` escaped as an exception into `src/server.js`'s `listening` handler, which calls it **above** `settle()` — stranding `startHttpServer`'s promise forever, for something that should have been a warning. Guarded on both sides now.
- Doctor's existing `local-server` warning covers runtime identity instead of gaining a parallel check.

Two more from the post-draft Codex review (both vacuity-checked — reverting each fails exactly its new tests):

- **The same-key v1 is now dropped in *all* outcomes of a usable fresh walk** (`a8417d8`). start/event kept the legacy v1 when their v2 write failed — a baseline an earlier High finding pinned. Re-judged: a dead v1 (crashed session resumed on the same sid/cwd) can never be promoted, the sweep has a 24h age floor, and nothing writes v1 any more, so a failed-write start/event was the only collector its raw command line had. Aligned with `claudePromote`'s privacy-first ordering; a live v1's worst case is one re-fresh, the bounded cost §3 already accepts. A degraded walk still keeps v1.
- **`clearRuntimeConfig` only unlinks a `runtime.json` this process owns** (`a90d0fc`). Installed + dev builds hold different singleton locks and share the one file; an earlier instance quitting later used to delete the survivor's identity — pre-gate that cost one port-probe fallback, post-gate it blinds every Windows hook until the survivor restarts. Legacy/corrupt files are still cleaned; a foreign `ownerPid` is refused without a liveness check (a dead foreign owner's file is already harmless to the gate).

## Testing

`npm test` — **5201 tests, 5183 pass, 0 fail, 18 skipped** as of `a90d0fc`, including the post-review additions (the 18 are the pre-existing non-win32 platform skips; the targeted tables below are from the original draft run at 5178).

Targeted:

| command | tests | pass | fail |
|---|---:|---:|---:|
| `node --test test/shared-process.test.js test/pid-cache.test.js test/clawd-hook-pid-cache.test.js` | 126 | 126 | 0 |
| `node --test test/pid-resolver-context.test.js` | 38 | 38 | 0 |
| `node --test test/server-config.test.js test/doctor-local-server.test.js` | 58 | 58 | 0 |
| `node --test test/codex-hook.test.js test/copilot-hook.test.js test/qwen-code-hook.test.js` | 126 | 126 | 0 |
| `node --test test/pid-resolver-offline-gate.test.js` | 27 | 27 | 0 |
| `node --test test/hook-adapter-offline-contract.test.js` | 24 | 24 | 0 |
| `node --test test/pid-cache-sanitized.test.js` | 17 | 17 | 0 |
| `node --test test/state-source-pid-merge.test.js` | 10 | 10 | 0 |
| `node --test test/server-runtime-write.test.js` | 8 | 8 | 0 |

Two baselines were **intentionally inverted**, both documented at the assertion:

- `stablePid decays to startPid on empty snapshot` → must now be `null`. Refusing to *cache* a degraded PID was never enough; it was still *shipped* on that event.
- `v2 write FAILURE ... does NOT delete v1` → now deletes it (see §3 above).

Every fix here was checked against a deliberately reverted implementation to confirm the test actually fails without it, rather than passing vacuously.

Two real regressions were found in review and fixed in this branch, both reproduced before being believed:

- **macOS/Linux headless was broken** by the derivation move — `resolveWithContext`'s non-Windows early return went through the raw walk instead of the deriving path, so every `claude -p` reported as interactive and would have shown up in the Session HUD. The non-Windows suite could not have caught it: its `ps` mock answered `"bash"` to every query, so no command line ever reached the derivation. The mock now has a seam.
- **`snapshot-self-not-found`** (§2 above) — the original fix only covered the empty-snapshot case.

## Real-machine spawn audit

Windows 11 x64, Node 24.12. **4688 process-creation auditing is off on this machine and turning it on is a system-wide policy change, so this uses `Win32_ProcessStartTrace` instead** — same ETW source, transient subscription, nothing left behind. Children are attributed by `ParentProcessId == the hook's own node pid`, which is the distinction plan §12.1.4 asks for: the CLI's own outer PowerShell wrapper is a sibling, not a child, and is never counted as the inner snapshot.

Only the outbound POST is blocked (a live Clawd was listening on 23333 throughout; a verification run must not inject a fake session into it). `execFileSync` is untouched, so a real PowerShell starts if the code still wants one — which is the entire point.

### The control comes first

A "0 spawns" number is worthless without proof the measurement can see a spawn at all:

| case | hooks | runtime.json | result |
|---|---|---|---|
| **A — control** | `origin/main` (pre-fix) | absent | **1 × powershell.exe**, 3/3 runs |
| **B — the fix** | this branch | absent | **0**, 3/3 runs |
| **C — not broken** | this branch | valid, live `ownerPid` | **1 × powershell.exe**, 3/3 runs |
| **D — cache hit** | this branch | valid, live `ownerPid`, same session as C | **0**, 3/3 runs |

A reproduces #681 on real hardware. B is the fix. C proves the gate did not simply break the online path. D is #627/#634's cache doing its job.

### All 13 adapters

Each adapter run twice with a fresh session id — once with a live runtime, once with none:

| | online | offline |
|---|---|---|
| clawd, codex, copilot, cursor, gemini, kimi, kiro, codebuddy, antigravity, qoder, qoderwork, qwen-code, reasonix | **1 spawn each** | **0 each** |

The online column is the point. Without it, "0 offline" cannot be distinguished from "this adapter never ran" — and that is not hypothetical: the first pass had **copilot at 0 online and 0 offline**, because copilot takes its event name from `argv[2]` rather than the stdin payload and the harness only fed it stdin. It had been passing the automated suite without exercising a single line of the resolver. Fixed in `116d94c`, along with a vacuity guard that now requires exactly one spawn per adapter with a live Clawd — verified by re-breaking copilot and watching the guard fail with `got []`.

### Two more runtime shapes

| scenario | runtime.json | result |
|---|---|---|
| crashed instance | `{app, port, ownerPid: <dead>}` | **0 spawns** |
| pre-#681 Clawd (**the real upgrade path**) | `{app, port}` — no `ownerPid` | **0 spawns** |

The second is not synthetic: it is byte-for-byte what the Clawd running on this machine right now has written. It is also the documented cost of this change — hooks fail closed until Clawd is restarted once on this build.

### The sanitized cache, from a real file

Written by a real hook, by the real resolver, into the real `%TEMP%`:

```json
{
  "version": 2,
  "namespace": "claude-code",
  "cwd": "D:/repo",
  "stablePid": 50864,
  "agentPid": 76872,
  "headless": false,
  "detectedEditor": "code",
  "ts": 1784111472148
}
```

No `agentCommandLine`. That is A2's whole point, on disk rather than in an assertion.

### The full handshake, end to end

The runs above all drove the gate by making `runtime.json` unreachable — the identical resolver input to a Quit, but a simulation of how that input arises. Two gaps followed from that, and the second was easy to miss: the app under test was never actually **this branch**. `writeRuntimeConfig`'s new `ownerPid` had never been written by a real Electron process even once; only unit tests had exercised it. So what had been proven was "an *old* app quits → *new* hooks stay quiet", which is one half of a handshake.

Both halves now, in one pass, by hand:

**1 — this branch's app starts and writes the identity**

```json
{ "app": "clawd-on-desk", "port": 23333, "ownerPid": 87208 }
```

`87208` is the live Electron main process. The field's mere presence is also the proof that the app under test is this branch: the released code does not write `ownerPid` at all.

**2 — the online path resolves, with a real session as the source**

Starting the app repointed Claude's hooks at this worktree, so the events came from the actual Claude Code session doing this work — not a synthetic payload. What the server recorded, and what those PIDs really are:

| recorded | actually | |
|---|---|---|
| `sourcePid=50864` | `Code.exe` | the VS Code terminal hosting the session — the focus target |
| `agentPid=76872` | `claude.exe` | the agent itself |
| `pidReachable=1` | | the focus target is alive |
| `headless=0` | | interactive, correctly |

**3 — and it resolves to exactly what the old code did**

```
19:40:56   old app + old hooks    agentPid=76872 sourcePid=50864 pidReachable=1 headless=0
19:44:56   this branch, both      agentPid=76872 sourcePid=50864 pidReachable=1 headless=0
```

Byte-identical. The gate, the no-degrade guard, and moving the headless derivation into the resolver changed *nothing* about what a working online resolve produces. That is the non-regression evidence — not an absence of complaints.

**4 — this branch's app quits, and its own hooks go quiet**

```
Clawd process (electron)      0
launch.js still running       0
runtime.json deleted          True
23333-23337 free              True
```

| hooks, in that exact state | result |
|---|---|
| `origin/main` (pre-fix) | **1 × powershell.exe** (`#35132`) |
| this branch | **0** |

The control is what makes the zero mean something: the old code really does start a hidden PowerShell after a normal Quit — #681 reproduced at full fidelity, nothing faked — and the same measurement reports zero for the fix.

### Still not verified

- **Terminal focus was not clicked.** The evidence is that `sourcePid` lands alive and correct (`Code.exe`), which is exactly what `session-focus.js` requires — but nobody clicked the HUD and watched a window come forward. The pet window cannot be screenshotted (MPO returns black), so log ground truth is the ceiling here.
- Codex App's six events in the reporter's own configuration and host form.
- The v0.12.0 legacy **v1** cache upgrade path against a real pre-existing file.
- Remote SSH; auto-start ON/OFF.
- x64 only. No ARM64 device.

## Deliberately not fixed

- **Windows online still runs the snapshot.** That is B1a's scope (#627, #634). This branch does not reduce the online alert surface at all, and #681 cannot close on it.
- **POSIX has the same "unverified startPid" hole** — if `ps` fails for `startPid`, the walk still reports it (verified: `{"stablePid":777777,"pidChain":[],"snapshotOk":true}`). Left alone because the plan scopes no-degrade to Windows and POSIX behavior is explicitly not to change here; the risk is also a tier lower, since `startPid` is `process.ppid` — the live parent waiting on the hook — rather than a separate cold-start PowerShell that can be blocked wholesale. Removing one `isWin` guard would close it symmetrically. Open question, not an oversight.

## Behavior change to know about

Until Clawd is restarted once on this build, the older `runtime.json` has no `ownerPid`, so hooks fail **closed** and omit process metadata — a new session has no click-to-focus target until then. `state`/`permission` POSTs keep working through the existing port fallback. One restart fixes it. Documented in `known-limitations`.

## Not claimed

No security product is detected, branched on, or whitelisted, and none was tested against. This does not claim Windows no longer scans processes, that #627/#634 are resolved, or that #681 can be closed.


rullerzhou-afk:
## What a Mac needs to verify here — and what it doesn't

This is a Windows issue, so most of this PR is structurally unreachable on macOS/Linux. Writing the boundary down so nobody re-derives it, and so that if a Mac run does happen it tests the one thing that actually needs it.

### Out of reach on POSIX

- The offline gate is inside `if (isWin)`, so `runtime.json` is never read there.
- The disk cache is further out still: `resolveWithContext` hits `if (!isWin) return freshMetadata()` **above** the `require("./pid-cache")` below it — macOS never loads the cache module at all.
- Both no-degrade guards are `if (isWin && …)`.

No gate, no disk cache, no sanitized-cache surface on a Mac.

### The one thing this PR does change on POSIX

`headless` derivation moved out of `hooks/clawd-hook.js` into the resolver, injected as `headlessCheck`. Same regex, same input string, different call site: `freshMetadata()` derives it from the same `meta.agentCommandLine` the adapter used to parse itself.

That move was broken once — `resolveWithContext`'s non-Windows early return went to `freshResolve()`, which carries no `headless`, so every `claude -p` on macOS/Linux reported as interactive and turned up in the Session HUD. Caught in review, fixed in `18e5a4f`, and covered by `test/clawd-hook-pid-cache.test.js`, which drives the real POSIX walk and the real `buildStateBody` across all four lifecycles rather than a stub.

### The honest gap

That test feeds a **hand-written** `ps` string:

```
node /opt/claude-code/cli.js --print
```

I have never captured what `ps -o command= -p <pid>` — the exact call the resolver makes on POSIX — actually prints for a `claude -p` run on a real Mac. If the real output is shaped differently, `/\s(-p|--print)(\s|$)/` may not match and headless would be wrong on macOS.

**This is pre-existing, not a regression from this PR.** Before this change `clawd-hook.js` ran the same regex against the same string from the same `ps` call. A Mac run would surface an old bug, not a new one — which is why it is not a merge blocker here.

### How to check it, ~5 minutes

1. Start Clawd on a Mac from this branch.
2. Run `claude -p "hi"`. It must **not** appear in the Session HUD — the HUD lists exactly the non-headless live sessions.
3. Run an ordinary interactive `claude`. It must appear, and clicking it must focus that terminal.

If step 2 shows the session, the `ps` assumption above is wrong. That is a real finding, but it belongs in its own issue against the pre-existing headless detection, not in this PR.

The cheapest thing worth capturing while there is the ground truth the fixture is guessing at:

```sh
# with `claude -p "…"` running
ps -o command= -p "$(pgrep -f claude | head -1)"
# and again with an interactive `claude`
```

Pasting that output lets the fixture stop being a guess.

### Also unverified on POSIX, deliberately

The unverified-startPid hole is symmetric and left alone here as a scope decision: on POSIX a successful walk that never finds the start process still returns `stablePid` from `terminalPid || lastGoodPid`. Windows now fails closed on `pidChain.length === 0`; POSIX does not. One `isWin` guard away if that scope is ever opened.

rullerzhou-afk:
## Real-machine verification on macOS — the `ps` assumption holds

Ran the Mac check from the comment above, on this branch (`efd1bd2`; one commit behind `main` — `1842ea8`/#584, unrelated to the resolver). Apple Silicon, Darwin 25.5.0, Node v24.16.0, Claude Code CLI `2.1.181` installed as a **native binary** at `~/.local/bin/claude`.

**Headline: the ground truth the fixture is guessing at is captured, and the regex verdict is unchanged — but the real shape is nothing like the fixture's.**

### The ground truth

```
$ claude -p "hi"            # pid 19015
$ ps -o comm= -p 19015
claude
$ ps -o command= -p 19015
claude -p hi
```

Interactive, for contrast (pid 12161):

```
$ ps -o comm= -p 12161
/Users/…/claude-code/2.1.209/claude.app/Contents/MacOS/claude
$ ps -o command= -p 12161
/Users/…/MacOS/claude --output-format stream-json --verbose --input-format stream-json
  --effort max --model … --permission-prompt-tool stdio --permission-mode auto …
```

| | real macOS | the fixture's guess |
|---|---|---|
| `comm=` | `claude` | `/usr/bin/node` |
| `command=` | `claude -p hi` | `node /opt/claude-code/cli.js --print` |

`/\s(-p|--print)(\s|$)/` matches `claude -p hi`, and does not match the interactive line — `--permission-prompt-tool` and `--permission-mode` both start `--p` but trip neither alternative. **"The honest gap" closes with the assumption intact. No finding, no new issue.**

### The A/B — the control the number needs

`headless` on the wire, from real hook POSTs. Method: Clawd on 23333, `runtime.json` repointed to a recording proxy on 23334 that logs each body verbatim and forwards upstream — the app stays live and nothing synthetic is injected. Real hooks, real resolver, real `ps`.

| run | command | `headless` in the POST |
|---|---|---|
| pid 28172 | `claude -p "hi"` | **`true`** — all four: SessionStart(`idle`), UserPromptSubmit(`thinking`), StopFailure(`error`), SessionEnd(`sleeping`) |
| pid 28821 | `claude`, real PTY via `script` | field absent |
| pid 30442 | `claude`, real Terminal.app | field absent |
| pid 14106 | this session, live interactive | field absent |

The field tracks `-p` and nothing else. Without the bottom three rows, `headless: true` cannot be distinguished from *always* true.

`assert.ok(!("headless" in body))` in the interactive test matches the wire exactly — absent, not `false`.

### `sourcePid` lands alive and correct

The same standard as the Windows handshake, on a real terminal:

| recorded | actually | |
|---|---|---|
| `source_pid=4218` | `/System/Applications/Utilities/Terminal.app/Contents/MacOS/Terminal` | alive — the focus target |
| `agent_pid=30442` | `/Users/…/.local/bin/claude` | alive — the agent |
| `pid_chain` | `[30442, 29771, 29769, 4218]` | claude → zsh → zsh → Terminal |

Terminal detection works on the real walk: `comm=` basename is `terminal`, which `BASE_TERMINAL_NAMES_MAC` carries.

### The app half, on a Mac

`npm start` from this branch wrote:

```json
{ "app": "clawd-on-desk", "port": 23333, "ownerPid": 23655 }
```

`23655` is the live Electron main process. Same argument as on Windows: the field's presence is the proof that the app under test is this branch. First time the app half of the handshake has run on macOS.

### `npm test` on macOS

**5146 tests, 5128 pass, 0 fail, 18 skipped** (~21s). The body's 5178/5160/18 are Windows.

### The POSIX unreachability claims, checked rather than read

`require.cache` around a `cacheable: true` context resolve on macOS:

```
before: []
after:  []
headless: false | cacheSource: fresh | agentPid: 14106
```

"macOS never loads the cache module at all" — confirmed on the running thing, not just from the early return.

---

## Three things worth folding in

**1. The POSIX headless test drives a branch a real Mac never takes.**

```js
// A `claude -p` run under node, the way it actually looks on macOS/Linux.
const asHeadlessClaude = () => {
  env.state.psComm = "/usr/bin/node";
  env.state.psCommand = "node /opt/claude-code/cli.js --print";
};
```

That comment does not hold for a native install, which is the default now. `psComm = "/usr/bin/node"` sends the walk down `agentCmdlineCheck` (`name === "node"`). A real Mac has `comm=` basename `claude`, so it goes down `agentNameSet.has(name)` — **a different branch**, three lines up.

Both branches set `agentCommandLine`, and both strings match the regex, so the verdict is identical and this is not a bug. But the test added to pin the regression review caught is exercising a path a current Mac install never reaches. A second case with `psComm = "claude"` / `psCommand = "claude -p hi"` would cover the real one.

**2. A Mac run does not need working auth.** The CLI's OAuth is expired on this machine — `claude -p "hi"` dies with `401 OAuth access token has expired`. All four hook lifecycles fired anyway, `headless: true` on each. The failure is downstream of the hook path, so the ~5-minute check works on a machine that cannot authenticate. Worth stating, since a reader who hits the 401 would reasonably assume the check is void.

**3. The POSIX unverified-startPid hole reproduces — but I had to induce it.** After killing the PTY's parent tree, SessionEnd came back:

```
event=SessionEnd  state=sleeping  agent_pid=28821  source_pid=28821  chain=[28821]
```

The walk collapsed to one entry and shipped the exiting `claude` itself as `sourcePid`, overwriting the correct `1165` the server already had — `terminalPid || lastGoodPid` doing exactly what "Deliberately not fixed" predicts. In fairness: I killed the parent to get there, it is SessionEnd/`sleeping` so the HUD excludes it regardless, and nothing here happens on its own. Prediction confirmed, not a new finding.

## Still not verified

- **Nobody clicked the HUD entry.** `sourcePid` lands alive and correct on a real Terminal.app, which is what `session-focus.js` needs — the same ceiling as the Windows run. The pet screenshots fine on macOS (no MPO black), but the Session HUD panel has to be opened from the pet first.
- The HUD gate itself (`session-hud.js:55`) is untouched by this PR and is a pure function of `headless`, so it was read rather than driven.
- Intel Mac — this is Apple Silicon only. And Linux.

PR #696, opened 2026-07-16
docs: fix known-limitations links and Chinese parity

Dxy2326:
## Summary
- fix the broken Back to README links in both Known Limitations guides
- update the Chinese Hermes limitation to match current clarify and permission-bubble support
- add the missing Chinese Wayland limitation so the English and Chinese tables stay aligned

## Root cause
The guides live under `docs/guides/`, but their README links only traversed one directory. The Chinese guide had also fallen behind the English capability notes.

## Validation
- relative Markdown link scan: all local targets exist
- limitation table parity: 32 English rows / 32 Chinese rows
- `git diff --check`

### DIFF
--- a/docs/guides/known-limitations.md
+++ b/docs/guides/known-limitations.md
@@ -1,6 +1,6 @@
 # Known Limitations
 
-[Back to README](../README.md)
+[Back to README](../../README.md)
 
 | Limitation | Details |
 |---|---|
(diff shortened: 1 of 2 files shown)


rullerzhou-afk:
Thanks for catching these documentation drift issues and keeping the English and Chinese limitation guides aligned! The corrected README links and Hermes/Wayland parity updates look good. 🙌

PR #697, opened 2026-07-16
fix: let permission bubbles return to terminal

### LINKED ISSUE #689: 权限请求/提示不审批关不掉！
是不是可以加个稍后处理/弹窗关闭等功能
我记得之前有个“前往终端”？

lurui1997:
## Summary

- add a Go to Terminal action to regular interactive permission bubbles
- reuse the existing deny-and-focus fallback without approving or denying on the user's behalf
- keep passive notifications, elicitation, and plan-review flows on their existing specialized paths
- add renderer-level DOM behavior coverage across supported permission sources

## Root cause

Regular permission bubbles exposed only Allow and Deny even though the backend already supported handing the undecided request back to the native terminal. Users therefore had no safe way to dismiss the Clawd bubble without making a decision.

## Validation

- targeted permission/bubble tests: 76 passed
- full suite: 5167 tests, 5149 passed, 0 failed, 18 skipped
- independent code review: approved

Fixes #689

### DIFF
--- a/src/bubble-renderer.js
+++ b/src/bubble-renderer.js
@@ -660,6 +660,17 @@ function renderElicitationTerminalFallback() {
   suggestionsContainer.appendChild(btn);
 }
 
+function renderRegularTerminalFallback(lang) {
+  const btn = document.createElement("button");
+  btn.className = "btn-suggestion";
+  btn.textContent = bubbleText(lang, "goToTerminal");
+  btn.addEventListener("click", () => {
+    disableAll();
+    window.bubbleAPI.decide("deny-and-focus");
+  });
+  suggestionsContainer.appendChild(btn);
+}
+
 function renderElicitationStep() {
   const total = elicitationQuestions.length;
   if (total === 0) {
@@ -780,7 +791,7 @@ function show(data) {
       });
       suggestionsContainer.appendChild(btn);
     }
-
+    renderRegularTerminalFallback(data.lang);
     revealCard();
     return;
   }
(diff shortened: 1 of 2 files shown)


rullerzhou-afk:
Hi @lurui1997 — thanks for the PR! The direction is exactly right for #689, and the renderer coverage was a good start. I did a deep review with wire-level tracing and found two blocking issues in the current cut, so I've taken over the fix on top of your commit (branch [`fix/689-go-to-terminal`](https://github.com/rullerzhou-afk/clawd-on-desk/tree/fix/689-go-to-terminal), your commit preserved as-is).

**P0-1 — Claude Code / CodeBuddy hang for up to 600s.** `deny-and-focus` routes regular CC-protocol requests to `dismissPermissionForTerminal()`, which removes the entry and hides the bubble but leaves the hook's HTTP response parked. CC blocks on the PermissionRequest hook (registered with `timeout: 600`) and shows **nothing in the terminal** until the connection dies — so "Go to Terminal" sent users to a blank terminal for up to 10 minutes. The repo already documents this failure mode in the bypass-gate note (`server-route-permission.js`): a *dropped* connection is what makes CC fall back to its native prompt. The fix destroys the socket (after detaching the abort handler), same mechanism the autoclose no-decision path has always used.

**P0-2 — Hermes fails open.** For Hermes, `handleDecide` answers `deny-and-focus` with a 204 no-decision — but the hermes plugin maps 204 to `None`, which its `pre_tool_call` gate treats **exactly like allow** (`hooks/hermes-plugin/__init__.py`). Clicking "Go to Terminal" on a Hermes card would have *executed* the tool the user meant to defer — fail-open past the opt-in `CLAWD_HERMES_PERMISSION_TOOLS` gate. Since the protocol can't express "no decision, user answers in the terminal", the bubble payload now forwards `isHermes` and the renderer suppresses the action on Hermes cards.

Validation: new wire-level test suite (mutation-checked), full suite 5219/0 failures, plus real-machine verification on live Claude Code sessions — both a regular Bash approval and the plan-review path now drop back to the native terminal prompt within ~2s of the click.

I'll land the combined branch shortly — your commit stays as the base of the change. Thanks again for kicking this off!

rullerzhou-afk:
Merged via #704 (81ceb20), which builds on this PR's commit — you're credited as co-author on the squash commit. Thanks again for the contribution!

PR #698, opened 2026-07-16
feat: explain unavailable session focus

### LINKED ISSUE #662: Feature request: explain why a session can't be focused, and offer an alternative action

lurui1997:
## Summary

- explain why non-focusable sessions cannot jump to a terminal in both Session HUD and Dashboard
- distinguish remote, WebUI, and missing-terminal-window cases across all five locales
- offer Open Folder only for eligible local sessions, with visible failure feedback
- pass only sessionId from renderers and re-resolve plus validate the directory in the main process
- share focus/folder policy between HUD and Dashboard and add renderer behavior coverage

## Security and behavior

The renderer never supplies a path. The main process resolves the current live session, rejects remote and WebUI sessions, and requires an absolute existing directory before calling shell.openPath. Terminal focus and PID resolution are unchanged.

## Validation

- focused tests: 22 passed
- full suite: 5168 tests, 5150 passed, 0 failed, 18 skipped
- independent code review: approved after four rounds

Fixes #662


rullerzhou-afk:
Thank you @lurui1997 for contributing this! The shared unavailable-focus reasoning, main-process revalidation, IPC boundaries, i18n coverage, and tests gave the feature a strong foundation. We added a maintainer follow-up for renderer layout/state edge cases and completed a real Windows Electron end-to-end smoke test; Open Folder successfully traversed the full renderer → preload → IPC → main-process → Explorer path. Merged as b830ff0. Really appreciate you taking on #662! 🙌

PR #700, opened 2026-07-16
fix: retain Codex rollout read positions

### LINKED ISSUE #691: Codex replay protection: record what we've read instead of guessing (#566 root cause)
Design note for #566, written after #584's first approach was withdrawn. Records why the obvious fix doesn't work, so the next attempt doesn't rediscover it.

## The bug

With the pet running for days, a finished Codex session replays as live `thinking → working → attention` with zero user activity.

## Why it happens

Both replay guards in `agents/codex-log-monitor.js` anchor on **monitor start**:

```js
// line-level
if (!tracked.backfilling && ts < this._startedAtMs - 1500) return;
// file-level
backfilling: stat.mtimeMs < this._startedAtMs - BACKFILL_GRACE_MS
```

Run the pet for a week and every line of a session that ran on Wednesday is "newer than monitor start", so both guards wave it through. Codex Desktop's launch-touch of an old rollout then refreshes the mtime, the file re-enters the pickup window, the monitor attaches **at offset 0**, and the whole finished session replays.

The offset-0 attach is the part that needs explaining, because the bookkeeping normally prevents it:

- `_tracked` (cap `MAX_TRACKED_FILES = 50`) never drops a file on its own — once attached it is polled every pass regardless of mtime, so a later touch reads forward from the saved offset.
- LRU eviction moves it to `_retiredTracked` (cap `MAX_RETIRED_TRACKED_FILES = 100`), which **preserves the offset** and restores it via `resumeOffset`.

So the offset has to be *gone*: evicted from `_tracked`, then pushed out of `_retiredTracked` by ~100 further retirements, and only then touched. Reachable for a heavy Codex user whose pet has been up for days — `_tracked` only grows until it hits 50. (`_pruneTrackedFilesIfNeeded` retires `neverEmitted` trackers first, so the exact volume depends on which files ever emitted; ~150 is the shape, not a guarantee.)

**That is the actual defect: we lose track of what we have already read.**

## What was tried and withdrawn (#584)

Re-anchoring both guards on per-file attach time (`tracked.attachedAtMs`) instead of monitor start, plus a filename-staleness signal for timestamp-less lines. Measured against real code, three versions, via `_poll()`:

| scenario | `main` | re-anchored | + snapshot exception |
|---|---|---|---|
| #566 (touched stale rollout) | ❌ full replay | ✅ idle only | ✅ idle only |
| short turn, discovered 6s later | ✅ full turn | ❌ **swallowed** | ❌ **swallowed** |
| restart 2s after a turn ended | ✅ silent | ❌ **replays** | ❌ **replays attention** |

It fixes #566 and breaks two other things. **This is not fixable by tuning the threshold** — the two failures are one contradiction seen from both sides:

- **Swallowing**: a turn that ended 6s ago and is only now discovered looks exactly like history.
- **Replaying**: a turn that ended 2s ago and was *already shown* looks exactly like new work.

The file bytes, timestamps, mtime, filename, and size can be **identical** in both worlds. Any predicate reading only those values must answer both cases the same way. A wider window swallows less and replays more; a narrower one does the reverse. `mtimeAtAttachMs` is a different guess, not more information.

Also worth recording: the restart-replay window is **not new** — `main` already replays a turn that ended within ~1.5s of a restart. The withdrawn work widened it to 5s. Measured:

```
turn ended N ms before restart:   main        re-anchored
   500ms                          replay      replay
  2000ms                          silent      replay
  4000ms                          silent      replay
  6000ms                          silent      silent
```

### The filename signal was unnecessary

Its only stated justification was timestamp-less historical lines defeating the line-level guard. Checked against real data: **383 rollouts, 209,309 lines, zero without a timestamp**; of the lines matching the record types this monitor maps, also zero.

Old Codex (`rust-v0.30.0`) did write bare `ResponseItem` lines with no timestamp — but this monitor only recognizes envelope keys like `response_item:function_call` and ignores the bare form entirely. Upstream introduced the `RolloutItem` envelope **in the same commit** as per-line `RolloutLine.timestamp` ([openai/codex@43809a4](https://github.com/openai/codex/commit/43809a454e5c6348418fc2d5ace1eb0a98f59847)). A line this monitor can recognize *and* lacks a timestamp is structurally impossible, not merely rare.

## Proposed approach: an in-process offset ledger

Stop asking "is this content history?" and start recording "how far have we read?". The monitor already computes the offset; it just lets it die with the LRU.

- Keep a `Map<filePath, offset>` that is **not** subject to the 50/100 eviction. `_tracked` / `_retiredTracked` keep caching the rich state as they do now.
- On attach, resume from the ledger offset when present. Only new bytes are processed — no timestamp comparison, so neither failure above can arise.
- On file shortening or identity change, **silently re-baseline**. Do not reset the offset to 0 and read as live.
- Lock it with an integration test that drives real churn (150+ rollouts through `_poll()`), asserting history does not replay *and* a new turn is not swallowed.

Memory cost is a path plus a number per rollout — kilobytes. Worth a cap or GC eventually, not on day one.

### Pre-existing edge to fix along the way

```js
const resumeOffset = retired && stat.size >= retired.offset ? retired.offset : 0;
```
A retired file that shrank below its saved offset resumes at 0 — but `backfilling` is gated on `!retired`, which is false, so it reads the whole file **as live**. Same defect class: offset lost, no backstop.

### On-disk persistence: not yet

An in-process ledger does **not** survive an app restart, so the ~1.5s restart-replay window on `main` stays as-is. That is the status quo and is not what #566 reports.

Persisting the ledger would close it, but only after deciding that cross-restart consumption history is actually required — it brings callback-vs-checkpoint commit ordering, truncation/replacement detection, atomic writes, and stale-record GC. The single-instance lock doesn't settle exactly-once at crash time. Worth its own discussion, not a silent add-on.

## Why the previous fixes didn't hold

`e5c0eab` (turn-end state resolution), `cf5e9dd` (long-lived session discovery), `08ce441` (stale attach state) addressed neighbouring problems. Only `6b5c408` (file-level gate) and `2d72dfe` (carried the line-level guard forward) touched the guards, and both kept monitor start as the reference.

Six attempts, all of them refining a *guess*. The thing worth recording is that the question was wrong: **"have we read this?" is bookkeeping, not inference.**

---

Evidence and the withdrawn implementation live on `backup/584-full-attempt` (not pushed). #584 is now scoped down to the independent `token_count` fix and no longer claims #566.

lurui1997:
## Summary

- retain a process-lifetime committed offset ledger independently of the active and retired tracker LRUs
- resume evicted rollouts from the last complete JSONL record instead of offset zero
- silently rebaseline truncated or replaced files and continue consuming later appends
- harden reads against partial lines, short reads, file replacement races, and unreliable Windows inode metadata
- add real _poll churn coverage across more than 150 rollouts

## Root cause

After both tracker LRUs forgot an old rollout, a later mtime touch reattached it at offset zero. Because replay guards were anchored to monitor startup, a long-running monitor treated the historical turn as live work.

## Validation

- codex-log-monitor tests: 58 passed
- full suite: 5169 tests, 5151 passed, 0 failed, 18 skipped
- independent code review: approved after three rounds

Fixes #691
Refs #566


rullerzhou-afk:
Thanks @lurui1997 for finding the real root cause and putting together the offset-ledger fix — this was a tricky long-running-session bug.\n\nBefore merging, we merged the latest `main` into this branch and added a follow-up on top of your work to reconcile it with #707's startup `request_user_input` recovery. In particular, recovery now follows the same disk-tail offset model, seeds `fileIdentity` and `_readPositions`, and includes regression coverage for recovery plus double eviction. The core approach and credit remain yours. Thanks again!

PR #710, opened 2026-07-18
fix: harden Electron install integrity

### LINKED ISSUE #709: Bug: Electron postinstall can silently leave an incomplete macOS app bundle
## Summary

On macOS arm64, Electron's npm postinstall can exit with status 0 while leaving a partially extracted `Electron.app`. The resulting package may contain the launcher executable but omit `Contents/Frameworks/` and `path.txt`.

This was observed in the local development checkout and is reproducible with the currently locked dependency combination. The issue tracks preventing the failure from recurring and, after a minimal upstream reproduction is confirmed, reporting it to Electron or `extract-zip`.

## Environment

- macOS 26.5.2 (build 25F84), arm64
- Node.js 24.16.0
- npm 11.13.0
- Electron 41.0.2
- `@electron/get` 2.0.3
- `extract-zip` 2.0.1
- `yauzl` 2.10.0

## Symptoms

`npm start` failed with:

```text
dyld: Library not loaded: @rpath/Electron Framework.framework/Electron Framework
```

The local package had these structural problems:

- `node_modules/electron/dist/Electron.app/Contents/MacOS/Electron` existed
- `node_modules/electron/dist/Electron.app/Contents/Frameworks/` was missing
- `node_modules/electron/path.txt` was missing
- the partial `dist/` was only about 244 KiB

Running `node install.js` could exit 0 while the same files were still missing.

## Evidence that the ZIP was valid

A fresh Electron 41.0.2 darwin-arm64 ZIP was downloaded and matched the checksum shipped in `node_modules/electron/checksums.json`:

```text
8e18ef53da62bca6132508721c1f94e06b5773b48d366b95e593479892f0a2fe
```

Additional checks:

- the ZIP size was 116,187,083 bytes
- `unzip -t` reported no errors
- the archive listing contained `Electron Framework.framework` and all normal helper/framework bundles
- disk space was sufficient
- no HTTP, npm, or Electron proxy was configured
- the extracted app did not have a `com.apple.quarantine` attribute

This makes a corrupt cached archive, disk exhaustion, and Gatekeeper unlikely explanations for the incomplete bundle.

## Reproduction evidence

Calling the installed extractor directly under Node 24.16.0 reproduced the silent partial extraction:

```bash
DEBUG=extract-zip node -e '
const extract = require("extract-zip");
console.log("start", process.version);
extract(process.argv[1], { dir: process.argv[2] })
  .then(() => console.log("resolved"))
  .catch((error) => {
    console.error(error);
    process.exitCode = 1;
  });
' electron-v41.0.2-darwin-arm64.zip /private/tmp/electron-extract-repro
```

Observed behavior:

- the process printed `start v24.16.0`
- debug output stopped after opening the read stream for `Electron.app/Contents/Resources/electron.icns`
- the extraction promise never logged `resolved`
- the process exited 0 in roughly 0.1 seconds
- the destination contained the same approximately 244 KiB partial tree

This currently proves a reproducible failure for this local environment and dependency combination. It does not yet prove that all Node 24 or macOS installations are affected.

## Workaround used

The checksum-verified ZIP was extracted with the macOS system `unzip`, and `path.txt` was restored with the exact content:

```text
Electron.app/Contents/MacOS/Electron
```

Afterwards:

- `Contents/Frameworks/` contained Electron Framework, helper apps, Mantle, ReactiveObjC, and Squirrel
- `require("electron")` returned the expected absolute executable path
- `Electron --version` printed `v41.0.2`
- `npm start` launched successfully
- the local state endpoint returned `{"ok":true,"app":"clawd-on-desk","port":23333}`
- the dyld error no longer occurred

## Why this needs tracking

A missing `path.txt` can be repaired independently and make `require("electron")` or pure-logic tests pass, while the app bundle is still missing its frameworks. The first reliable failure may only appear when Electron is actually launched.

The installation exit code alone is therefore not a sufficient integrity check.

## Expected outcome

One of the following, depending on the confirmed upstream cause:

1. identify and document a known-good development Node/dependency combination;
2. update or override the affected extraction dependency;
3. add a postinstall or prestart integrity check that verifies at least:
   - `path.txt` has the expected value;
   - the Electron executable exists;
   - `Contents/Frameworks/Electron Framework.framework` exists on macOS;
4. ensure an incomplete extraction fails with a non-zero status;
5. file and link a minimal upstream Electron or `extract-zip` reproduction.

## Acceptance criteria

- A clean Electron reinstall on supported macOS/Node versions produces a complete app bundle.
- The structural checks above pass.
- `npm start` launches without a dyld framework error.
- Any partial extraction is reported as a hard failure rather than exit 0.

rullerzhou-afk:
## Summary

- move Electron from 41.0.2 to 41.10.2 on the same major release line, replacing the legacy `extract-zip` / `yauzl` installer path with Electron's maintained extractor
- establish Node 24.18.0 as the repository and CI baseline (`engines >=22.12.0`)
- add a read-only Electron integrity verifier at root postinstall and before development launch, with explicit skip/override semantics and actionable recovery output
- document the supported setup and add macOS false-green plus Windows/Linux fixture coverage

## Root cause

Node 24.16.0's classic-stream regression ([nodejs/node#63487](https://github.com/nodejs/node/issues/63487)) interacts with Electron 41.0.2's legacy extraction chain so `install.js` can exit 0 after creating only a 244 KiB partial app bundle. `path.txt` and `Contents/Frameworks` remain missing, and the first visible failure is dyld at launch.

Electron 41.10.2 uses `@electron-internal/extract-zip`, so this fix stays on Electron 41 while removing the affected stream chain.

## Verification

- old chain, Node 24.16.0: two isolated Electron 41.0.2 installs exited 0 with the same 244 KiB partial bundle and missing `path.txt` / `Frameworks`
- new chain, Node 24.16.0: two isolated Electron 41.10.2 installs completed and passed structural verification
- clean full-repo `npm ci` on Node 24.18.0: postinstall verifier passed
- targeted verifier tests: 24 passed
- full `npm test`: 5,359 tests; 5,341 passed; 0 failed; 18 skipped
- `npm run build:mac`: x64 and arm64 DMGs produced successfully
- real `npm start`: Electron main/helper/renderer processes remained healthy, `/state` returned OK, no dyld/Framework error appeared, and the smoke processes were terminated afterward
- final bundle checks: complete Frameworks set, exact 36-byte `path.txt`, absolute `require("electron")` result, and `Electron --version` = `v41.10.2`

Closes #709


rullerzhou-afk:
Addressed the pre-merge review in `3b62da6`:

- the launch wiring test now separately proves that both the verifier call and Electron `require` exist before comparing their order, closing the `indexOf() === -1` false-green path;
- added `npm run verify:electron` to the AGENTS command list.

Verification after the patch:

- targeted verifier suite: 24 passed, 0 failed;
- full `npm test`: 5,359 tests; 5,341 passed; 0 failed; 18 skipped.

The narrow `ELECTRON_OVERRIDE_DIST_PATH` / missing-`path.txt` resolver mismatch is tracked separately in #711 so #710 does not broaden override semantics at the merge boundary.

PR #721, opened 2026-07-20
fix(win): stop treating the desktop shell as a fullscreen app

### LINKED ISSUE #719: 当clawdondesk运行时，无法在桌面对文件进行重命名
### Describe the bug

On Windows, while Clawd on Desk is running, renaming a newly created folder
or text file directly on the desktop is cancelled almost immediately
(usually within about one second).

The rename input box disappears and the item reverts to its default name,
such as "New folder" or "New Text Document".

The same Desktop directory works normally when opened in a regular File
Explorer window using `shell:desktop`.

Completely quitting Clawd restores normal desktop rename behavior.

### Steps to reproduce

1. Start Clawd on Desk.
2. Right-click the Windows desktop and create a new folder or text document.
3. Try to enter a new name immediately.
4. Observe that the rename input closes within about one second and the
   default name is restored.
5. Press `Win + R`, run `shell:desktop`, and rename an item inside that
   File Explorer window.
6. Observe that renaming works normally there.
7. Completely quit Clawd and repeat the test directly on the desktop.
8. Observe that desktop renaming works normally again.

### Expected behavior

The desktop rename input should remain active until the user confirms or
cancels the rename.

### Actual behavior

The desktop rename input loses focus or is cancelled automatically within
about one second while Clawd is running.

### Isolation results

- Direct Windows desktop view: affected
- The same directory opened through `shell:desktop`: not affected
- After completely quitting Clawd: not affected
- Starting Clawd again: [please verify and fill in]

This suggests that the files, Desktop directory and permissions are normal,
and that the issue affects the Windows desktop shell view or its focus state.

### Environment

- Clawd on Desk version: [check Settings / Doctor]
- Installation method: [Windows x64 installer / ARM64 / source]
- Windows: 25H2, build 26200.8875
- Display scaling: [fill in]
- Free roam enabled: [yes/no]
- Mini mode: [yes/no]

### Possibly related implementation

This may be related to the focusable Windows hit window, but this is only a
possible lead, not a confirmed root cause.

Related issue: #538

The current implementation creates the Windows hit window as focusable and
only disables its activation when a fullscreen foreground application is
detected. The normal Windows desktop is therefore not covered by the
fullscreen focus-stealing workaround.

KaiC5504:
Fixes #719.

Root cause: win-fullscreen-detect.js judges "fullscreen app in foreground" by geometry alone. Clicking the bare desktop makes the shell window (Progman, or a WorkerW when a wallpaper host re-parents the icon view) the foreground window, and its rect covers the whole monitor — so the probe reported a fullscreen app. The ~1s focusable poll then set the hit window non-focusable, and Electron's setFocusable(false) on Windows ends in Focus(false) → Deactivate(), which hands the foreground away — cancelling an in-place rename on the desktop. It also silently made the topmost watchdog stand down whenever the desktop had focus. This matches every isolation result in the issue: the ~1s delay is the poll interval, an Explorer window doesn't cover the monitor, and quitting Clawd kills the poll.

Fix: after the geometry match, read the foreground window's class via GetClassNameW (same koffi idiom as win-foreground-terminal.js) and exclude Progman/WorkerW (case-insensitive). The class read only happens in the rare covers-monitor case; a failed read keeps the old geometric answer.

Evidence (100ms foreground watcher on Windows 11 build 26200, comparing the old geometry-only verdict vs the new one):

```
12:08:37.800 class=CASCADIA_HOSTING_WINDOW_CLASS coversMonitor=false oldProbe=normal newProbe=normal
12:08:43.556 class=Chrome_WidgetWin_0     coversMonitor=false oldProbe=normal newProbe=normal
12:08:43.968 class=Progman                coversMonitor=true oldProbe=FULLSCREEN newProbe=normal   <-- MISDETECT (old code flips Clawd's hit window here)
12:08:44.932 class=CASCADIA_HOSTING_WINDOW_CLASS coversMonitor=false oldProbe=normal newProbe=normal
12:08:49.925 class=Chrome_WidgetWin_0     coversMonitor=false oldProbe=normal newProbe=normal
12:08:50.338 class=Progman                coversMonitor=true oldProbe=FULLSCREEN newProbe=normal   <-- MISDETECT (old code flips Clawd's hit window here)
12:08:50.447 class=XamlExplorerHostIslandWindow_WASDK coversMonitor=false oldProbe=normal newProbe=normal
12:08:52.753 class=Progman                coversMonitor=true oldProbe=FULLSCREEN newProbe=normal   <-- MISDETECT (old code flips Clawd's hit window here)
12:08:55.690 class=CASCADIA_HOSTING_WINDOW_CLASS coversMonitor=false oldProbe=normal newProbe=normal

14:58:06.743 class=Chrome_WidgetWin_1     coversMonitor=true oldProbe=FULLSCREEN newProbe=FULLSCREEN
```

Desktop clicks stop being classified as fullscreen; a real fullscreen app (browser video, last line) is classified identically before and after.

Note on reproducing: the misdetection fires on every machine (watcher above), but whether the rename visibly cancels depends on Windows' foreground lock allowing the background process to deactivate the foreground — it reproduces on some machines (as in the issue report) while others block the yank and only suffer the silent watchdog stand-down.

Tests: 5 new cases in test/win-fullscreen-detect.test.js (Progman/WorkerW excluded, normal class still fullscreen, class-read failure falls back to geometry, plus unit cases for the new helper); the fake koffi gained a GetClassNameW stub. File passes 17/17.

Verified live: desktop rename survives, pet drag works, fullscreen video still classified fullscreen (overlay behavior unchanged).

### DIFF
--- a/src/win-fullscreen-detect.js
+++ b/src/win-fullscreen-detect.js
@@ -24,6 +24,25 @@ const FULLSCREEN_TOLERANCE_PX = 2;
 
 const MONITOR_DEFAULTTONEAREST = 2;
 
+// #719: the desktop itself passes the geometry test. Clicking the desktop
+// makes the shell window the foreground window — Progman, or one of
+// explorer's WorkerW siblings when a wallpaper host re-parents the icon view —
+// and its rect covers the whole monitor, so geometry alone classifies the
+// bare desktop as a fullscreen app. That made the hit window go non-activating
+// ~1s after any desktop click (startFocusablePoll), and Electron's
+// setFocusable(false) on Windows ends in Focus(false), which deactivates —
+// cancelling e.g. an in-place file rename on the desktop. Excluding the shell
+// window classes keeps "fullscreen" meaning an actual app.
+const DESKTOP_SHELL_WINDOW_CLASSES = new Set(["progman", "workerw"]);
+// Ample for real class names; matches win-foreground-terminal.js.
+const CLASS_NAME_BUF_LEN = 256;
+
+// Win32 class-name comparison is case-insensitive.
+function isDesktopShellWindowClass(className) {
+  return typeof className === "string"
+    && DESKTOP_SHELL_WINDOW_CLASSES.has(className.toLowerCase());
+}
+
 // Pure geometry: does the window rect cover the entire monitor rect (not just
 // the work area)? Exported so the decision logic is unit-testable without FFI.
 function rectCoversMonitor(winRect, monitorRect, tolerance = FULLSCREEN_TOLERANCE_PX) {
@@ -48,6 +67,7 @@ function createForegroundFullscreenProbe(options = {}) {
   let GetWindowRect;
   let MonitorFromWindow;
   let GetMonitorInfoW;
+  let GetClassNameW;
   let monitorInfoSize;
   try {
     const koffi = options.koffi || require("koffi");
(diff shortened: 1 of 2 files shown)


rullerzhou-afk:
Thank you for the careful investigation and focused fix! We independently verified the Windows desktop rename regression on real hardware, exercised the real Win32/Koffi path, and confirmed that fullscreen behavior remains intact. The targeted Progman/WorkerW exclusion and regression coverage look solid. We’re merging this now — thanks again for the thorough work.

PR #723, opened 2026-07-21
fix: keep tray flash icon at a stable size (#722)

### LINKED ISSUE #722: Menu bar tray icon turns into a large orange dot on new message (shifts icon position) — please add an off switch or a lighter indicator
## Description

When a new message arrives, the menu bar tray icon is replaced by a large solid orange dot instead of showing a subtle badge. This is visually heavy and, because the dot is wider than the normal icon, it also shifts the icon's position in the menu bar (neighboring items jump around).

## Steps to reproduce

1. Run clawd-on-desk with the menu bar tray icon enabled.
2. Trigger / receive a new message notification.
3. Look at the menu bar tray icon.

## Expected

A lightweight indicator that keeps the tray icon recognizable and its width stable — e.g. just the unread count, or a small dot badge overlaid on the existing icon.

## Actual

The whole icon turns into a big orange circle, and the icon position shifts.

## Request

Either of these would work for me:

1. **Add a setting to turn the new-message indicator off.**
2. **Fix it and use a lighter-weight indicator** — showing only the unread count would be enough. Please also keep the template image at a fixed width so the menu bar doesn't reflow.

## Environment

- macOS (Apple silicon)
- clawd-on-desk: latest

## Screen recording

https://github.com/user-attachments/assets/4aeaff1b-c1b5-4033-9c36-538b607c13d0

chen86860:
Fixes #722.

## Problem

When a task completes, `flashTaskbar()` blinks the tray icon between the normal icon and `assets/tray-icon-flash.png`. On macOS the two frames were not the same size:

| frame | asset | logical size |
| --- | --- | --- |
| normal | `tray-iconTemplate.png` (16×16) + `@2x` (32×32) | **16pt** |
| highlight | `tray-icon-flash.png` (32×32), loaded via `.resize({width: 32, height: 32})` at 1x | **32pt** |

The menu bar lays out in points, so the highlight rendered twice as wide as the normal icon. Every blink resized the icon and pushed the neighbouring status items sideways — the "big orange dot that shifts the icon position" in the issue.

## Fix

New `src/tray-flash-icon.js` owns tray icon loading:

- **macOS** — the 32px highlight is added as the `@2x` representation of a 16pt image (`addRepresentation({scaleFactor: 2, dataURL})`), so it occupies the exact same slot as the normal icon and stays crisp on retina. Falls back to a 16pt downscale if `addRepresentation` is unavailable.
- **Windows / Linux** — unchanged; those trays lay out in pixels and both assets are already normalised to 32×32.
- `menu.js`'s `createTray()` now uses the same normal-icon loader instead of its own copy, so the two paths can't drift apart again.

Asset paths stay on `path.join(__dirname, ...)`, and the artwork itself is untouched. The existing `flashTaskbarOnComplete` setting (Settings → General, default on) already covers the issue's "add an off switch" alternative, so no new preference was added.

## Verification

`AGENTS.md` asks for code-review-first notes on macOS-only paths and treats tray behaviour as manual-verification-first, so the status is spelled out below.

**Verified**

- `npm test` — 5625 tests, 0 failures (Node built-in runner).
- New `test/tray-flash-icon.test.js` covers the macOS `@2x` path, the `addRepresentation` fallback, the non-mac pixel path, and a missing / unreadable asset.
- `nativeImage` sizing measured under this repo's actual Electron (41.10.2), not just asserted against a stub:
  ```
  normal:       16×16 pt, scaleFactors [1, 2]
  flash before: 32×32 pt          ← the reflow
  flash after:  16×16 pt, scaleFactors [2]
  ```
- Manual QA on macOS 26.5.2 (Apple silicon), running from source: the tray icon keeps a constant width while flashing and no longer shifts neighbouring menu bar items.

**Not verified**

- Windows and Linux were not run. Those branches are a pure refactor — the same `createFromPath(...).resize({width: 32, height: 32})` as before, just moved into the shared module — so behaviour there is unchanged by inspection, not by execution.
- The `addRepresentation` fallback branch (16pt downscale) has not been exercised on a real runtime; it only triggers if `addRepresentation` throws, which it does not on Electron 41.

**Residual risk**

Low. The change is confined to how the two tray images are constructed; the flash timing, the `flashTaskbarOnComplete` gate, and the stop-on-click behaviour are all untouched.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -121,6 +121,7 @@ const {
   getProportionalPixelSize,
 } = require("./size-utils");
 const { keepOutOfTaskbar } = require("./taskbar");
+const { loadTrayNormalIcon, loadTrayFlashIcon } = require("./tray-flash-icon");
 const createTopmostRuntime = require("./topmost-runtime");
 const { WIN_TOPMOST_LEVEL } = createTopmostRuntime;
 const createThemeFadeSequencer = require("./theme-fade-sequencer");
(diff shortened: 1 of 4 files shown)


rullerzhou-afk:
Thanks @chen86860! This was a focused, well-tested fix. We independently reproduced the nativeImage sizing and ran a macOS menu-bar smoke on the reviewed head; the tray width remained stable with no neighboring-item shift. Merged as 8f6fda8. Appreciate the careful implementation and clear verification notes!

PR #729, opened 2026-07-23
feat(hooks): wire the 12 remaining adapters into the shared pid cache (#634)

LinYsssss:
## What

The #634 resolver mechanism (lifecycle `resolve(ctx)` + namespaced v2 disk cache) already lives in `shared-process.js`, with `clawd-hook.js` as its first consumer. This wires the **12 not-yet-migrated adapters** into it: each now passes `{ namespace, sessionId, cacheCwd, lifecycle, cacheable }` at its existing `resolve()` / `deps.resolvePid()` call sites, mirroring the clawd-hook pattern.

There are **no resolver-core changes**. The returned shape remains field-compatible, so body building is untouched. The disk-cache behavior is Windows-only; the context path remains a no-op elsewhere.

## Per-adapter mapping

| Adapter | namespace | start | prompt | end | cacheable |
|---|---|---|---|---|---|
| cursor | `cursor-agent` | `sessionStart` | `beforeSubmitPrompt` | `sessionEnd` | `sid !== "default" && cwd` |
| copilot | `copilot-cli` | `sessionStart` | `userPromptSubmitted` | `sessionEnd` | same |
| codex | `codex` | `SessionStart` | `UserPromptSubmit` | — | `!== "codex:default"` (raw-id guard) |
| gemini | `gemini-cli` | `SessionStart` | `BeforeAgent` | `SessionEnd` | `!== "gemini:default"` |
| kimi | `kimi-cli` | `SessionStart` | `UserPromptSubmit` | `SessionEnd` | raw id ≠ default |
| qwen-code | `qwen-code` | `SessionStart` | `UserPromptSubmit` | `SessionEnd` | raw id truthy ≠ `"default"` |
| qoder / qoderwork | `qoder` / `qoderwork` | `SessionStart` | `UserPromptSubmit` | `SessionEnd` | raw id truthy ≠ `"default"` |
| codebuddy | `codebuddy` | `SessionStart` | `UserPromptSubmit` | `SessionEnd` | `sid !== "default"` |
| reasonix | `reasonix` | `SessionStart` | `UserPromptSubmit` | `SessionEnd` | raw id truthy ≠ `"default"` |
| antigravity | `antigravity-cli` | — (no start hook) | — | — | explicit non-default `conversationId` + `workspacePaths[0]` only |
| kiro | `kiro-cli` | `agentSpawn` | **none** | — | **always false** |

## Design notes

- **Raw-session-id guard:** eight adapters prefix their ids, so a Claude-style `!== "default"` check would let an `<agent>:default` fallback key one shared cache entry across id-less sessions. The guard keys off the raw payload id and also rejects a literal raw `"default"`.
- **Antigravity requires an explicit cache identity:** `normalizeSessionId()` retains the historic transcript-dirname fallback for state compatibility, but that fallback never keys the disk cache. The [official payload shape](https://antigravity.google/docs/hooks#common-input-fields) ends in `.system_generated/logs/transcript.jsonl`, so the immediate parent is the shared name `logs`, not a conversation identity. Only the explicit per-conversation UUID is cacheable.
- **Antigravity's cache cwd is the stable workspace:** `resolveCwd()` prefers per-tool `toolCall.args.Cwd`, which can drift within one conversation and fragment the cache. The cache key uses `workspacePaths[0]`; the event/body cwd keeps the existing `resolveCwd()` behavior.
- **Kiro degrades gracefully:** it has no stable session id, so `cacheable:false` keeps the existing per-event fresh snapshot rather than returning empty cache-only metadata.
- **Stop/agentStop remain `"event"`:** they are turn-completion events, not session teardown. **SessionEnd maps to `"end"`** everywhere that hook exists, including Qoder and QoderWork.
- The offline-contract Cursor row moves `beforeSubmitPrompt` to `preToolUse`: prompt is deliberately spawn-free and can no longer anchor the one-spawn vacuity guard. A separate subprocess assertion preserves Cursor's exact prompt stdout.

## Acceptance mapping

- **Cache hits do not spawn snapshots:** win32 subprocess rows run the real hook entrypoints as fresh processes against live v2 entries and assert zero snapshot attempts for Cursor, CodeBuddy, Reasonix, and explicit-id Antigravity. Test cache files are isolated under a per-test `TEMP` / `TMP`; the suite never seeds or deletes the developer's real cache.
- **Prompt/end miss paths remain spawn-free:** Cursor prompt and Qoder SessionEnd are asserted on cache misses. Qoder SessionEnd also deletes only its matching live entry.
- **Unstable identities degrade safely:** Kiro and literal/default-id cases ignore seeded entries and resolve fresh. Antigravity transcript fallback paths are explicitly non-cacheable.
- **No cross-session PID bleed:** all prefixed/default guards are pinned, and the Antigravity regression uses the official transcript shape where distinct conversations share the parent name `logs`.
- **Consumer contract:** every `createPidResolver` consumer must pass a cache context or remain in the explicit WorkBuddy exemption.

## Windows smoke (2026-07-31)

- **Real Qoder CLI + administrator `Win32_ProcessStartTrace`:** SessionStart produced one snapshot; subsequent prompt events and SessionEnd produced zero. Two concurrent sessions in the same cwd created distinct cache files, and ending either session deleted only its own entry.
- **Antigravity authenticated CLI:** Google OAuth token exchange failed before a session or hook could start, so this run does not claim an authenticated CLI-layer result.
- **Antigravity Windows process-chain fallback:** two isolated `agy.exe` parent processes ran the real hook against official-shape payloads. Each first `PreInvocation` produced one real PowerShell snapshot; `PostInvocation` and `Stop` produced zero. Same-cwd conversations produced distinct cache files, all hook exits were status 0, and Stop preserved `{"decision":"allow"}` stdout. This is a process-tree/cache validation, not a substitute for the blocked authenticated CLI smoke.

## Out of scope, tracked

`workbuddy-hook.js` is the only remaining bare `resolve()` consumer. It is the sole contract-test exemption and should move in a small follow-up. This PR intentionally leaves #634 open because WorkBuddy and the remaining authenticated/CLI acceptance work stay outstanding.

## Testing

- `node --test test/hook-pid-cache-contexts.test.js` — **22/22 passed** on the final PR head.
- Merge-state full suite against current `main` — no PR-only failures; the failure set matched clean `main` under the same Windows harness.
- Administrator ETW smoke as documented above.
- Environment: Windows 11 + Windows Terminal as default terminal.

Refs #634

### DIFF
--- a/hooks/antigravity-hook.js
+++ b/hooks/antigravity-hook.js
@@ -161,6 +161,25 @@ function resolveCwd(payload) {
   return "";
 }
 
+// #634: cross-process pid cache context for the shared resolver. Antigravity
+// has no session-start hook (earliest event is PreInvocation), so every event
+// uses the "event" lifecycle — first resolve populates the cache, later ones
+// hit it (zero snapshot spawns). cacheable compares against the exact
+// normalizeSessionId fallback so an id-less payload ("antigravity:default",
+// cf. #583) never keys a shared entry; the transcript-dirname fallback is a
+// real per-conversation id and stays cacheable.
+function pidCacheContext(payload) {
+  const sessionId = normalizeSessionId(payload && payload.conversationId, payload);
+  const cwd = resolveCwd(payload);
+  return {
+    namespace: "antigravity-cli",
+    sessionId,
+    cacheCwd: cwd,
+    lifecycle: "event",
+    cacheable: sessionId !== "antigravity:default" && !!cwd,
+  };
+}
+
 function normalizeToolInputValue(value, depth = 0) {
   if (depth > TOOL_INPUT_DEPTH_MAX) return null;
   if (Array.isArray(value)) {
@@ -403,7 +422,7 @@ async function sendHookEvent(payload, argvEvent, deps = {}) {
   const outLine = stdoutForEvent(hookName);
   const remote = !!env.CLAWD_REMOTE;
   const pidMeta = shouldResolvePid(hookName, env)
-    ? (deps.resolvePid ? deps.resolvePid() : undefined)
+    ? (deps.resolvePid ? deps.resolvePid(pidCacheContext(payload)) : undefined)
     : undefined;
   const body = buildStateBody(hookName, payload || {}, {
     remote,
--- a/hooks/antigravity-hook.js
+++ b/hooks/antigravity-hook.js
@@ -164,10 +164,11 @@ function resolveCwd(payload) {
 // #634: cross-process pid cache context for the shared resolver. Antigravity
 // has no session-start hook (earliest event is PreInvocation), so every event
 // uses the "event" lifecycle — first resolve populates the cache, later ones
-// hit it (zero snapshot spawns). cacheable compares against the exact
-// normalizeSessionId fallback so an id-less payload ("antigravity:default",
-// cf. #583) never keys a shared entry; the transcript-dirname fallback is a
-// real per-conversation id and stays cacheable.
+// hit it (zero snapshot spawns). Only an explicit conversationId may key the
+// cache. normalizeSessionId still keeps the historic transcript-dirname
+// fallback for the state body, but that dirname is not proven session-unique
+// (for example, multiple transcript files may share a `jetski` directory), so
+// using it on disk could pin one conversation to another's live terminal PID.
 //
 // The cache key deliberately does NOT use resolveCwd(): that helper prefers
 // toolCall.args.Cwd, which varies per tool call (subdirs, slash spelling) and
@@ -183,14 +184,19 @@ function stableWorkspaceCwd(payload) {
 }
 
 function pidCacheContext(payload) {
-  const sessionId = normalizeSessionId(payload && payload.conversationId, payload);
+  const conversationId = payload && payload.conversationId;
+  const rawConversationId = conversationId != null ? String(conversationId).trim() : "";
+  const sessionId = normalizeSessionId(conversationId, payload);
   const cacheCwd = stableWorkspaceCwd(payload);
   return {
     namespace: "antigravity-cli",
     sessionId,
     cacheCwd,
     lifecycle: "event",
-    cacheable: sessionId !== "antigravity:default" && !!cacheCwd,
+    cacheable: !!rawConversationId
+      && rawConversationId !== "default"
+      && rawConversationId !== "antigravity:default"
+      && !!cacheCwd,
   };
 }
 
--- a/hooks/antigravity-hook.js
+++ b/hooks/antigravity-hook.js
@@ -168,15 +168,29 @@ function resolveCwd(payload) {
 // normalizeSessionId fallback so an id-less payload ("antigravity:default",
 // cf. #583) never keys a shared entry; the transcript-dirname fallback is a
 // real per-conversation id and stays cacheable.
+//
+// The cache key deliberately does NOT use resolveCwd(): that helper prefers
+// toolCall.args.Cwd, which varies per tool call (subdirs, slash spelling) and
+// would fragment one conversation across many v2 cache files — each a miss
+// that can spawn another snapshot. The key needs a per-SESSION constant, so it
+// uses workspacePaths[0] only; the event/body cwd keeps resolveCwd() behavior.
+function stableWorkspaceCwd(payload) {
+  if (payload && Array.isArray(payload.workspacePaths)) {
+    const first = payload.workspacePaths.find((entry) => typeof entry === "string" && entry);
+    if (first) return first;
+  }
+  return "";
+}
+
 function pidCacheContext(payload) {
   const sessionId = normalizeSessionId(payload && payload.conversationId, payload);
-  const cwd = resolveCwd(payload);
+  const cacheCwd = stableWorkspaceCwd(payload);
   return {
     namespace: "antigravity-cli",
     sessionId,
-    cacheCwd: cwd,
+    cacheCwd,
     lifecycle: "event",
-    cacheable: sessionId !== "antigravity:default" && !!cwd,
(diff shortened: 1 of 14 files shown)


rullerzhou-afk (changes requested):
Thanks for putting this together. The overall direction looks right: reusing the shared PID cache should substantially reduce repeated PowerShell snapshots (and the associated terminal flash) on Windows. I checked the current head (`4c810c2`) and also tested it merged with the latest `main`.

I’m requesting changes for two issues that affect the cache lifecycle/benefit:

1. **Qoder and QoderWork map `SessionEnd` to `event` instead of `end`.**

   Both adapters’ `HOOK_MAP`s include `SessionEnd`, and both installers register it, but `EVENT_TO_LIFECYCLE` does not contain `SessionEnd`. The fallback therefore sends it through the `event` lifecycle. On a cache hit, that path touches and retains the cache instead of deleting it; on a miss, it may take a fresh PowerShell snapshot, even though the `end` contract is cache-only and never spawns.

   Could you please add `SessionEnd: "end"` to both mappings, update the comments/PR table that currently say Qoder has no end hook, and add regression coverage for:

   - the emitted lifecycle;
   - deleting an existing cache at `SessionEnd`;
   - not spawning on a `SessionEnd` cache miss.

2. **Antigravity uses a per-tool working directory as the cache key.**

   `pidCacheContext()` currently sets `cacheCwd` from `resolveCwd(payload)`, while `resolveCwd()` intentionally prefers `toolCall.args.Cwd` over `workspacePaths[0]`. That preference makes sense for the reported event/body `cwd`, but it is unstable as a session cache key. Within one conversation, calls from `D:\repo`, `D:\repo\src`, or even the spelling variant `D:/repo` produce different v2 cache files. Each new value is a cache miss and can launch another PowerShell snapshot, which reduces the main benefit of this PR and leaves extra caches for stale sweeping.

   Please decouple the two concepts: keep the existing event/body `cwd` behavior, but derive `cacheCwd` from a stable session/workspace value such as `workspacePaths[0]`. A regression test should vary `toolCall.args.Cwd` (including slash spelling) within one conversation and assert that the cache context remains stable.

A couple of non-blocking follow-ups:

- **WorkBuddy still calls bare `resolve()`.** The repository currently has 14 `createPidResolver` consumers, and WorkBuddy is a Windows/macOS desktop integration, so it remains in the Windows flash impact area. It is okay to handle this separately, but if it stays out of this PR, please track it explicitly and avoid closing the all-agent scope of #634 via `Fixes #634`.
- Qwen/Qoder/QoderWork/Reasonix use a truthiness-only raw session-ID guard, so the literal placeholder `"default"` is cacheable. Their current IDs appear to be real session IDs, so I do not consider this blocking, but aligning them with the other adapters’ `raw !== "default"` guard would be useful defensive hardening.
- It would be valuable to add a contract test asserting that every resolver consumer either passes a cache context or is listed as an explicit exemption. The current test locks the consumer count but does not catch a consumer remaining on the bare path.

For completeness, I ran both the targeted PID-cache/adapter tests and the full suite. The PR head passed (`5859 passed, 19 skipped, 0 failed`), and the latest `main` merged with this PR also passed (`5978 passed, 20 skipped, 0 failed`). The merge itself is clean. Those results are encouraging; the two issues above are uncovered scenarios rather than existing test failures.

This remains a Windows-specific optimization. On macOS/Linux the resolver exits to fresh process metadata before loading the disk-cache path, so there is no analogous terminal-flash problem.

LinYsssss:
Both blocking items are fixed in 0ef072a, plus the three follow-ups.

**1. Qoder/QoderWork SessionEnd.** Confirmed — both HOOK_MAPs carry `SessionEnd` past the permission-mapping comment block (my verification pass truncated the map read there), and both installers register it. Both adapters now map `SessionEnd: "end"`; the "no SessionEnd hook" comments and the PR table are corrected. Coverage, per your three points: lifecycle-emit units for both adapters assert `SessionEnd → "end"`; a win32 subprocess run of the real qoder hook shows a seeded live v2 entry is **deleted** at SessionEnd with **zero** spawn attempts; and SessionEnd on a cache miss stays spawn-free.

**2. Antigravity cache key.** Decoupled as requested: the cache key now uses `workspacePaths[0]` (`stableWorkspaceCwd()`), while the event/body cwd keeps `resolveCwd()`'s per-tool behavior. Regression test: within one conversation, `toolCall.args.Cwd` varies across `D:/repo`, `D:\repo`, `D:/repo/src`, and absent — `cacheCwd` stays byte-stable at `workspacePaths[0]` and cacheable stays true.

**Follow-ups.**
- Literal `"default"` raw ids are now rejected by qwen/qoder/qoderwork/reasonix (units, plus a reasonix subprocess row that ignores a seeded `reasonix:default` entry and keeps its per-event fresh snapshot).
- Consumer contract test added: every `createPidResolver` consumer must build a cache context or sit in an explicit exemption list; `workbuddy-hook.js` is the sole exemption with a pointer back to the issue, and a second assertion fails if an exempted file stops being a consumer. The PR body now says **Refs #634** (not Fixes) so merging leaves the workbuddy remainder open — I can send that migration as a small follow-up PR once this lands.

**Local runs.** The new file is 19/19; full suite on this machine is 5864 pass / 19 skipped / 1 fail — the known local `antigravity-install` node-path failure that reproduces identically on clean `main` (nvm-junction environment). One observation from getting there: an installed Clawd (auto-restarting) squats port 23334 and deadlocks `mobile-preview-server.test.js` against the fixture server — worth knowing for anyone running the suite on a machine with the app installed; a test-time port override might be a cheap future hardening.

Marking ready for review.

rullerzhou-afk (approved):
Re-reviewed current head 4fb64b3c after the Antigravity identity fix, isolated cache tests, and Windows ETW smoke. The previous requested changes are resolved.

PR #731, opened 2026-07-23
fix: localize settings and tutorial window titles

YOIMIYA66:
## Summary
- Localize the native Settings and onboarding tutorial window titles for all supported UI languages.
- Keep BrowserWindow titles and renderer document titles synchronized during initial load and live language changes.
- Preserve the existing language preference flow and window lifecycle behavior.

## Problem context
- Settings always used the fixed native title `Clawd Settings`.
- The tutorial supplied a translated BrowserWindow title, but `tutorial.html` later replaced it with the fixed document title `Getting started`.
- As a result, Windows title bars stayed in English while the rest of the UI was displayed in Chinese, Traditional Chinese, Korean, or Japanese.

## What changed
- Added localized Settings window-title strings for English, Simplified Chinese, Traditional Chinese, Korean, and Japanese.
- Made the Settings window resolve its title from the current language when it is created and reapply it after page load.
- Added a language side effect that refreshes open Settings and tutorial native titles without recreating either window.
- Updated both renderers to set `document.title` from their current localized state so page-title updates cannot restore the fixed English HTML title.
- Extended Settings window, tutorial, effect-router, and locale-parity tests for initial titles and live language changes.

## Behavior after this PR
- Settings opens with a title matching the saved UI language, such as `Clawd 设置` for Simplified Chinese.
- The tutorial title uses the selected language instead of reverting to `Getting started` after its HTML loads.
- Changing language while either window is open updates its title immediately.
- Language persistence, Settings state updates, tutorial navigation, and window reuse behavior remain unchanged.

## Validation
- `node --test test/settings-window.test.js test/tutorial.test.js test/settings-effect-router.test.js test/i18n.test.js`
- `npm test` (`fail 0`, `cancelled 0`, 18 conditionally skipped)
- `git diff --check`
- Manual Windows check with Simplified Chinese preferences confirmed the loaded native Settings title is `Clawd 设置`.

## Platform note
- Automated verification ran on Windows with Node.js and Electron 41.10.2.
- The native Settings title was manually verified through the loaded Electron window. Manual cross-platform title-bar QA remains useful because native chrome rendering differs by OS.


rullerzhou-afk (changes requested):
Thanks for the careful work here — the overall direction looks good, and the initial-load/title handling is thoughtfully put together.

There is one reachable localization gap I think we should close before merging:

When the tutorial is already open and the language changes through Settings, the tray menu, or another settings-controller entry point, the centralized language effect only updates the tutorial's native BrowserWindow title. The tutorial renderer does not receive a fresh `tutorial:state`, so its `STATE.lang` / `STATE.i18n`, `document.title`, body copy, and buttons all remain in the previous language while the native title shows the new one.

This is easy to reach by opening “Show getting started guide” from Settings and then changing the Settings language. Could you please broadcast the updated tutorial state from the centralized language effect (or add an equivalent localization-sync method), and cover the external language-change path in a test? The test should assert the new tutorial state/dictionary as well as the native title, rather than only exercising `tutorial:set-lang`.

One small implementation detail: simply adding `_tutorial.sendState()` to `syncWindowTitles` will make the tutorial's own language picker send state twice, because `applyUpdate("lang", ...)` synchronously triggers the effect before the existing `tutorial:set-lang` handler sends again. It would be nice to centralize or otherwise avoid that duplicate if practical.

Also, as a non-blocking test improvement, the Settings fake BrowserWindow currently has no `webContents`, so the production-critical `did-finish-load` title reapply is not exercised. If convenient, please add a focused assertion for that callback while touching these tests.

Everything else in the change looks good. The targeted tests and the full suite both pass locally. Thanks again!

rullerzhou-afk (approved):
Re-reviewed exact head 6fba65e. The previous localization-sync gap is addressed: external language changes now refresh the tutorial's native title and renderer state, while the tutorial picker avoids duplicate state pushes. The focused tests and the full suite pass on the current main merge result, and the native Settings/tutorial titles were also verified in a running Electron build. Approved — thank you!

PR #735, opened 2026-07-24
feat(agents): add TRAE Work support

lauiiiiii:


rullerzhou-afk:
Thanks for adding TRAE Work support. I checked the current official TRAE Work package: it does contain command/HTTP hook machinery and event names such as `SessionStart`, `UserPromptSubmit`, `PreToolUse`, `PostToolUse`, `Stop`, and `Notification`, so the overall direction is viable.

However, this PR currently only registers an agent descriptor. In Clawd, `eventMap` is descriptive metadata; it is not used to translate incoming events at runtime. The `/state` route still expects a normalized payload containing a concrete `state`, `agent_id`, and `session_id`. As a result, a raw TRAE hook cannot update Clawd through the code in this PR alone.

Could you please complete the integration and include real-machine smoke evidence in the PR description?

1. Add the actual TRAE-to-Clawd event path: either a TRAE command-hook sender or a dedicated raw-hook adapter. It should read the active port from `~/.clawd/runtime.json`, map TRAE's raw `hook_event_name` to Clawd states, send `agent_id: "trae-work"`, and provide a stable session ID. Please do not hard-code port 23333.
2. Add install/uninstall and startup-sync support for `.trae/hooks.json`. The installer must merge and remove only Clawd-owned entries, preserving all existing user hooks and settings. Please also wire the Settings install/enable state and Doctor checks consistently with the other managed agents.
3. Complete the descriptor and automated coverage. On the current exact head (`6ef4979`), the targeted agent/registry tests report 4 failures: missing `processNames.mac`, missing required capability fields (including `permissionApproval`), an unchanged registry expectation, and missing explicit `startupRecoveryProcessNames`. Please only claim platforms/processes actually verified. In particular, do not match bare `Electron` on macOS; the current Work bundle's main executable is generic and would create false positives.
4. Keep the first version state-only unless TRAE permission-decision behavior is separately verified; set permission/interactive capabilities accordingly and leave approval in TRAE's native UI.

For the real-machine smoke result, please include the TRAE Work version, OS, and account/build context; whether Settings → Hooks is visible; redacted raw payload samples; the observed Clawd transitions for prompt/tool/stop/notification; dynamic-port behavior; restart and disable/re-enable behavior; install/uninstall results; and confirmation that pre-existing user hooks remain unchanged. If the Hooks UI is feature-gated or unavailable for the tested account, please state that explicitly.

Once the implementation and this evidence are available, I can review the complete integration path again.

PR #738, opened 2026-07-24
feat(discord): opt-in Clawd state-animation mirror in Rich Presence

KaiC5504:
Follow-up to #215, as invited in its closing comment — proposed there after verifying v0.12.0, split out here so it can be reviewed independently.

Extends the opt-in Rich Presence (#508) with a Clawd state-animation mirror: `large_image` follows the pet's current state animation instead of the static coarse-state icon — the crab types while working (or builds/grooves with the session-count tiers), blows bubbles or reads a book during the idle rotation, dozes off through the sleep sequence, crab-walks while roaming. **The mirror is its own opt-in, off by default** — the existing toggle keeps exactly the coarse/static behavior it always promised.

Idle
Thinking
Working — typing
Working — another sprite
Sleeping

## How it works

- Same mechanism v0.12.0 already uses — an external image URL. Discord animates external GIF URLs in `large_image` (uploaded portal assets render static), so no Developer Portal changes and nothing for BYO-app users to upload.
- Main funnels the pet's state animations through one `state-change` send (applyState plus the tick.js idle rotation); a small tap there feeds the presence bridge the current (state, svg), and the bridge maps clawd svgs to hosted GIFs. Click/drag reactions, low-power pauses and tint/accessory changes ride other channels and are deliberately out of scope. Zero new dependencies; the existing 4s SET_ACTIVITY coalescing absorbs rapid flips.
- The activity text stays session-derived: a mirrored visual only selects the image, so a transient pet animation (e.g. `mini-enter`) never captions an active working session as "Idle", and done/interrupted still recover through the badge fields.
- Because some animations are finer-grained than the coarse-state contract (working tiers reveal live session counts; mini/sleep/error variants are more specific), the mirror sits behind its own "Mirror the pet's animation" switch, whose description discloses this and the initial Clawd-only scope. Unknown svgs (calico, cloudling, custom themes) fall back to the coarse-state sprites, so nothing breaks for non-clawd themes.
- 21 sprites (~3.2 MB) under `assets/discord-presence/`, generated from the existing `assets/gif/` pet art by a dev-only script in `tools/` (crop to content, center, upscale — the raw sprites otherwise float tiny on the card). A manifest-consistency test keeps the runtime map, the generator's FILES list and the committed assets in lockstep.
- `GIF_BASE_URL` points at this repo's `main`, the same durable pattern as `CLAWD_ICON_URL`. That means the images 404 until this merges — the screenshots above are from the live-verified build with sprites temporarily served from my fork.
- Side effect worth having: finished/interrupted turns no longer show a plain idle image — the pet's happy/error one-shots now reach Discord.

## Scope

Per review: the full mirror is kept, gated behind the separate off-by-default opt-in. Per-theme sprite sets remain a possible data-only extension later (closer to what boomho sketched in #215).

## Review follow-up (2026-08-01)

All review items addressed in 3959554:

- Separate off-by-default opt-in (`mirrorPetAnimation`) with a Settings row and disclosure copy in all five locales; the existing toggle keeps the coarse/static behavior.
- `onVisual()` now stores the visual and defers its reconcile to a microtask, so a synchronous `onSnapshot()` from the same session update supplies the current session before anything is sent — a leading-edge send can no longer pair a new animation with the previous agent/project. Standalone idle-rotation changes still publish. The regression test covers old snapshot → new visual → new snapshot and was verified red against the synchronous version.
- Activity text is session-derived only; `VISUAL_LABEL` is gone.
- Dead `clawd-conducting.gif`, its mapping and its generator entry removed (the clawd theme remaps that display hint to juggling).
- Manifest-consistency test added; "every visual / exact mirror" wording tightened to the state-animation scope in code comments and this description.

## Testing

- `test/discord-presence-rpc.test.js` + `test/discord-presence-settings.test.js`: 31/31 — svg→gif mapping, opt-in gating (mirror off by default), session-derived text alongside a mirrored image, the visual/snapshot ordering regression, standalone idle-rotation publish, non-clawd fallback, mirror-with-no-session, project-name opt-in composition, manifest consistency, and a live publish over the mocked IPC socket.
- Full suite: no new failures (6113 tests; the remote-ssh/installer reds on Windows are pre-existing — identical set with this branch's changes reverted).
- Live-verified on Windows 11 + Discord stable, including the new toggle: off → the coarse/static presence, on → mirrored animations.


rullerzhou-afk:
Thanks for putting this together — I genuinely like this direction. The screenshots make a strong case for it: seeing the crab type, build, read, blow bubbles, sleep, and roam makes Rich Presence feel much more alive and much more like Clawd.

The implementation is also thoughtfully scoped: no new runtime dependency, the generated GIFs are reproducible, non-Clawd themes do not hit dead URLs, and the focused test coverage is solid. I verified that the 20 Discord Presence tests pass, the GIF generator reproduces the committed assets cleanly, and the PR does not add any new failures to the current full-suite baseline.

After a detailed review, I think there are two things we should address before merging.

### 1. The full mirror is a richer disclosure than the existing Discord opt-in describes

The current Settings copy promises that only the agent type and a coarse state — idle, thinking, working, or waiting — are shared.

The exact animation can reveal more than that. For example, the Clawd working tiers distinguish 1 / 2 / 3+ live sessions, the juggling tiers expose another concurrency level, and mini, sleep, error, idle variants, and display hints are all more specific than the existing coarse-state contract.

I do not think this makes the feature a bad idea — quite the opposite, I would like to keep the full mirror. But I think it should be a separate opt-in, off by default, rather than silently expanding what existing Discord Presence users already agreed to share.

Something like:

```text
[ ] Mirror the pet's animation on Discord
```

The existing toggle could retain the coarse/static behavior, while users who want the richer and more playful presence can explicitly enable the animation mirror. The alternative would be the proposed seven coarse-state GIFs, but I personally prefer preserving the full version behind its own opt-in.

### 2. `onVisual()` can publish a new visual with the previous session

The normal update order is currently:

```text
session changes
→ state-change / onVisual()
→ emitSessionSnapshot() / onSnapshot()
```

`onVisual()` publishes immediately using `lastSnapshot`. If the previous Discord send was more than four seconds ago, this can send a leading-edge activity containing the new animation but the old agent/project. The correct snapshot then arrives immediately, but the existing four-second coalescing delays the correction.

For example, a transition from Codex in `old-project` to Claude Code in `new-project` can briefly publish:

```text
Codex / Working · old-project
```

with the new Claude visual, before correcting a few seconds later.

I think the cleanest fix is for `onVisual()` to store the visual and defer its reconcile to a microtask. A synchronous `onSnapshot()` from the same session update can then supply the current session before anything is sent, while standalone idle-rotation visual changes will still publish normally.

It would be good to add a regression test covering:

```text
old snapshot → new visual → new snapshot
```

### A few smaller scope/cleanup notes

- A known visual currently controls both `large_image` and the activity text. That means an active working session can temporarily say `Idle` during `mini-enter`. I suggest keeping the text session-derived and mirroring only the image; done/interrupted can still be recovered through the existing badge fields.
- “Every visual” / “exact mirror” is slightly broader than what is implemented. Click and drag reactions, low-power pauses, tint/accessory appearance, and roam direction do not travel through `state-change`. “Clawd state-animation mirror” would describe the current scope more precisely.
- Calico, Cloudling, and custom themes fall back to Clawd GIFs, so the UI or copy should make the initial Clawd-only scope clear.
- `clawd-working-conducting.svg` is not reachable in the current Clawd theme—the display hint is remapped to juggling—so `clawd-conducting.gif` and its mapping appear to be dead assets.
- A small manifest-consistency test covering runtime GIF targets, generator `FILES`, and committed assets would help prevent future drift.

The overall direction is a strong yes from me. This is a fun, very on-brand extension, and the asset work and live verification are appreciated. Once the privacy opt-in and visual/snapshot ordering are addressed, I would be happy to approve it.

rullerzhou-afk:
Thank you, @KaiC5504 — this is a thoughtful feature, and the separate opt-in/privacy boundary made it possible to ship safely. 🙌

Before merging, we pushed a small maintainer follow-up after a deep review and a real Windows + Discord smoke test. It:

- keeps the existing coarse activity-text contract while the opt-in mirrors only the pet image, so a transient error GIF cannot leave a public/sticky Error label;
- reconciles same-tick visual/session updates in both event orders and leaves a little more RPC rate-limit headroom;
- hardens the Clawd-only sprite lookup and the development GIF URL override;
- merges the current main branch (#792) and adds regression coverage.

The live smoke covered working → completion/idle plus an injected failure: the GIF stayed in sync, and the error visual correctly kept the activity text as Working. Local cross-feature tests passed 520/520, and all GitHub audit, unit, and package checks are green.

The core implementation and assets remain your contribution — thanks again for building this! We're merging it now.

KaiC5504:
Thank you for the thorough review and the follow-up hardening — keeping the activity text coarse and image-only is a cleaner boundary than what I had, and the extra smoke testing is much appreciated. Glad the crab is livelier on Discord now. Thanks for merging! 🦀

PR #739, opened 2026-07-24
fix: align tutorial language picker with settings

YOIMIYA66:
## Summary

- Replaces the onboarding tutorial's native language `<select>` with the same custom picker presentation used by Settings.
- Centralizes picker interaction, accessibility state, async rollback, and listener cleanup in a shared browser component.
- Moves the welcome content upward so all language options remain visible above the fixed tutorial footer.
- Preserves the supported locale list, native language labels, language persistence, `tutorial:set-lang` IPC, and localized window-title updates.

## Problem context

The tutorial used a platform-native selector while Settings used a custom dropdown. This made the first-run experience visually inconsistent and left the two controls with different keyboard and close behavior.

Copying the Settings implementation into the tutorial would duplicate document listeners, focus handling, and save rollback logic. After sharing the dropdown, its taller menu also exposed a layout issue: the vertically centered welcome content placed the lower options behind the fixed footer at the tested Windows window size.

## What changed

- Added a shared `createLanguagePicker({ value, options, ariaLabel, onChange })` browser component.
  - Supports ArrowUp/ArrowDown, Enter, Space, Escape, and outside-click closing.
  - Maintains `aria-expanded`, `aria-hidden`, and `aria-selected` state.
  - Rolls back optimistic selection when a callback fails.
  - Exposes `setValue()` and `dispose()` so rerenders do not retain document listeners.
- Moved the existing Settings picker styling into a shared stylesheet with Settings and tutorial theme-variable fallbacks.
- Updated Settings to use the shared component while keeping its existing `settingsAPI.update("lang", ...)`, error Toast, and rollback behavior.
- Updated the tutorial welcome step to use the shared component with native language names and the existing `tutorial:set-lang` path.
- Disposes the tutorial picker before each wizard rerender.
- Shifted the complete welcome block upward by 56px so the five-option menu clears the footer without resizing the window or changing menu dimensions.
- Expanded browser-environment and tutorial tests for shared asset loading, keyboard selection, save failure rollback, outside-click/Escape closing, listener cleanup, and the welcome-page offset.

## Behavior after this PR

- Settings and onboarding present the same language picker styling and selection states.
- Mouse and keyboard users can open, navigate, select, and dismiss both pickers consistently.
- Failed Settings language saves restore the committed language and retain the existing error feedback.
- Repeated Settings or tutorial renders release picker-level document listeners.
- Tutorial language persistence, supported locales, localized content, and localized window titles behave as before.
- Existing `.language-picker` and `.language-picker.open` class names are retained, so the overflow handling proposed separately in PR #737 remains compatible.

## Validation

- `node --check src/language-picker.js`
- `node --check src/settings-tab-general.js`
- `node --check src/tutorial-renderer.js`
- `node --test test/settings-renderer-browser-env.test.js test/tutorial.test.js test/i18n.test.js`
  - 197 tests passed.
- `node --test test/tutorial.test.js test/settings-renderer-browser-env.test.js`
  - 188 tests passed after the welcome-layout adjustment.
- `git diff --check origin/main...HEAD`
- `npm test`
  - Branch result: 6,072 passed, 13 failed, 18 skipped.
  - The same 13 tests fail on clean `origin/main` with 6,070 passed and 18 skipped.
  - Baseline failures are confined to Remote SSH/POSIX assumptions on Windows: CRLF-only shell-script assertions, POSIX remote-home validation, and exact `0600` mode checks in `remote-deploy.test.js`, `remote-ssh-deploy.test.js`, `remote-ssh-identity.test.js`, `remote-ssh-path-isolation.test.js`, and `server-config.test.js`.

## Platform note

- Automated verification ran on Windows x64 with Node.js v22.22.2.
- Manual Windows testing identified the footer overlap shown by the shared dropdown; the final layout moves the welcome block upward to keep the menu visible. A final light/dark and 125%/150% display-scaling pass is still recommended.

## Scope control

- This PR only aligns and shares the tutorial language picker and adjusts its welcome-page layout.
- It does not change locale persistence, locale availability, tutorial navigation, Settings information architecture, or window-title localization.
- Settings card overflow/clipping remains the separate scope of PR #737.


rullerzhou-afk:
Thanks for the thoughtful work here. The shared picker direction is good: the component extraction is clean, the Settings persistence path remains intact, the CSP and asset ordering are correct, and listener disposal is handled carefully.

We independently reviewed the current exact head `8b530809864d0b4a499a48881d20aef880ea0261`. Before merging, we found two functional issues that should be addressed.

### 1. Failed concurrent saves can roll back to an uncommitted value

In `src/language-picker.js`, `previous` is taken from the current optimistic `activeValue`, rather than the last committed value.

Reproduction:

1. The committed language is English.
2. Select Chinese while save #1 remains pending.
3. Select Japanese while save #2 remains pending.
4. Make both saves fail.
5. Save #1 is correctly ignored as stale, but save #2 rolls the picker back to Chinese.

The final UI shows Chinese even though the persisted language is still English.

Please keep a committed baseline that advances only after a successful save, or let the Settings caller explicitly restore `readers.getLang()` on failure. A behavioral test with two in-flight failed changes should be added.

### 2. The tutorial’s page-contained menu is clipped at small/high-zoom layouts

The old native `<select>` popup was rendered outside the page, but the new menu is absolutely positioned inside:

```css
.body.step-welcome {
  overflow: hidden;
}
```

Moving the welcome block from `-4px` to `-56px` helps at the default 100% layout, but it does not solve the collision:

- At the minimum window size and 100%, the last locale is already partially clipped.
- At 125%–150%, one or more locales become unreachable.
- At higher zoom, the upward translation also begins clipping the hero at the top.
- Because the welcome body cannot scroll, the user has no recovery path.

Simply changing the container to `overflow: visible` is not sufficient: the menu can escape the window or be covered by the fixed footer.

Please make the shared picker collision-aware—preferably flip the menu upward when there is insufficient space, with a bounded `max-height` and internal scrolling as a fallback. Once placement is adaptive, the welcome translation can return near its original position.

Please also add a behavioral layout test covering the minimum tutorial size and enlarged text scale. The current tutorial tests mostly assert source strings and stub the picker, so they do not exercise this interaction.

Non-blocking accessibility note: after selection, the picker correctly restores focus to its trigger, but the tutorial immediately re-renders and removes that focused node, leaving focus on the body. The custom picker also no longer has the native select’s type-ahead and Home/End behavior.

These findings are not intended to create an obligation for you. If you would prefer, we can take over and finish the two required fixes. Thanks again—the overall component-sharing approach is solid.

rullerzhou-afk:
Thanks for following up. I re-reviewed the current exact head 1fd5696d50629d7541e7657743cdd95a437090eb.

The concurrent rollback issue is resolved: the committed baseline now advances only after success, and the new behavioral cases pass. The adaptive menu also flips and stays fully visible at the default layout and at the minimum window with 125% text scale in real Electron.

One blocking layout edge remains before merge:

- At the minimum 560x600 BrowserWindow and 150% text scale, the visible tutorial body ends at y=313.7 while the language trigger starts at y=317.1. The trigger is entirely behind the fixed footer.
- At 160%, which is the project TEXT_SCALE_MAX, the body ends at y=290.3 while the trigger spans y=316.6 to 352.6.
- The welcome body can technically scroll now, but its scrollbar is explicitly hidden. The closed welcome screen therefore shows no language control and no indication that more content exists. Wheel scrolling or keyboard focus can recover it, but a pointer user cannot discover or access the control initially.

The current minimum-size enlarged-text test hardcodes the trigger inside the boundary, so it cannot catch this real layout. Please keep the trigger initially visible at 560x600 for 150% and 160%, or provide an explicit visible scrolling affordance or initial positioning strategy. The behavioral test should assert that the trigger is inside the visible body before opening, then that the first and last options remain inside it after opening.

Everything else from the previous blocking feedback looks addressed. Thanks again for the careful follow-up.

rullerzhou-afk:
Thank you so much @YOIMIYA66 for this contribution and for the patient follow-up. The core design and implementation here—the shared picker, Settings/tutorial integration, async rollback handling, adaptive placement, and the accompanying tests—are a real improvement. We took over only the final small resize and accessibility polish to avoid putting you through another review round. After a full re-review and real Electron validation, the PR is now merged. We really appreciate the care you put into this!

PR #746, opened 2026-07-25
fix: replace native integration cleanup confirmation

YOIMIYA66:
## Summary
- Replace the About page's native integration-cleanup confirmation with the shared Settings confirmation modal.
- Keep the existing cleanup command, progress state, result summary, and success/failure toasts unchanged.
- Add localized confirmation copy and focused coverage for the destructive-action gate.

## Problem context
The **About > Remove Clawd integrations** action used `window.confirm()`, so it opened an operating-system dialog that did not match the rest of Settings. The warning was also stored as one combined string instead of using the title, detail, and action structure supported by the existing Settings modal.

## What changed
- Gate `cleanupIntegrations` behind `showSettingsConfirmModal()`.
- Use a danger-styled confirmation action and keep Cancel as the default-focused action.
- Treat cancellation and other non-confirm results as no-ops, while preventing duplicate prompts during an outstanding confirmation.
- Preserve the existing disabled/running button state and status/toast reporting while cleanup is in progress.
- Split the confirmation copy into title, detail, confirm, and cancel strings for English, Simplified Chinese, Traditional Chinese, Korean, and Japanese.
- Add renderer-level tests for cancellation, duplicate clicks, confirmed cleanup, success state restoration, and command failures.

## Behavior after this PR
- Removing Clawd integrations now uses the same in-app modal styling and dismissal behavior as other Settings confirmations.
- Cancel, Escape, backdrop dismissal, or closing without the danger action does not run cleanup.
- Confirming still invokes the existing `cleanupIntegrations` command exactly once.
- Cleanup continues to mark managed integrations as uninstalled and disabled while preserving the existing success/failure summaries and Kiro note behavior.
- No integration cleanup or persistence semantics are changed.

## Validation
- `node --test test/settings-about-confirmation.test.js test/settings-renderer-browser-env.test.js test/settings-actions.test.js test/cleanup-integrations.test.js test/i18n.test.js`
  - 374 passed, 0 failed
- `git diff --check`
- Manual Windows QA confirmed the About cleanup confirmation renders in-app and the cleanup flow behaves as expected.

## Platform note
Automated verification and manual QA were performed on Windows. The modal uses the existing cross-platform Settings renderer implementation.


rullerzhou-afk:
Thank you, @YOIMIYA66! This is a thoughtful polish for a destructive action: the shared Settings modal fits the product much better, Cancel remains the safe default, and the underlying cleanup semantics stay unchanged.

I reviewed the exact head, ran the focused tests and the full suite locally (6,104 passed, 0 failed, 18 skipped), and launched the branch in an isolated macOS instance to verify the About UI and cleanup entry. Everything looks good. Merging now — thanks again for the contribution!

PR #749, opened 2026-07-26
feat: add per-session permission automation

### LINKED ISSUE #608: Feature request: per-session auto-approve toggle
# Feature request: per-session auto-approve toggle

## The gap I hit

`autoApproveAllPermissions` is great when I trust everything, and the default per-request bubble is great when I want to sign off on each tool call. There is a middle case that isn't well-served today:

- I have **one specific session** I trust (e.g. a scoped refactor in a scratch worktree, an isolated exploration, a long-running agent I'm babysitting elsewhere)
- I still have **other sessions running in parallel** where I want to keep manual approval

Turning `autoApproveAllPermissions` on for the trusted session means every other session becomes YOLO too, which is more blast radius than I want. I end up either mashing Allow on the trusted session or turning global auto-approve on/off around it, both of which defeat the point.

## What I'd like

A per-session variant of the same idea: whitelist a specific `sessionId` so that its future permission requests are auto-approved, without changing anything for other sessions. The whitelist should be **ephemeral** (dropped on Clawd restart), same safety envelope as `autoApproveAllPermissions` today — a "turn everything off" mode must never silently persist across restarts.

Rough shape I have in mind, so we're talking about the same thing:

- A new schema field `autoApproveSessionIds: { [sessionId]: true }`, `ephemeral: true`, gated behind a `setSessionAutoApprove` command in `settings-actions.js` (the pattern `setAutoApproveAll` already uses).
- `maybeAutoApprovePermission` opens the door if **either** the global switch is on **or** the current `sessionId` is on the whitelist.
- Two entry points to flip a session on:
  - A dashboard-card toggle next to the existing session actions (obvious, discoverable).
  - An "always approve for this session" tertiary button on the permission bubble itself (contextual, one click when you already know you trust the session you're looking at).
- Bubble button opts out for tool inputs where "always approve" wouldn't make sense — passive notifies (Codex/Kimi state-only), elicitation, `ExitPlanMode`, pure-metadata passthrough tools, hardware-buddy test permissions.

## Why open an issue first

Before I put a PR up, I want to check whether this direction fits how you think about the permission surface. Two things I could see going differently:

1. **Session vs. tool granularity.** You may prefer a tool-level allowlist ("always approve `Bash(git status)`") over session-level. Session-level is what fits my workflow, but I don't want to preempt your model if you have a clearer plan for tool-level.
2. **Confirmation gate.** `setAutoApproveAll` requires a `confirmed:true` on its command. My default assumption for the per-session variant was to skip that gate — the blast radius is one session, both entry points already read as explicit user actions (dashboard toggle, contextual bubble button), and `ephemeral: true` keeps it from surviving a restart. But you might want the same confirmation behavior for consistency.

Happy to defer to your call on both.

## Environment

- macOS 15.x, Clawd `0.10.0` (main @ 02a46b2), Claude Code as the primary agent.
- I have a working local implementation already (17 files, tests green under `npm test`, en/zh/zh-TW/ko/ja i18n). Not opening the PR yet — waiting on your read on this issue.

Thanks for the project.

rullerzhou-afk:
## Summary

- add an in-memory per-session automation store and coordinator with Follow global, Ask every time, and Auto allow tool requests modes
- expose session controls in Dashboard and on eligible permission bubbles, including confirmation, pending-request sweeps, orphan cleanup, and lifecycle revocation
- extend Telegram and Feishu approval cards with two-step session trust and exact-grant remote revocation
- fail closed on identity and live-route checks; this first audited path enables session automation only for local, process-bound Codex CLI sessions, while questions, plan reviews, passive notifications, headless/subagent/remote sessions, and placeholder identities remain manual
- merge the latest `main` Telegram native-only architecture without restoring the retired sidecar
- parent Dashboard-triggered warnings to the invoking Dashboard window on Windows, and keep the native session selector stable across one-second Dashboard refreshes

## Validation

- targeted session automation / remote approval suites: 195 passed, 0 failed
- focused Dashboard, IPC, coordinator, and dialog-parent regression suites after the Windows fixes: 49 passed, 0 failed
- `npm test` on Windows: 6322 total, 6289 passed, 20 skipped, 13 failed; the same 13 failures reproduce on current `main` in this checkout and are existing Windows CRLF / POSIX-path / Unix-mode assertions
- repository asset + Telegram retirement audit suites: 33 passed, 0 failed
- `git diff --check`

## Windows real-machine QA

Ran an isolated Electron instance with separate HOME, userData, runtime port, and global automation mode `off`.

- Dashboard session automation selector remains open and can select Auto allow tool requests
- the native warning receives keyboard focus without clicking it first
- pressing Enter activates the intentionally default Cancel action
- cancellation leaves the session on Follow global, creates no session grant, leaves global mode `off`, and does not persist warning dismissal

Real Telegram/Feishu service smoke was not rerun in this handoff.

Closes #608


rullerzhou-afk:
## Windows real-machine validation needed before merge

I plan to run a Windows smoke pass before merging this PR. The pass should cover:

- **Windows app compatibility:** launch Clawd, open Dashboard, render the per-session controls, handle multiple concurrent sessions, and confirm restart/session-end/agent-disable cleanup.
- **Local Codex CLI (feature path):** switch one eligible session between Follow global, Ask every time, and Auto allow tool requests; verify other sessions are unaffected; verify questions and plan reviews still wait for the user.
- **Claude Code (regression path):** verify its existing permission bubble and global automation behavior still work on Windows. Per-session automation is currently expected to remain unavailable for Claude until its session identity lifecycle is separately verified.
- **Telegram:** exercise the two-step session-trust confirmation, pending-request handling, exact-session isolation, persistent revoke action, and final card status updates.
- **Feishu:** exercise the same trust/confirm/revoke lifecycle through the real Windows app and real service connection.
- **Safety gates:** verify DND, disabled agent/permission switches, headless or subagent requests, stale requests, and app restart never silently auto-allow.

I will report the observed Windows results here, including any untested boundary or platform-specific issue.

PR #750, opened 2026-07-26
fix(feishu): resolve remote approval recipients by email

### LINKED ISSUE #683: Feishu/Lark 远程审批：审批人 open_id 拿不到，普通用户配不出来
# Feishu/Lark 远程审批：审批人 open_id 拿不到，普通用户配不出来

## 问题

配置飞书/Lark 审批渠道时，第 2 步要填「审批人 user id」。这一步实际上把绝大多数用户挡在门外。

证据不是推测——**我自己（本项目作者）在配置时就卡住了**：

- 我打开自己应用的开放平台页面，从头到尾找不到 open_id 在哪
- 原话："这个是啥？我怎么填，我怎么不记得上次填过这个"
- 最后是靠 API Explorer 调 `contact/v3/users/batch_get_id` 用邮箱换出来的

如果作者本人都找不到，普通用户不可能配得出来。

## 为什么这么难

1. **`open_id` 是按应用隔离的**——同一个人在不同应用下的 open_id 不同。所以它**不在 Lark/飞书客户端的个人设置里**，用户没法"看自己的 ID"。概念上就反直觉。
2. **只能从 API 拿**——要么用 API Explorer（独立工具，不在应用页里，还要先 Refresh Token、还要知道选哪个接口），要么自己写代码。
3. **现在的提示只有懂的人看得懂**：占位符「例如 ou_xxx」，说明「这里要和卡片回调里的用户 ID 类型一致」。这句话预设你已经知道什么是 open_id 和卡片回调。

同类前科：#493 里 RebornQ 卡住不是权限问题，是订阅了旧版 `card.action.trigger_v1` 回调——同样是"关键配置项藏得太深"。

## 方案（已定：C）

| | 做法 | 权限成本 |
|---|---|---|
| A | 设置页加可展开的「怎么拿 open_id」引导 + 深链到 API Explorer 的具体接口 | 0 |
| B | 让用户直接填**邮箱**，Clawd 调 `contact/v3/users/batch_get_id` 换成 open_id 存起来 | 要 `contact:user.id:readonly` |
| **C** | **A + B：默认填邮箱，解析失败就展开 A 的引导兜底** | 可选，不强制 |

选 C。B 的体验是碾压性的——用户只填自己邮箱，`open_id` 这个概念根本不用出现；但它要一个只读权限，用 A 兜底就不强制用户开。

## 已验证的实现依据

在真实 Lark 租户上实测过（2026-07-15）：

```
POST https://open.larksuite.com/open-apis/contact/v3/users/batch_get_id?user_id_type=open_id
Authorization: Bearer <tenant_access_token>
{"emails":["<用户邮箱>"]}

→ 200 {"code":0,"msg":"success","data":{"user_list":[{"email":"…","user_id":"ou_…"}]}}
```

- 用 `tenant_access_token` 即可，不需要 user token
- 需要 `contact:user.id:readonly`（应用没开时返回 `code=99991672`，错误信息会列出缺哪个 scope——可以拿来触发兜底引导）
- 同一接口在 `open.feishu.cn` 上也存在，两个平台通用

被拒的接口（说明不能靠这些绕过）：
- `contact/v3/scopes` → 需要 `contact:contact:*`
- `im/v1/messages/:id` → 需要 `im:message:readonly`
- `im/v1/chats` → 需要 `im:chat:readonly`

## 实现要点

- 输入框接受**邮箱或 open_id**：`ou_` 开头直接存，否则当邮箱去解析
- 解析失败要分清原因：`99991672`（缺权限）→ 展开引导；查无此人 → 提示邮箱不对
- 解析出来的 open_id 存进 prefs，**回调匹配逻辑完全不用改**（还是 open_id）
- 邮箱不落盘——只用来换 ID
- 引导里放的开放平台链接必须走 `escapeWithLink` 白名单（已支持 `open.feishu.cn` / `open.larksuite.com`）
- 现有的 `open_id` / `union_id` / `user_id` 手填路径保留

## 顺带修

**超时的测试卡片仍然挂着可点的按钮，点了静默无反应。** 60 秒超时后 pending 被清，但卡片没更新——用户看到的还是一张"可以点"的卡。应该更新成「已取消」。

## 范围

不塞进 #682。理由：#682 是平台适配（已 +2400 行、Lark 真机验完），而这个问题**和 Lark 无关**——飞书用户一样配不出来，是渠道自身的缺陷。#682 合了之后另开 PR。

Cobb04:
## Problem

Configuring Feishu/Lark remote approval currently requires users to obtain an
app-scoped open_id manually, which is difficult to discover and blocks normal
setup.

## Solution

- Accept an approver email in open_id mode and resolve it through
  `contact/v3/users/batch_get_id`.
- Keep `ou_...` values on the existing manual open_id path.
- Preserve the existing manual user_id and union_id paths.
- Persist only the resolved open_id, not the submitted email.
- Reuse persisted credentials only when they match the selected platform.
- Return stable, localized lookup errors without exposing raw API messages.
- Provide a Feishu/Lark API Explorer fallback guide when automatic lookup
  fails.
- Update unanswered test cards to a cancelled, non-actionable state after
  timeout.

## Security and privacy

- Submitted emails are used only for lookup and are not written to prefs.
- Transient App ID and App Secret values are not persisted.
- Successful lookup stores only the resolved open_id.
- Renderer responses contain stable status and error codes only.
- Diagnostic logging excludes secrets, tokens, emails, open_id values, and
  raw request bodies.

## Tests

- `git diff --check`
- JavaScript syntax checks for all changed production files
- `npm run verify:electron`
- Focused tests: 427 passed
- i18n tests: 9 passed
- Full suite: 6203 passed, 0 failed, 18 platform-specific skipped

## Manual validation

Validated with a personal Feishu test tenant:

- email-to-open_id lookup
- message-card delivery
- `card.action.trigger` approval callback
- test-card timeout cancellation

## Limitations

- Lark tenant behavior was not manually validated.
- Automatic email lookup requires `contact:user.id:readonly`.
- Users without that permission can use the API Explorer fallback and paste
  the resulting `ou_...` value manually.

Closes #683


rullerzhou-afk:
Thanks for this PR. I agree with the product direction: resolving an approver email through `batch_get_id`, persisting only the resulting ID, preserving the manual ID paths, returning stable localized errors, and providing a fallback guide are all the right shape for #683. I also verified that the lookup helper's own structured log contains only `stage`, `httpStatus`, and `businessCode`.

Before merge, however, I think we need to address three connected correctness/security issues in the same round. The common problem is that credential identity is currently not recorded, so neither persisted credentials nor values derived from them can be checked for provenance.

## 1. The SDK's default logger exposes App Secret and lookup email

`createLarkClient()` sets `loggerLevel`, but does not inject a logger. I reproduced this with the lockfile version, `@larksuiteoapi/node-sdk@1.71.1`, the production `lookupOpenIdByEmail()`, and real Axios error objects:

- a `batch_get_id` HTTP 500 prints `config.data` containing the submitted email;
- a `tenant_access_token` failure prints the request body containing the App ID and App Secret.

This happens before Clawd's sanitized `log()` callback runs. It goes to the main-process console rather than one of Clawd's debug log files, but it still violates the existing redaction boundary and can be copied into bug reports.

Changing only the level is not sufficient. I measured the following on 1.71.1:

| configuration | captured console bytes | leaked values |
| --- | ---: | --- |
| current `warn` | 25331 | App ID + App Secret |
| `fatal` | 25360 | App ID + App Secret |
| `error` | 25331 | App ID + App Secret |
| injected no-op logger | 0 | none |

`fatal` is especially misleading because `LoggerLevel.fatal === 0`, while the SDK constructs the proxy with `params.loggerLevel || LoggerLevel.info`, so it falls back to `info`.

Please inject an explicit logger into the REST `Client`. The `WSClient` and `EventDispatcher` accept the same logger interface, so I would inject it there as defense in depth as well; the confirmed leak is on the REST path.

Please also add a focused regression test using the real locked SDK and real Axios errors, capturing every `console.*` method and asserting that the output contains none of: email, App ID, App Secret, bearer token, Authorization header, or request body. The existing fake SDK tests are still useful for business logic, but structurally cannot exercise `LoggerProxy` or `formatErrors()`.

## 2. Persisted credentials and resolved `open_id` values are not bound to a platform/App

There are two reproducible failures here.

First, a transient lookup can produce a silently invalid configuration:

1. App B is currently persisted.
2. The user enters App A credentials in Replace mode without saving them.
3. Email lookup resolves `open_id(A)` and writes only that ID to prefs.
4. The user closes Settings, cancels, the secret save fails, or later saves App C.
5. Runtime combines the persisted credentials with `open_id(A)`, while `readiness()` still returns ready.

Second, the env file has no platform provenance. The check in `feishuApprovalResolveApprover()` compares the requested platform with `getFeishuApprovalPrefs().platform`, but both normally come from the same selected prefs value; it does not say which platform produced the credentials on disk.

Using the real command registry, production env reader/writer, production lookup helper, and SDK 1.71.1, I reproduced this sequence:

- env contains a Feishu App ID/Secret;
- prefs platform is switched to `lark` without resaving credentials;
- email lookup sends the Feishu App Secret to `open.larksuite.com`;
- the lookup succeeds and `readiness()` returns `ready: true`.

There is an additional SDK issue: the default cache is module-global and namespaces tenant tokens by App ID only, not by domain. With the same App ID, a fresh Lark client reused a token minted by `open.feishu.cn` and sent it to `open.larksuite.com` without fetching a Lark token.

I recommend the simpler persisted-only flow:

- Save Step 1 before Step 2 becomes available.
- `resolveApprover` reads only persisted credentials; remove transient App ID/Secret from this command.
- Store credential platform provenance atomically with the credentials, e.g. `FEISHU_CREDENTIAL_PLATFORM=feishu|lark` in `feishu-approval.env` (please avoid reusing `writeSecretsEnvFile()`'s existing `platform` argument, which means `process.platform`).
- Bind auto-resolved approvers to the credential identity that produced them: source=`lookup`, selected platform, and a non-secret App identity binding (full hash or an opaque credential-generation ID are both reasonable; exact representation is up to you).
- `resolveApprover`, `readiness()`, and runtime startup must fail closed on platform/App mismatch.
- Give each SDK client an isolated cache, or provide a cache namespace that includes domain + App identity. Do not put App Secret in a cache key.

One nuance: manual IDs should remain supported, but “manual” should not mean “valid forever under any credential.” A manually pasted `open_id` is still app-scoped. `user_id` is tenant-scoped and `union_id` is developer-scoped, so their invalidation rules are different. At minimum, a manual `open_id` must be re-confirmed or invalidated after the App/platform changes. Please make the transition rules explicit and test them.

For legacy env files without platform provenance, stamping the current prefs platform is a compatibility assumption, not proof of origin. The safest migration is to mark the credentials unbound and require one Step 1 resave. If preserving current behavior is preferred, please document/test that assumption and ensure it cannot silently bypass later mismatch checks.

## 3. Lookup can race full-object config updates

The lookup pending state does not fully gate Enable, timeout, and Test. Because the renderer saves complete `feishuApproval` objects, a lookup response and a concurrent config update can overwrite each other's fields.

Adding only `feishuApprovalResolveApprover.lockKey = "feishuApproval"` is not sufficient: normal config updates currently use `update:feishuApproval`, while commands use `domain:feishuApproval`. Even if all paths share a lock, queued full objects based on stale renderer snapshots can still overwrite newer values.

Please:

- disable all config-mutating controls while lookup is pending;
- avoid committing a stale complete object after lookup—merge the approver fields against the latest main-process snapshot, or use an explicit version/CAS check;
- if a shared lock is added, cover the update entry and every relevant command consistently rather than only the new lookup command.

## Small follow-ups worth doing in this round

These are not blockers by themselves, but they are small and directly related:

- `client.close()` resolves pending test requests and removes the abort listener, so the timer is cleared and `abortOutcome` never terminalizes the already-sent test card. Please cover both timer abort and `close()`.
- Clear the App Secret draft when switching Feishu/Lark and when navigating away from this settings section. Closing the Settings window already destroys the renderer, but tab/section navigation does not. The unconfigured edit state also needs a visible Cancel/clear path.
- The copy says API Explorer, but the URLs are documentation pages. Please use the actual deep links:
  - `https://open.feishu.cn/api-explorer?project=contact&resource=user&apiName=batch_get_id&version=v3`
  - `https://open.larksuite.com/api-explorer?project=contact&resource=user&apiName=batch_get_id&version=v3`
  Test the pathname and all four query parameters rather than only the host.
- Make the Step 2 label/hint email-first; currently email is discoverable mainly through the placeholder.
- Automatic fallback-guide expansion currently synthesizes `header.click()`, which persists an expanded state to localStorage. `defaultCollapsed: false` would not fix this because stored state wins over the default. Please expose a non-persisting one-shot expand path instead.

## Suggested regression coverage

At minimum, please cover:

1. real SDK lookup HTTP failure and token failure produce zero sensitive console output;
2. persisted Feishu credentials + selected Lark fails before any outbound request, through lookup, readiness, and runtime startup;
3. same App ID across Feishu/Lark cannot reuse a token across domains;
4. App A lookup followed by App C credentials makes the looked-up approver stale;
5. rotating only the Secret for the same App identity preserves a valid binding;
6. platform switch clears the draft and invalidates/reconfirms identifiers according to their documented scope;
7. lookup concurrency cannot revert Enable/timeout or run Test against an inconsistent snapshot;
8. both timer abort and `client.close()` make test cards non-actionable;
9. actual Explorer deep links are asserted structurally.

The core implementation is otherwise strong: the request shape and `99991672` mapping are sound, email is not persisted, the domain comes from a closed enum, the stale-lookup sequence guard works, the normal abort ordering is correct, and all existing manual ID paths remain present. I also merged the PR into the latest `main`; the focused affected tests pass and the full-suite failures are identical to the current `main` baseline.

Once the credential provenance/binding and SDK logging/cache boundaries are closed, I am happy with the feature direction and expect this to be mergeable.

Cobb04:
Thank you for the thorough review and for validating the implementation against the real SDK and the latest `main`.

I appreciate the detailed findings, especially around SDK logging, credential provenance/platform binding, token cache isolation, and the configuration race condition. We agree that these should be addressed before merging.

We’ll work through the blocking issues and add the relevant regression coverage, then update the PR once the fixes are ready. Thanks again for taking the time to investigate this so deeply.

Cobb04:
> Thank you for the thorough review and for validating the implementation against the real SDK and the latest `main`.
> 
> I appreciate the detailed findings, especially around SDK logging, credential provenance/platform binding, token cache isolation, and the configuration race condition. We agree that these should be addressed before merging.
> 
> We’ll work through the blocking issues and add the relevant regression coverage, then update the PR once the fixes are ready. Thanks again for taking the time to investigate this so deeply.

Quick status update: the three blocking areas from your review have now been implemented on the branch and are passing our focused, security, concurrency, and full-suite checks.

The current PR head is still an intermediate checkpoint, though, and I’m not requesting re-review yet. I’m finishing the small follow-ups from the same review, including:

- terminalizing test cards when the client closes;
- clearing unsaved credential drafts on platform/section changes;
- switching the fallback links to the actual API Explorer URLs;
- making the approver setup copy email-first;
- preventing automatic fallback-guide expansion from changing the user’s saved collapse preference.

I’ll run the complete regression and maintainability review once those are finished, then post a final blocker/follow-up-to-test mapping with the exact results.

No action is needed yet — please hold off on final testing or re-review until that ready-for-review update. Thank you again for the detailed guidance.

Cobb04:
Thank you again for the detailed review. I have pushed the completed remediation and the related small follow-ups. The current PR head is `8c11fb46e8e535aa79822306cf8ae6bdae1c7b32`, reconciled with the `v0.14.0` main commit `86125b9ff72eb3c85ba6e731889a42a6a6b5c006`.

## Maintainer blockers

| Concern | Resolution | Regression evidence |
|---|---|---|
| SDK default logging could expose credentials and lookup input | Injected an explicit non-emitting logger into the REST client, WS client, and EventDispatcher. Main-process SDK/Axios failures are converted to stable sanitized results. | Locked real-SDK tests cover tenant-token and `batch_get_id` failures, real Axios error objects, and every `console.*` channel. Sentinels for email, App ID, App Secret, bearer token, Authorization, and request bodies remain absent. |
| Credentials, platform, approver IDs, and token cache lacked provenance boundaries | Credentials now persist their platform atomically. Lookup reads saved credentials only. Lookup-derived approvers are bound to source, platform, and App identity; readiness, Test, lookup commit, and runtime startup use the same fail-closed rules. Each SDK client has an isolated platform/App cache namespace, without putting the Secret in the key. | Coverage includes Feishu/Lark mismatch before outbound I/O, App changes, Secret-only rotation, legacy/unbound data, manual `open_id`/`user_id`/`union_id`, same-App-ID cross-domain cache isolation, and positive same-client token reuse. |
| Lookup could race stale full-object configuration writes | Renderer writes are now allowlisted field patches. Approver updates remain command-authoritative and merge against the latest main-process state under the Feishu approval lock. Generic update/bulk/hydrate/IPC paths cannot overwrite this state. Overlapping lookup/cancel commands are non-writing, use explicit read-only dependencies, and cannot smuggle a commit. | Deterministic barriers cover lookup A/B supersession, cancel/consume ordering, credentials/Test ordering, enabled and timeout updates in both lock orders, stale full-object attempts, pending-lookup Test, and commit guard edge cases including inherited properties, Proxy traps, and throwing/changing getters. |

## Compatibility and transition behavior

- Existing manual `open_id`, `user_id`, and `union_id` paths remain available.
- A manual `open_id` is not silently trusted after its App/platform identity changes; the ID types retain their scope-aware confirmation rules.
- Legacy credentials or approvers without provenance remain readable but are treated as unbound and fail closed until the user resaves or reconfirms them.
- Rotating only the Secret for the same platform/App identity preserves a valid approver binding.
- A failed secrets write does not mutate route state, candidates, grants, card work, revision, or runtime synchronization.
- Lookup responses expose only an opaque lookup handle; the resolved approver ID remains main-process-side until the authoritative commit.

## Small follow-ups from the same review

These are also complete:

- Timer abort and `client.close()` now terminalize an already-sent test card exactly once, including close-before-send-settles and abort/close overlap. The terminal card is non-actionable.
- Credential drafts, including the App Secret draft, are cleared on platform changes, section navigation, Cancel/Clear, and editing exit without deleting saved credentials.
- The fallback links now use the exact Feishu and Lark API Explorer URLs. Tests assert protocol, host, pathname, and all four query parameters.
- Step 2 copy is email-first in all five supported locales while preserving the manual-ID route.
- Automatic fallback-guide expansion uses a one-shot non-persisting `expand({ persist: false })` path. Normal mouse and keyboard toggles still persist as before.
- Lookup lifecycle failures map only allowlisted stable codes to localized messages; unknown failures retain the sanitized fallback.

## Scope and maintainability

I kept the remediation limited to the reviewed Feishu/Lark approval boundaries:

- no dependency, lockfile, workflow, release-document, or unrelated integration changes;
- no App Secret fingerprint or Secret-derived cache key;
- field patching is restricted to the three existing mutable fields;
- concurrent execution is limited to lookup/cancel, receives explicit read-only dependencies, and is prohibited from persisting or notifying;
- unused lookup generation/planner metadata and duplicated controller test setup were removed (the cleanup reduced that diff by 95 lines);
- no speculative configuration framework, state-management replacement, or unrelated UI refactor was introduced.

Relative to the pinned main, the diff contains only `src/` and `test/` files; most added lines are executable negative/security/concurrency regression coverage.

## Verification at the pushed head

Environment:

- Node `24.18.0`
- npm `11.16.0`
- Electron `41.10.2`
- `@larksuiteoapi/node-sdk@1.71.1`

Local gates:

- focused 12-file approval suite: **866 passed, 0 failed, 0 skipped**
- deterministic seven-case race selection: **70/70 across 10 runs**
- pending lookup/Test overlap: **10/10 across 10 runs**
- full suite: **6,913 total; 6,892 passed; 0 failed; 21 skipped**
- `npm run verify:electron`: passed
- `npm run verify:release`: passed
- syntax checks for every changed JavaScript file: passed
- `git diff --check`: passed
- the repository does not define lint or typecheck scripts, so I have not claimed those gates

Manual validation on the exact final head with a Feishu test tenant covered:

- saved credential and email-to-`open_id` setup;
- runtime reaching the green Running state;
- approved test-card delivery;
- a 60-second unanswered test card becoming Cancelled and non-actionable;
- credential-draft clearing across platform/navigation transitions;
- the exact API Explorer deep link;
- non-persistent automatic fallback-guide expansion.

A live Lark tenant was not available for manual validation; its platform, lookup, cache, provenance, and renderer paths are covered deterministically in the automated suite.

The two fork workflows for this head currently report `action_required` and are awaiting maintainer approval; I am not treating them as passed CI.

I believe all three merge blockers, the requested regression matrix, and the small related follow-ups are now addressed. The PR is ready for another review when convenient. Thank you again for the exceptionally concrete security and concurrency analysis.

rullerzhou-afk:
Thank you again for the very substantial remediation work here. I completed another focused review of the current head (`8c11fb46`). The logger/cache isolation, credential provenance, persisted-only lookup, fail-closed readiness checks, sanitized SDK errors, and the security regression matrix are all valuable work and should be preserved.

I am sorry to ask for another round after the amount of work already invested, but I do not think the current head is safe to merge yet. The remaining concern is no longer the original product direction; it is the orchestration around lookup and commit. I found several merge-blocking user-visible failures with runnable reproductions:

1. In `open_id` mode, a valid email such as `ou_admin@example.com` is classified as a manual ID because the `ou_` check runs first. `feishuApproval.saveManualApprover` accepts it, so the raw email is persisted in prefs and later used as a receive ID. This violates the feature's privacy guarantee. The renderer and main process need one shared recipient classifier, with email classification first, and main must reject any email-shaped value on the manual-save path.

2. Resaving legacy credentials clears an existing Verification Token and Encrypt Key when those fields are left blank, even though the UI explicitly says blank fields keep their current values. The legacy fail-closed/resave policy is still correct; the silent clearing behavior is not. Identity-preserving writes should retain optional fields, while identity replacement/legacy rebinding must use explicit copy and confirmation that match the actual behavior.

3. Typing an approver email can destroy focus and caret at the invalid/valid boundary. Preflight changes synchronously request a full content render, and `renderContent()` replaces the tab DOM. The same can happen when the first character is typed into a credential field while a valid email draft exists. Preflight should update only the button state and a visible status/error row; input handlers must not rebuild the tab.

4. With the locked SDK version, `im.v1.message.create/patch` resolves HTTP 200 responses even when the business `code` is nonzero. As a result, a failed create is logged as sent with an empty message ID and the Test UI waits 60 seconds before reporting “no button response”; a failed timeout patch is treated as successful terminalization. Please add a shared sanitized SDK response assertion to every message create/patch boundary, and require a non-empty `message_id` after create. This can be a small prerequisite PR if that keeps the change easier to review, but #750 would need to rebase onto it before approval.

There are also smaller required fixes in the same area: the new structural test is line-ending-sensitive and fails on a Windows `core.autocrlf=true` checkout; closing Settings does not cancel an in-flight lookup; `missing-approver` currently maps to the “email not found” message; and the disabled Save reason is only exposed through `title`, which is not keyboard/screen-reader reachable.

The larger issue is that the renderer receives a lookup handle and immediately calls commit without any intervening user decision. The two-stage protocol therefore does not provide a product capability, but it adds a retained lookup state machine, multiple phases/IDs, and generic concurrent-command exceptions in the settings controller. Those mechanisms are also where the lifecycle gaps are accumulating.

For the next revision, I would like to converge on these invariants:

- expose one operation to the renderer: submit an email and receive the final success/error result;
- keep the resolved `open_id` entirely in main and do not return a lookup handle to the renderer;
- let main own an `AbortController` for the Settings `webContents`, aborting on tab exit/window destruction;
- perform the network lookup outside the settings-controller lock;
- after lookup, use a short internal controller command to re-read and compare credential identity + secrets revision, then commit only if they still match;
- remove `src/feishu-approval-lookup.js` and the generic `concurrentDeps` / `commitForbidden` controller mechanism added for this flow;
- update preflight UI in place rather than rebuilding the tab.

This preserves the strongest parts of the current implementation: persisted-only credentials, no email persistence, no resolved ID crossing into the renderer, provenance validation, and the controller as the only settings writer. It also avoids holding the Feishu settings lock while waiting on token/lookup network requests.

To keep this from becoming another long patch-by-patch cycle, could you please reply with a short implementation plan for this convergence before making the next code changes? We can align on the shape first and then do one focused review against behavior-level acceptance cases.

I also want to be explicit that there is no pressure to take on another architectural revision if the scope has grown beyond what you intended for this contribution. If you would prefer to stop here, that is completely understandable—please let us know, and we can take over the remaining rework on a maintainer branch while preserving and crediting the solid security/provenance work already completed. Either choice is fine; the priority is getting the feature into a simpler, maintainable, and reliably testable shape.

Cobb04:
> Thank you again for the very substantial remediation work here. I completed another focused review of the current head (`8c11fb46`). The logger/cache isolation, credential provenance, persisted-only lookup, fail-closed readiness checks, sanitized SDK errors, and the security regression matrix are all valuable work and should be preserved.
> 
> I am sorry to ask for another round after the amount of work already invested, but I do not think the current head is safe to merge yet. The remaining concern is no longer the original product direction; it is the orchestration around lookup and commit. I found several merge-blocking user-visible failures with runnable reproductions:
> 
> 1. In `open_id` mode, a valid email such as `ou_admin@example.com` is classified as a manual ID because the `ou_` check runs first. `feishuApproval.saveManualApprover` accepts it, so the raw email is persisted in prefs and later used as a receive ID. This violates the feature's privacy guarantee. The renderer and main process need one shared recipient classifier, with email classification first, and main must reject any email-shaped value on the manual-save path.
> 
> 2. Resaving legacy credentials clears an existing Verification Token and Encrypt Key when those fields are left blank, even though the UI explicitly says blank fields keep their current values. The legacy fail-closed/resave policy is still correct; the silent clearing behavior is not. Identity-preserving writes should retain optional fields, while identity replacement/legacy rebinding must use explicit copy and confirmation that match the actual behavior.
> 
> 3. Typing an approver email can destroy focus and caret at the invalid/valid boundary. Preflight changes synchronously request a full content render, and `renderContent()` replaces the tab DOM. The same can happen when the first character is typed into a credential field while a valid email draft exists. Preflight should update only the button state and a visible status/error row; input handlers must not rebuild the tab.
> 
> 4. With the locked SDK version, `im.v1.message.create/patch` resolves HTTP 200 responses even when the business `code` is nonzero. As a result, a failed create is logged as sent with an empty message ID and the Test UI waits 60 seconds before reporting “no button response”; a failed timeout patch is treated as successful terminalization. Please add a shared sanitized SDK response assertion to every message create/patch boundary, and require a non-empty `message_id` after create. This can be a small prerequisite PR if that keeps the change easier to review, but #750 would need to rebase onto it before approval.
> 
> There are also smaller required fixes in the same area: the new structural test is line-ending-sensitive and fails on a Windows `core.autocrlf=true` checkout; closing Settings does not cancel an in-flight lookup; `missing-approver` currently maps to the “email not found” message; and the disabled Save reason is only exposed through `title`, which is not keyboard/screen-reader reachable.
> 
> The larger issue is that the renderer receives a lookup handle and immediately calls commit without any intervening user decision. The two-stage protocol therefore does not provide a product capability, but it adds a retained lookup state machine, multiple phases/IDs, and generic concurrent-command exceptions in the settings controller. Those mechanisms are also where the lifecycle gaps are accumulating.
> 
> For the next revision, I would like to converge on these invariants:
> 
> - expose one operation to the renderer: submit an email and receive the final success/error result;
> - keep the resolved `open_id` entirely in main and do not return a lookup handle to the renderer;
> - let main own an `AbortController` for the Settings `webContents`, aborting on tab exit/window destruction;
> - perform the network lookup outside the settings-controller lock;
> - after lookup, use a short internal controller command to re-read and compare credential identity + secrets revision, then commit only if they still match;
> - remove `src/feishu-approval-lookup.js` and the generic `concurrentDeps` / `commitForbidden` controller mechanism added for this flow;
> - update preflight UI in place rather than rebuilding the tab.
> 
> This preserves the strongest parts of the current implementation: persisted-only credentials, no email persistence, no resolved ID crossing into the renderer, provenance validation, and the controller as the only settings writer. It also avoids holding the Feishu settings lock while waiting on token/lookup network requests.
> 
> To keep this from becoming another long patch-by-patch cycle, could you please reply with a short implementation plan for this convergence before making the next code changes? We can align on the shape first and then do one focused review against behavior-level acceptance cases.
> 
> I also want to be explicit that there is no pressure to take on another architectural revision if the scope has grown beyond what you intended for this contribution. If you would prefer to stop here, that is completely understandable—please let us know, and we can take over the remaining rework on a maintainer branch while preserving and crediting the solid security/provenance work already completed. Either choice is fine; the priority is getting the feature into a simpler, maintainable, and reliably testable shape.

Thank you very much for your detailed review. I fully understand your main concern—the current two-phase lookup/commit flow does introduce additional lifecycle and concurrency complexity, yet it doesn't provide a meaningful confirmation step for users. I also realize that reviewing such code is quite exhausting, and I sincerely apologize for the extra burden my over-engineering has placed on you.
🙇🙇🙇
Moving forward, I would like to continue with the convergence pass. I plan to pause further code changes for now and first post a concise implementation plan, covering: the proposed single-operation main-process flow, cancellation and credential revision checks, recipient classification, SDK response validation, the remaining smaller fixes, and the associated regression test coverage.

Before starting the next round of implementation, I will wait for alignment on this plan to ensure we are on the right track and avoid going down the wrong path again.

Cobb04:
Thank you for the second detailed review, and for the runnable reproductions — they made confirmation straightforward.

I reproduced all nine. `ou_admin@example.com` classifies as a manual `open_id` because the `ou_` test runs before the email test, and `saveManualApprover` persists it. Legacy resave clears Verification Token and Encrypt Key, because `credentialPlatform === "unknown"` takes the identity-replacement branch. Preflight calls `requestRender({content:true})`, which clears `#content`, replacing the input node mid-typing. The create/patch boundaries never inspect the business `code`. Closing Settings strands the resolved `open_id`. The shared error string, `title`-only disabled reason and CRLF-fragile needle are as you describe.

Your diagnosis of the root cause matches mine: the renderer orchestrates an operation that belongs to main, which forces the retained handle, the phases, and the concurrent-command exceptions.

Proposed convergence. One renderer operation submits an email and receives a final sanitized result — no handle, no `open_id`. Main owns the `AbortController`, refuses non-Settings senders, and aborts on tab exit, `webContents` destruction and `render-process-gone`, with distinct reasons so a late frame still distinguishes cancelled from superseded. Credential identity and secrets revision are captured in one synchronous block; the lookup runs outside the settings lock; a short internal command then re-checks the abort signal *after* acquiring the lock, re-reads platform, credential platform, App ID and revision, then commits only approver/provenance fields, preserving concurrent enabled/timeout edits. It goes on the existing `INTERNAL_SETTINGS_COMMANDS` allowlist, so the renderer cannot call it.

The smaller defects are addressed alongside: a shared classifier checking email before any prefix rule, plus a hard email rejection on main's manual-save path; a business-code and `message_id` assertion at all six create/patch boundaries; identity-preserving writes that keep blank optional credentials, with App replacement warning and writing nothing until confirmed; distinct copy for missing-approver and email-not-found; and an in-place preflight preserving the input node, focus and caret, exposing the disabled reason via visible `aria-describedby` text rather than `title`. Behavior tests replace the CRLF-fragile source slicing.

To remove: `src/feishu-approval-lookup.js`, handles and tombstones, resolve-then-commit, `concurrentDeps` / `commitForbidden`, and their tests.

Could you confirm this shape, and whether you want the SDK assertion as a separate prerequisite PR, before I start?

Cobb04:
Friendly ping on the convergence plan from Aug 5.

Still waiting on confirmation before coding:
1. Is the proposed single-operation main-process flow (no handle / no open_id to renderer; abort on Settings lifecycle; lookup outside lock; short internal commit with credential revision check) the shape you want?
2. Prefer the SDK business-code / message_id assertion as a separate prerequisite PR, or in the same revision?

Happy to start as soon as this is confirmed. If you'd rather take over the remaining rework on a maintainer branch, that's also fine — please just say which path you prefer.

rullerzhou-afk:
Shannon, I owe you a sincere apology for leaving this unanswered.

I asked you to pause implementation and post a convergence plan first. You did exactly that on the same day, with a careful and concrete response, but I failed to return to the PR after that review session ended. I had recorded a follow-up window locally for the case where no response arrived, and then missed that you had already replied. That was entirely my oversight—not a reflection of any concern about your plan or a lack of appreciation for the work you have put into this PR.

Thank you for reproducing all of the reported issues, engaging seriously with two unusually detailed review rounds, and being willing to simplify the architecture after already investing substantial effort in the security, provenance, cache-isolation, concurrency, and regression work. That effort is recognized and valuable.

I have reviewed your Aug 5 convergence plan again, and I confirm that it is the shape I want:

- one renderer-visible operation, with no lookup handle or resolved `open_id` crossing into the renderer;
- main-process ownership of lookup cancellation and Settings lifecycle;
- network lookup outside the settings-controller lock;
- a short internal commit that rechecks credential identity and secrets revision;
- removal of the lookup coordinator and generic concurrent-command exceptions;
- shared recipient classification, sanitized SDK response validation, and in-place preflight UI updates.

Please proceed with that plan. The SDK business-code and `message_id` assertions do not need to be split into a prerequisite PR; including them in this revision is fine.

Again, I am sorry that you had to ping after already doing exactly what I asked. Thank you for your patience and for the care you have continued to put into this contribution.

Cobb04:
Thank you for the careful reply and for confirming the Aug 5 plan — and no worries about the delay; I appreciate you coming back to it with such clarity.

I'll proceed with that shape now, including the SDK business-code / `message_id` assertions in this same revision. I'll ping here again when the convergence pass is ready for review.

Cobb04:
Thank you for confirming the convergence plan. The convergence pass is now pushed.

**This update supersedes my Aug 3 remediation summary.** The two-stage lookup handle / concurrent-command design described there has been removed.

- Current PR head: `6c624bdb394790501a63a9dc71ab86967ed374f0`
- Compared against latest `main`: `cab64a86632d9194b06d28cfc3b42cefc1202816`

### Confirmed convergence shape

| Confirmed invariant | Final implementation |
|---|---|
| One renderer-visible operation | The renderer sends only `feishuApproval.saveApproverByEmail({ email })` and receives a final sanitized status/error code. No lookup handle or `open_id` crosses into the renderer. |
| Main owns cancellation and Settings lifecycle | The trusted Settings sender owns one main-process lookup lifecycle. New lookup, explicit cancel, sender destruction, `render-process-gone`, and disposal all abort the active operation. |
| Network lookup stays outside the Settings controller lock | Email lookup is handled at the IPC/main boundary. It does not hold a `SettingsController` command lock while waiting on the network. |
| Commit is short and revalidates authority | A renderer-blocked internal command takes the existing Feishu lock, rereads platform/App identity/revision, fails closed on change, and merges only the authoritative approver/provenance fields so current `enabled` and timeout values are preserved. |
| Remove the extra mechanism | The lookup coordinator file, handles/phases/tombstones, two-stage resolve/commit commands, and generic concurrent-command exceptions are gone. `src/settings-controller.js` is byte-identical to latest `main`. |
| Shared classification, SDK validation, and in-place preflight | Renderer and main share the same email-first classifier; manual save rejects email input. All six SDK message boundaries validate business status, and create calls require a non-empty `message_id`. Preflight now patches the mounted row in place with visible/ARIA-linked feedback, preserving focus and caret position. |

### Other reviewed behavior

- Saved credentials remain authoritative; unsaved renderer credential drafts are never used by lookup.
- Feishu/Lark cache identity and lookup-derived approver provenance remain bound to platform + App ID, while Secret-only rotation preserves the same App identity.
- Readiness, Test, and runtime startup use the same fail-closed binding rules. Manual `open_id`, `user_id`, and `union_id` remain supported.
- Legacy same-App credential confirmation preserves blank optional Token/Encrypt Key fields. App replacement performs zero writes until explicit confirmation, then replans against current state before writing.
- The requested follow-ups are included: exact API Explorer links, email-first guidance, non-persistent fallback expansion, distinct “not configured”/“not found” messages, credential-draft cleanup, and terminalization of sent test cards on timeout, client close, and true app quit.

### Verification

- Focused PR #750 gate: **932 passed, 0 failed**
- Deterministic concurrency matrix: **11 cases × 10 runs = 110/110 passed**
- Full-suite comparison:

  | | Total | Passed | Failed | Skipped |
  |---|---:|---:|---:|---:|
  | PR head | 8,015 | 7,959 | 30 | 26 |
  | exact latest-main baseline | 7,868 | 7,812 | 30 | 26 |

  The 30 failures match latest `main` exactly by test file and line and are confined to the local DSH install-bridge environment; the affected production and test files are byte-identical to `main`. The PR adds **147 passing tests and no additional failure or skip**.
- Electron verification: `41.10.4`; release verification: `v0.15.0`; changed-JS syntax and diff/package checks passed.
- Manual exact-head Feishu QA passed: real email lookup/save, focus/caret preservation, API Explorer navigation, test-card delivery/approval, timeout terminalization, and `⌘Q` terminalization.
- No live Lark tenant was used; Lark platform/cache/provenance behavior is covered by deterministic automated tests.
- Latest-main-relative scope is 22 allowlisted files (10 production, 12 test), with no package, lockfile, workflow, or release-document changes.
- Independent final spec/security review reported no findings. Independent standards review reported no blocking/important findings and recommended no last-minute cleanup commit.

GitHub currently reports no status checks for this head, so I am not claiming CI passed. The code is frozen at the SHA above and is ready for another review when convenient. Thank you again for the detailed guidance.

rullerzhou-afk:
Thank you for the contribution and for revising the original approach. During maintainer review we found several remaining authority-boundary, shutdown-lifecycle, accessibility, and card-update race edges, including one reproduced against a real Lark card. We therefore took over the remaining hardening work and pushed maintainer commit 145b3e6d directly to this PR branch (maintainer edits were enabled). The commit preserves your email-to-open_id lookup flow while making manual-ID validation authoritative in the main process, keeping fallback edits draft-only, draining terminal card work on quit, improving lookup status accessibility, and ensuring a late Lark action cannot overwrite either the desktop decision or the final card state. Validation completed: 1000 related tests passed, 113 focused client/permission tests passed, 144 strict unhandled-rejection tests passed, and Windows/Lark smoke confirmed the desktop decision wins, stale buttons disappear, and exit remains bounded. The full suite had only the two previously known unrelated Windows codex-install child-process flakes. No further action is needed from you; thank you again for starting and iterating on this feature.

PR #753, opened 2026-07-26
feat: single-window mode — merge hit window into render window

### LINKED ISSUE #752: [Feature Request] Single-Window Mode for Tiling Wayland Compositors
# [Feature Request] Single-Window Mode for Tiling Wayland Compositors (niri, Hyprland, etc.)

## Problem

Clawd's current **two-window architecture** (render window + transparent hit/input window) doesn't work well on tiling Wayland compositors:

- **niri** (scrollable-tiling): no way to bind two independent floating windows together for dragging and workspace switching. Dragging the hit window moves it but leaves the render window behind.
- **Hyprland / Sway / river**: even with `window-rule` tricks, keeping two windows perfectly stacked and synchronized is fragile.

### Why the two windows?

The render window (`Clawd`) shows the pet SVG animation via `renderer.js`. A separate hit window (`clawd-on-desk`) — a transparent overlay — captures all pointer/drag/drop events via `hit-renderer.js`, then relays input through IPC to main process.

## Proposed Solution

Add a **Single-Window Mode** toggle (Settings → General) that merges the hit window's input handling directly into the render window:

### Implementation (surprisingly minimal!)

Three files need changes, one file stays completely untouched:

| File | Change |
|---|---|
| `src/main.js` | Skip `petWindowRuntime.createHitWindow()` when single-window mode is on; redirect `sendToHitWin()` → `sendToRenderer()` |
| `src/preload.js` | Add hit IPC methods (`dragLock`, `dragMove`, `dragEnd`, etc.) to the existing `electronAPI` bridge |
| `src/renderer.js` | Append pointer/drag/click/drop event handlers (from `hit-renderer.js`), adapted to use `container`/`electronAPI` |

**Zero changes needed to `src/pet-interaction-ipc.js`** — it registers all IPC handlers globally via `ipcMain.on()`, so it works identically regardless of which window sends the message.

### Key adaptations

| Hit renderer | Merged renderer |
|---|---|
| `#hit-area` element | `#pet-container` element (already has `pointer-events: none` on SVG) |
| `window.hitAPI.xxx` | `window.electronAPI.xxx` |
| `window.hitThemeConfig` | `window.themeConfig` |
| Separate `isReacting` vars | Prefixed `_inputReacting` (avoids collision with visual reaction state) |

### Diff size

~150 lines added across 3 files. The entire hit window creation, IPC routing, and input event loop collapse into the renderer.

## Benefits

- ✅ Single floating window → works on ALL Wayland compositors
- ✅ Drag, click, file drop all work as before
- ✅ Mini mode, reactions, context menu — all preserved
- ✅ Backward compatible: when the toggle is off, the two-window path is unchanged
- ✅ Minimal diff, easy to review

## Proof of Concept

I've already tested this successfully on niri 26.04 with clawd v0.13.0. Full documentation and source at:

https://github.com/alexkazx23-beep/clawd-on-desk/tree/single-window

Would you be open to a PR?

Alxkcx:
## Summary

Merges the transparent hit window's input handling directly into the render window, eliminating the separate input overlay BrowserWindow. This fixes clawd on tiling Wayland compositors (niri, Hyprland, Sway, etc.) where binding two independent floating windows is not supported.

Closes #752

## Changes

| File | Change |
|---|---|
| `src/main.js` | Skip `createHitWindow()`, redirect `sendToHitWin()` → `sendToRenderer()` |
| `src/preload.js` | Add all hit IPC methods to `electronAPI` bridge |
| `src/renderer.js` | Append pointer/drag/click/drop handlers adapted from `hit-renderer.js` |

## What's NOT changed

**`src/pet-interaction-ipc.js`** — ZERO changes. All IPC handlers are registered globally via `ipcMain.on()`, so they work identically regardless of which window sends the message.

## Key design decisions

- Input state uses prefixed variables (`_inputReacting`, `_dragInputReacting`) to avoid collision with the visual reaction state
- `#hit-area` → `#pet-container` (which already has `pointer-events: none` on the SVG object)
- All events on `container` element instead of a separate window
- No new dependencies, no config changes needed

## Tested on

- niri 26.04 Wayland compositor
- clawd-on-desk v0.13.0
- Drag, click, right-click, file drop all working

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -1,4 +1,4 @@
-const { app, BrowserWindow, screen, ipcMain, globalShortcut, nativeTheme, dialog, shell, nativeImage, powerSaveBlocker, powerMonitor, clipboard, safeStorage } = require("electron");
+const { app, BrowserWindow, screen, ipcMain, globalShortcut, nativeTheme, dialog, shell, nativeImage, powerSaveBlocker, powerMonitor, clipboard } = require("electron");
 // ── Linux/Wayland: relaunch under XWayland so the pet is draggable (issue #441) ──
 // Native Wayland ignores client-side window positioning and blocks global cursor
 // queries, so the pet spawns centered, can't be dragged, and has no tracking;
@@ -81,12 +81,6 @@ const {
 } = require("./settings-size-preview-session");
 const { registerSettingsIpc } = require("./settings-ipc");
 const createSettingsEffectRouter = require("./settings-effect-router");
-const {
-  getPetTintIdForTheme,
-  resolvePetTintPayload,
-  getPetAccessoryIdForTheme,
-  resolvePetAccessoryPayload,
-} = require("./pet-customization-catalog");
 const { registerSessionIpc } = require("./session-ipc");
 const { createSessionFolderOpener } = require("./session-open-folder");
 const { registerPetInteractionIpc } = require("./pet-interaction-ipc");
@@ -143,9 +137,7 @@ const {
 } = require("./session-focus");
 const { focusCodexThreadTarget } = require("./session-focus-handoff");
 const { isSessionInProgress } = require("./state-session-snapshot");
-const { restoreSessionsFromRecoveryLeases } = require("./session-recovery-loader");
 const { getAllAgents, getAgent } = require("../agents/registry");
-const { getAgentIconUrl } = require("./state-agent-icons");
 // ── Autoplay policy: allow sound playback without user gesture ──
 // MUST be set before any BrowserWindow is created (before app.whenReady)
 app.commandLine.appendSwitch("autoplay-policy", "no-user-gesture-required");
@@ -256,7 +248,6 @@ const SIZES = {
 // `_settingsController.applyUpdate()`, which auto-persists.
 const prefsModule = require("./prefs");
 const { createSettingsController } = require("./settings-controller");
-const { loadOrCreateInstallationIdentity } = require("./remote-ssh-identity");
 const { createTranslator, i18n, SUPPORTED_LANGS } = require("./i18n");
 const {
   getBubblePolicy,
@@ -388,10 +379,6 @@ const _settingsController = createSettingsController({
     uninstallAutoStart: _uninstallAutoStartHook,
     resolveTextScaleDisplayKey: () => getSettingsDisplayKey(),
     syncClaudeHooksNow: () => _server.syncClawdHooks({ source: "settings", automatic: false }),
-    setClaudeQuotaCollectionEnabled: (enabled) => _server.setClaudeQuotaCollectionEnabled({
-      enabled,
-      source: "settings-quota-collection",
-    }),
     uninstallClaudeHooksNow: _uninstallClaudeHooksNow,
     startClaudeSettingsWatcher: () => _server.startClaudeSettingsWatcher(),
     stopClaudeSettingsWatcher: () => _server.stopClaudeSettingsWatcher(),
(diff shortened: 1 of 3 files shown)


rullerzhou-afk (changes requested):
Thanks for working on #752. Consolidating the input and render surfaces is a reasonable direction for compositors that cannot keep two floating windows bound together. I reviewed the current head (`d84666220a1525f36d5892cec50594a173469c3d`), and the implementation still has several blocking gaps.

## Required changes

1. **Complete the input-window ownership migration.**

   `createHitWindow()` is disabled, but `startMainTick()` still unconditionally calls `win.setIgnoreMouseEvents(true)`. Therefore the new pointer, drag, context-menu, and drop listeners in `renderer.js` cannot receive any input on any platform.

   Simply making the render window interactive is not sufficient either: `#pet-container` covers 100% of the transparent BrowserWindow, while the old path dynamically sizes and shapes `hitWin` to the active theme/state hitbox and clips it at mini-mode monitor seams. Without an equivalent native hit-region or dynamic click-through design, the invisible padding around the pet will consume desktop input.

   The platform-specific owners must move as well. Windows focusability/fullscreen and size-preview protection, plus the macOS editing-overlap click-through fallback, still operate only on `hitWin`; they become no-ops after this change.

2. **Wire the hit-layer configuration, state gates, and reload lifecycle into the single-window path.**

   The render config returned by `getRendererConfig()` does not contain `reactions`; those are only present in `getHitRendererConfig()`. Consequently `_inputReactions` remains empty and click/drop reactions cannot play. Mini pre-entry also still sends its early input gate through `hit-state-sync`, but the merged renderer never subscribes to `onHitStateSync`, so input is no longer blocked while the mini transition owns window movement.

   Theme switching has another two-window dependency: `theme-fade-sequencer` requires both render and hit windows before reloading. With `hitWin === null`, it takes the fallback path without reloading the render contents (a direct harness observed `reloads: 0`), while the new `theme-config` is only sent from `did-finish-load`. Theme/variant changes can therefore leave the renderer on stale configuration. The lifecycle needs explicit one-window and two-window behavior.

3. **Rebuild this as a narrow, gated change on current `main`.**

   This patch removes 758 lines and reverts unrelated functionality already present in its own base, including Remote SSH installation identity/isolation wiring, startup session recovery, account quota/HUD wiring, pet tint/accessories, permission automation modes, tutorial agent icons, and renderer transition fixes. These removals are unrelated to #752 and must not be part of this change.

   The implementation is also unconditional: it changes Windows, macOS, Linux X11, and XWayland rather than introducing the opt-in mode proposed in #752. Please preserve the existing dual-window path and add an explicit platform/setting gate with a fallback. Note that this app still defaults Linux to XWayland; removing the second window may address compositor pairing there, but it does not by itself solve native-Wayland positioning/global-cursor limitations.

4. **Add tests for the path that will actually run.**

   No tests were changed. The existing `hit-renderer` tests still exercise the old file that is no longer loaded, so their 40/40 result does not cover the merged renderer. On the same focused test set, the PR's base passes 157/157 while this head passes 101/157 (56 failures). Please add coverage for single-window selection, exact hit routing, reactions, mini pre-entry, theme reload, Windows focusability, macOS overlap dodge, and the dual-window fallback.

   The resulting implementation also needs real Electron validation on niri/Hyprland under the intended XWayland/native-Wayland mode, plus Windows and macOS regression checks if those platforms remain in scope. In particular, verify transparent-area click-through, sustained dragging, right-click/drop, mini mode, monitor seams, fullscreen/focus behavior, and text-input overlap.

The IPC observation in the PR is generally correct: the `ipcMain.on()` handlers are mostly sender-independent, and the drop acknowledgement can reply to the render sender. The blockers are the native input ownership, configuration/lifecycle migration, unrelated regressions, and missing coverage around that IPC layer.

Issue #743, opened 2026-07-25
希望能支持一下orca


lyings:
https://github.com/stablyai/orca，orca这个项目目前可以把所有的agent cli 集中到一起管理，当前clawd 会话/审批无法跳转到指定的orca中的终端窗口，只能打开orca。希望能增加直接跳转的对应orca对应终端的功能。

KaiC5504:
I have a working implementation of this and I'll open a draft PR shortly — happy to be assigned if you'd like.

Two findings worth putting here in case anyone else picks this up:

- Adding `Orca` to the editor process list doesn't work. Orca hosts its terminals in a detached
  `orca-terminal-daemon.exe` whose parent PID is already dead, so `Orca.exe` never appears in the
  agent's ancestry walk and that list is never consulted for it.
- `ORCA_TERMINAL_HANDLE` from the environment goes stale — handles are re-issued per runtime session,
  so `terminal switch` returns `terminal_handle_stale`. The stable identifier is `ORCA_PANE_KEY`
  (`"<tabId>:<leafId>"`), resolved against `orca terminal list --json` on each focus.

Only the Windows path is verified on real hardware. @lyings — if you're on macOS or Linux, a check on
the pane-switch half once the PR is up would be a big help.

lyings:
期待你的PR，我使用的是MacOS。

rullerzhou-afk:
感谢建议，也谢谢你说明使用的是 macOS。Orca 终端聚焦支持已经通过 #759 合并到 `main`。

我们已分别在真实 macOS 和 Windows 设备上验证：Clawd 可以根据稳定的 Orca pane identity 切换到正确的终端标签并将 Orca 带到前台；相关失败路径也保持 fail-closed，不会把 Direct Send 内容粘贴到错误标签。

当前支持边界是本地 macOS / Windows；Linux 尚未接入，Remote SSH 下的可见 pane 切换也还没有作为完整真机能力承诺。核心的 macOS Orca 支持已经落地，因此先关闭这个 Issue。感谢你提出需求并耐心等待！

PR #759, opened 2026-07-27
feat(focus): support Orca terminals on focus-to-terminal

KaiC5504:
Refs #743.

Focus-to-terminal does nothing when an agent CLI runs inside [Orca](https://github.com/stablyai/orca).
Clicking the pet or hitting the permission hotkey leaves the foreground where it was.

## Why the obvious fix doesn't work

Adding `'Orca'` to `$editorProcessNames` in `src/focus.js` has no effect. That list is only consulted for
processes found in the agent's ancestry walk, and `Orca.exe` is never in it:

1. Orca hosts terminals in a detached, persistent `orca-terminal-daemon.exe` whose own parent PID is
   already dead, so the 8-deep parent walk dead-ends. `Orca.exe` itself is a child of `explorer.exe` —
   a separate process tree. (VS Code works today because its pty host is a real child of `Code.exe`.)
2. Orca's only visible window is titled just `Orca`, with no path in it, so the cwd/title matching in
   `FindByPidTitles` finds nothing.
3. Orca is Electron but *not* a VS Code fork — no `product.json`, no extensions directory — so the
   auto-installed VS Code extension route can't apply either.

Before this change, on Windows:

```
__CLAWD_FOCUS_RESULT__ {"reason":"wt-title-mismatch-no-pid-window","targetHwnd":null,"confirmed":false}
```

## Trap worth documenting

`ORCA_TERMINAL_HANDLE` is present in the environment and looks like the right thing to use. It isn't —
handles are re-issued per runtime session, so the env value goes stale:

```
env  ORCA_TERMINAL_HANDLE  term_4602ecfa-…  →  {"ok":false,"error":{"code":"terminal_handle_stale"}}
live handle, same pane     term_63323b46-…  →  {"ok":true}
```

The stable identifier is `ORCA_PANE_KEY` (`"<tabId>:<leafId>"`), which matches `tabId` + `leafId` in
`orca terminal list --json`. Anyone implementing this from the obvious env var ships a broken fix.

## Approach

A new terminal-identity field `orcaPaneKey` rides the **existing `tmuxSocket` seam** — the same five
layers, so no new mechanism is introduced: hook env parse → `orca_pane_key` on the wire → `/state` and
`/permission` normalizers → `src/state.js` → `focusEntry` → `normalizeFocusRequest`.

Focus itself needs both halves; neither is sufficient alone (measured — `terminal switch` returns
`ok` but leaves the foreground on the other app, and a bare window raise lands on whatever tab was
last active):

| Half | Mechanism |
| --- | --- |
| Raise the window | Windows: Win32 `SetForegroundWindow` on the single visible unowned `Orca` window, inside the generated PowerShell (reason `orca-window`). macOS: `/usr/bin/open` on the absolute `Orca.app` bundle path, run only after the switch is confirmed (reason `orca-app-raised`), with the generic macOS raise suppressed for a pane-keyed request — it derives its target from the source process's `ps` comm, which for an Orca session resolves to `Orca.app` itself, so it would activate the IDE before the switch had answered. |
| Switch to the pane | `orca terminal list --json` → resolve the live handle by pane key (falling back to the longest `worktreePath` prefix of the request cwd, normalising `\` vs `/` — Orca reports both forms) → `orca terminal switch --terminal <handle> --json`, with one re-resolve on `terminal_handle_stale`. Handle cached per session so the steady state skips `terminal list`. |

The switch reports back rather than running fire-and-forget. It resolves `{ok, match, reason}` — never
rejects — where `match` is `"exact"` for a pane-key hit and `"cwd"` for the worktree fallback, and the
Windows focus promise carries it as `focusResult.orcaPane` for sessions that have a pane key. Telegram
Direct Send pastes only on `ok && match === "exact"`; every other outcome returns
`orca_pane_unconfirmed` before the clipboard is written, so the existing clipboard fallback handles
delivery instead of typing into whichever pane happened to be focused. Equal-best `worktreePath`
matches fail as `orca-pane-ambiguous` and switch nothing, and only exact matches enter the handle
cache — a cached cwd guess would replay as an ordinary cache hit and be indistinguishable from a real
pane match.

A stored pane key is dropped when an event reports a terminal identity (`sourcePid` / `wtHwnd`) that
differs from the stored one. Clearing on a session-start event covers producers that emit one, but
`antigravity-hook.js` emits none of the three spellings and its id normalizes `conversationId`, so
resuming the same conversation from another terminal landed back on the same entry with a key still
naming the old pane — which then outranked the identity that had just arrived and was correct. Only
values present on both sides count: most events carry no process metadata, and treating absent as
moved would blank the key on the very next event.

**Windows and macOS ship; Linux deliberately doesn't.** Because the raise is required, the pane switch
is dispatched only from branches whose raise can be trusted. On macOS that means `open` against an
absolute bundle path (`/Applications/Orca.app`, then `~/Applications/Orca.app`) rather than
`open -a Orca`, whose name resolution is loose enough to reach OrcaSlicer; the exact path can't, needs
no Automation consent, and restores minimized windows. It runs *after* a confirmed switch, unlike the
Windows raise which fires from inside the generated script: `open` launches Orca if it isn't running,
and because Orca's terminal daemon outlives its window, a sticky pane key pointing at a dead session
would otherwise cold-start the IDE on a click. The Linux raise (`wmctrl -x -a orca` → `xdotool`) was
written and then removed — I have no hardware to check it on, and it matches WM_CLASS `orca` as a
substring, which also matches GNOME's Orca screen reader and OrcaSlicer, so a miss would activate an
unrelated window and log it as a success. That is exactly what the other branches refuse to do.

Two details that matter for correctness:

- **Precedence.** The Orca branch runs *before* the title-matched window cache and the
  `wt_hwnd`-from-hook fallback, and both of those are skipped when the session is Orca-hosted.
  Placed after them, a SessionStart next to an open Windows Terminal window would focus that
  unrelated terminal and report success.
- **Trust gate.** Any shell launched *from* an Orca pane inherits `TERM_PROGRAM=Orca` and
  `ORCA_PANE_KEY`, so the key is only honoured when Orca is the innermost thing in the terminal stack.
  Ten markers veto it: `WT_SESSION`, `ALACRITTY_WINDOW_ID`, `WEZTERM_PANE`, `KITTY_WINDOW_ID`,
  `KONSOLE_VERSION`, `GNOME_TERMINAL_SCREEN`, `ConEmuPID`, plus the multiplexers `TMUX`, `STY` and
  `ZELLIJ` — a multiplexer server outlives the pane it started in, so re-attaching a session from
  another terminal would otherwise carry a pane key that no longer describes where the agent is.

Latency on Windows, measured 3×: `terminal list` 314–379 ms, `terminal switch` 323–337 ms warm; handle
caching cuts the steady state to ≈330 ms. The Win32 raise happens inside the generated script and is
instant, and the pet and dashboard call sites don't await the focus promise, so the CLI round-trip stays
off the click's synchronous path. Direct Send is the one caller that does wait on it, by design — that
wait is what makes the paste land in the right pane.

A cold CLI is much worse than warm, and worse than I first measured. The timeout is 6 s and a killed
call is reported as `orca-cli-timeout` rather than a missing pane, but in daily use the *first* focus of
a session has been observed blowing straight through it — a cold `terminal list` killed at exactly
6.0 s, with Orca already raised on whichever tab was last active and no switch at all. The next click
meets a warm CLI and lands in ~0.5 s. I've listed that under known limitations rather than papering over
it with a longer timeout; see the question at the end.

## Scope

Focus behaviour changes on Windows and macOS. The *field* is carried by every producer that already
sends `tmux_socket` — one mechanical line each across the agent hooks, plus the Python (`hermes`) and
`.mjs` (`opencode-family`) plugins — because the collection is a pure env read with no platform
component and the alternative is one agent working and the rest silently not. Confirmed in review that
this stays broad rather than being trimmed to Claude Code.

## Verification

**Windows is verified on real hardware.** The reason flips from the diagnosis above to a real HWND, and
the pane switch round-trips against the live CLI on both the cold and cached paths:

```
orcaHosted=true  + wtHwnd=4660 → reason=orca-window  targetHwnd=9372300  confirmed=true
orcaHosted=false               → wt-title-mismatch-no-pid-window  targetHwnd=null  (correctly not raised)
live CLI round-trip            → branch=orca reason=orca-pane-switched
```

Also checked: an agent in Windows Terminal with Orca open in the background still focuses Windows
Terminal (no stealing), the VS Code path is untouched, and Direct Send now waits for the switch before
pasting.

**macOS is implemented but has not been run on real hardware** — bundle locations, LaunchServices
behaviour and the raise-after-switch ordering are covered by tests against mocks only. Direct Send is
Windows-only (`platform_unsupported`), so the paste gate cannot be exercised there by construction.

**Linux is out of scope here** (see above) — an Orca-hosted session on Linux behaves exactly as it does
today, falling back to the process-tree walk that cannot see the daemon. @lyings, if you're on Linux
and want to pick that up, the pane-switch half is the same CLI call and already tested; it needs a
by-name window raise you can verify locally.

Tests: new `test/focus-orca.test.js` (branch ordering, the focus-steal gate, the ten-marker veto loop,
stale-handle re-resolve, `worktreePath` fallback and its ambiguous-tie failure, CLI-missing and
CLI-timeout degradation, the switch outcome contract, the macOS bundle lookup and raise ordering, the
macOS CLI candidate paths, and — driven through the public `focusTerminalWindow()` entry with `ps`
reporting the packaged daemon — that a missing pane never calls `/usr/bin/open`, that a successful
switch completes before the raise, and that a second request cannot start a concurrent switch; the
platform dispatch call sites, producer/consumer parity) plus extensions to the hermes-plugin,
antigravity-hook, pi-extension-core, opencode-family, server-route-permission, telegram-direct-send
(the exact-match paste gate and its clipboard fallback), focus-windows (the focus promise waiting on
the switch) and state suites. 16 failures in the full suite on my Windows machine are unrelated to the
files touched here (remote-ssh deploy/path, codex-install, doctor and server-config path/env artifacts;
three of them are caused by Orca itself setting `CODEX_HOME`).

## Known limitations, left in deliberately

- **The trust gate is env-based, so a terminal that advertises nothing defeats it.** A bare
  `conhost`/`pwsh` on Windows, and `xterm`/`urxvt`/`st`/`xfce4-terminal`/`terminator` on Linux, set no
  marker — start an agent in one of those *from* an Orca pane and it still looks like the pane, so Orca
  is raised and reported as a confirmed focus while the real window is never found. Gating on the
  process chain instead (`orca-terminal-daemon` does appear in `pid_chain`) would close this properly,
  but it would reintroduce the dependency on the pid walk that this design deliberately avoids.
- **With two or more Orca windows, Windows raises nothing** (`orca-window-ambiguous`): the whole window
  title is `Orca`, so there is no disambiguator, and unlike the Windows Terminal cascade there is no
  earlier strategy to fall back to. The pane switch still moves the tab in a window that stays behind.
- **On macOS, Orca installed outside `/Applications` and `~/Applications` is never raised**
  (`orca-app-not-found`). The tab still switches; the window stays behind. Reading the bundle path from
  LaunchServices would close it, at the cost of another CLI hop on a path that already waits on two.
- The pane key doubles as the capability flag, so a missing or malformed key falls back to the old
  Windows-Terminal guess, which can still steal focus. Fixing that properly needs a second boolean
  carried through all five layers.
- The pane switch sits behind the existing `sourcePid` early return even though it only needs the key
  and cwd.
- The pane-key → handle cache has no epoch token, so it can only be invalidated by a stale-handle error.
- `src/state.js` doesn't re-normalize the key on read the way `ghosttyTerminalId` does.
- **On Windows the first focus of a session can land on the wrong tab.** The raise is instant and the
  switch is not, so when a cold `terminal list` is killed at the 6 s timeout, Orca is already in front
  showing whichever tab was last active. Clicking again fixes it. A longer timeout trades one bad
  outcome for another (seconds of apparently nothing), and pre-resolving the handle when a pane-keyed
  session first registers would fix it properly — I'd rather agree that with you than grow this diff
  again unasked.
- **On macOS a failed switch now activates nothing at all**, because the generic raise is suppressed
  for pane-keyed requests. That is deliberate — the alternative cold-starts Orca for a dead session —
  but it means a click can look like it did nothing, with the reason only in `focus-debug.log`.
- Adds a runtime dependency on the `orca` CLI: `PATH` first, then `%LOCALAPPDATA%\Programs\orca` on
  Windows and `/usr/local/bin/orca` or `~/.local/bin/orca` on macOS, since a Finder-launched app
  inherits none of the shell's `PATH`. Without it the window still raises and only the tab switch is lost.


rullerzhou-afk:
Thanks for taking this on — the overall direction looks solid. In particular, using the durable `ORCA_PANE_KEY`, resolving runtime-scoped handles through `orca terminal list`, retrying stale handles once, and separating the Orca window raise from the internal pane switch all make sense.

I reviewed exact head `56c7dd7624098f9defc3849d8a33dc129ae4c514` in a detached worktree and traced the hook → route → session state → focus → Telegram Direct Send paths. The focused suite passed 531/531, and the full suite passed 6243 with 0 failures and 18 platform skips.

I found three behavioral cases that I think should be addressed before merging:

1. **Direct Send can paste into the previously active Orca pane and still report success.**

   The Windows focus promise is resolved when the Orca window itself is confirmed in the foreground. `scheduleOrcaPaneFocus()` is fire-and-forget and starts later, while Direct Send waits a fixed 1200 ms before sending Ctrl+V.

   I reproduced four cases using the real focus and Direct Send code with only PowerShell/Orca CLI mocked:

   - cold `terminal list` taking 2.6s;
   - pane not found;
   - CLI timeout;
   - `terminal switch` failure.

   In all four cases Ctrl+V was still sent and the result was `pasted_without_enter` / `delivered: true`, even though the requested pane had not been selected. Not pressing Enter reduces the impact, but the reply is still placed in another session's composer and the success acknowledgement is incorrect.

   The safest short-term behavior would be to fall back to focus-only/clipboard for Orca. A complete fix would make the pane switch return an awaitable result and only allow Direct Send to paste after an **exact pane-key match** has switched successfully.

2. **An old Orca pane key can remain attached to an Antigravity session indefinitely.**

   `mergeOrcaPaneKey()` keeps the previous key unless the event is `SessionStart`, `sessionStart`, or `agentSpawn`. Antigravity emits none of those events. If the same conversation later runs from another terminal, its new `sourcePid` and `wtHwnd` are stored but the old `orcaPaneKey` remains.

   On Windows that stale key selects the `$orcaHosted` branch and gates off both the window cache and the otherwise-correct `wtHwnd`, so focus is redirected back to Orca and reported as successful.

   This needs a tested clearing rule — for example an explicit present/cleared/omitted state for the pane key, or a carefully scoped terminal-identity-change rule.

3. **The cwd fallback is ambiguous when one worktree has multiple terminals.**

   If the exact pane key is no longer present, `resolveOrcaHandle()` falls back to the longest matching `worktreePath`. When several terminals have the same worktree path, the current strict `>` comparison silently keeps the first entry returned by Orca and logs the switch as successful.

   That is useful as a best-effort manual navigation fallback, but it is not precise enough for Direct Send. Equal-best matches should fail as ambiguous, and the focus result should distinguish an exact pane match from a cwd fallback. Direct Send should only paste for the former.

One scope note: #743's reporter is on macOS, while this PR currently dispatches the Orca pane switch only from the Windows path. Could you change `Closes #743` to `Refs #743` or “Partially addresses #743” so the issue stays open for the macOS half?

If you add the macOS implementation, I would be very happy to test it on a real Mac. I am also happy to help complete the macOS window-activation/focus portion if that would be useful.

Thanks again — this is a substantial and thoughtful foundation, and I would be glad to re-review the follow-up.

KaiC5504:
All three are real. 1 and 3 are fixed; 2 isn't, and I want to explain why before you re-review. Head is now `6f98f09`.

**1 — Direct Send pasting into the previously active pane.**

Took the complete fix rather than the focus-only fallback. Falling back to clipboard for every Orca session would mean Direct Send never types into Orca again, and that's the case the field exists for — the defect is a missing completion signal, not something inherent to the pane switch.

`scheduleOrcaPaneFocus()` now returns `{ok, match, reason}` and never rejects. `requestWindowsFocus()` composes it onto the focus promise as `focusResult.orcaPane`, but only when the request carries a pane key: the pet and dashboard call sites don't await that promise, so nothing on the click path got slower and only Direct Send pays the round-trip. The adapter pastes only when `ok === true && match === "exact"`, and otherwise returns `orca_pane_unconfirmed` *before* the clipboard is written, so the existing clipboard fallback owns delivery and tells the user where the text went.

Your four cases are covered by tests, plus two more: an ambiguous worktree, and a switch that succeeded on a cwd match only. A cwd match deliberately does not clear the gate — it identifies the worktree, not the tab, and tab identity is the whole dependency.

One thing that fell out of this which you didn't raise: the handle cache stored whatever resolved, including a cwd guess, and on the next focus that replayed as an ordinary cache hit — indistinguishable from an exact match, so the gate would have passed on the second send. Only exact matches are cached now.

**3 — ambiguous cwd fallback.**

Equal-best `worktreePath` lengths now fail as `orca-pane-ambiguous` and switch nothing. `resolveOrcaHandle()` returns `{handle, match, failure}` with `match` as `"exact" | "cwd" | null`, which is what feeds the gate above. Length comparison is still sound for detecting the tie: two distinct paths of equal length cannot both be prefixes of one cwd, so equal length does mean the same worktree.

**2 — stale pane key on Antigravity sessions.**

Not addressed, deliberately, because the fix depends on your answer to the scope question at the end of the description. If the field is trimmed to Claude Code, Antigravity never carries a pane key and this case does not exist. If it stays broad it needs the clearing rule you describe — and the right shape of it (explicit present/cleared/omitted vs. a scoped terminal-identity-change rule) depends on which producers can be relied on to signal a session start at all, which is the same question. I'd rather write it once against the final surface. It's the only finding still open; say the word on scope and it goes in this PR.

**Scope note.** `Closes #743` → `Refs #743`, done.

**macOS.**

Implemented. The switch dispatches from `executeMacFocusRequest`, and the raise is `/usr/bin/open <absolute bundle path>` — `/Applications/Orca.app`, then `~/Applications/Orca.app`. Deliberately not `open -a Orca`: name resolution there is loose enough to reach OrcaSlicer, which is the same failure mode that made me pull the Linux `wmctrl` match. An exact bundle path can't, needs no Automation consent, and restores minimized windows.

It is stricter than Windows in one respect — the raise runs only after a confirmed switch. Windows raises from inside the generated PowerShell, so it fires regardless. On macOS a miss must never pull Orca forward, because `open` would *launch* it, and Orca's terminal daemon outlives its window: a sticky pane key pointing at a session whose window is gone would otherwise cold-start the IDE on a click. With no bundle found it logs `orca-app-not-found` and the switch still stands. Linux still doesn't dispatch.

I'd like to take you up on the Mac testing, because **none of the macOS path has touched real hardware** — bundle locations, LaunchServices behaviour and the raise-after-switch ordering are proven against mocks only. Most useful cases:

- pet click for an Orca-hosted session with Orca behind another app — right window forward, right tab;
- same with the Orca window minimized;
- same with Orca not running — expected: nothing launches, `focus-debug.log` shows the switch failing and no raise;
- Orca installed anywhere other than `/Applications` or `~/Applications` — expected `orca-app-not-found`, and I'd want to know if that's a real layout.

Direct Send is Windows-only (`platform_unsupported`), so the paste gate can't be exercised on a Mac by construction.

**Verification.** Windows re-verified on real hardware after the change — the switch lands on the correct tab and the focus result now waits on it. The full suite here is unchanged apart from the additions: the same 16 failures in the same 7 untouched files described in the description.

On your offer to help with window activation: macOS is covered, but Linux is still open and I'd rather it came from someone with the hardware, for exactly the reason I pulled it.

rullerzhou-afk:
Thanks for the detailed follow-up. I re-reviewed exact head `6f98f09b2aee5511ef3ebc3294e220054bf039a6`.

Findings 1 and 3 look resolved. Direct Send now waits for an exact, successful pane switch before pasting; ambiguous cwd matches are rejected; and keeping cwd-derived handles out of the exact-match cache closes the second-send edge case as well. In my environment, the focused suite passed 551/551 and the full suite passed 6263 with 0 failures and 18 platform skips.

On scope: please keep the broad producer coverage rather than trimming this to Claude Code only. With that decision, the Antigravity stale-key case remains in scope for this PR.

I do not want to prescribe the internal representation, but the regression case should cover the same normalized Antigravity session first reporting an Orca pane key, then reporting a fresh terminal identity without a pane key. The old key must no longer outrank the new `sourcePid` / `wtHwnd`, while ordinary later events from an unchanged terminal should retain the existing sticky behavior.

I also found two blockers in the new macOS path:

1. **The generic macOS focus path can raise Orca before the pane switch is confirmed.**

   `executeMacFocusRequest()` starts `focusTerminalWindowLegacy()` before `scheduleOrcaPaneFocus()`, and the Orca CLI path itself is delayed by 150 ms.

   I reproduced this through the public `focusTerminalWindow()` path, with `ps` returning the packaged Orca daemon path implied by Orca's current `fork()` implementation and with the external commands mocked. Even when `terminal list` returned no panes, the observed order was:

   ```text
   /usr/bin/open /Applications/Orca.app
   orca terminal list --json
   ```

   This means a missing or stale pane can still raise or cold-start Orca before the later switch fails. For an Orca-keyed macOS request, the Orca path should own the switch-then-raise sequence, and the macOS in-flight guard should remain active until that sequence settles.

   I think the public-path regression coverage should lock down three behaviors:

   - a missing pane never calls `/usr/bin/open`;
   - a successful `terminal switch` completes before `open`;
   - while the first Orca CLI sequence is still running, another focus request cannot start a concurrent pane switch.

2. **macOS CLI discovery currently depends on the inherited `PATH`.**

   Off Windows, `orcaCliCandidates()` only tries `orca`. Orca's current installer registers `/usr/local/bin/orca`, with `~/.local/bin/orca` as the fallback when `/usr/local/bin` is unavailable. A Finder/LaunchServices-launched Clawd process may not inherit a PATH containing either directory.

   Please make the macOS lookup robust to those official install locations—for example, by retaining the PATH lookup first and then trying `/usr/local/bin/orca` and `~/.local/bin/orca`.

This Mac does not currently have Orca installed, so I am not claiming a real Orca/LaunchServices smoke yet. The first issue is a reproduced Clawd control-flow problem using the runtime path implied by Orca's current daemon implementation; the second follows from Clawd's current candidate list and Orca's registered CLI locations.

Once these three items are addressed, I can run the real macOS smoke tests on my Mac—including the background, minimized, and not-running cases—and I would also be happy to help complete the macOS portion if useful. Thanks again—the two Windows correctness issues are now addressed cleanly, but I do not think the current head is ready to merge yet.

KaiC5504:
All three are in; head is `000a07d`.

**1 — the generic macOS raise firing before the switch.**

Confirmed and fixed. Reproducing it needed the fixture my suite never had: with `ps` reporting the packaged daemon, `extractMacAppBundlePath()` resolves an Orca session straight to `Orca.app`, and the ordering falls out exactly as you described —

```text
ps -o pid=,comm= -p 3333
/usr/bin/open /Applications/Orca.app
orca terminal list --json
orca terminal switch --terminal term_… --json
```

Without that fixture the bundle lookup found nothing and the raise silently skipped `open`, which is why the existing public-path test passed while the bug was live. That blind spot is the more useful half of this finding.

`executeMacFocusRequest()` now suppresses `focusTerminalWindowLegacy()` for a pane-keyed request and hangs `finalize` off the Orca promise instead, so the in-flight guard spans both CLI hops and the raise. Your three behaviours are locked down through `focusTerminalWindow()`: a missing pane never calls `/usr/bin/open`; a successful switch completes before the raise; and a second request cannot start a concurrent switch. That last one places the second request deliberately past `MAC_FOCUS_THROTTLE_MS` — inside the window the throttle defers it anyway and the test would prove nothing — and also asserts the queued request eventually runs, since a guard that never releases would deadlock the queue and still pass the narrower assertion.

One consequence worth stating plainly: with the generic raise suppressed, a failed switch on macOS now activates nothing at all. A click can look like it did nothing, with the reason only in `focus-debug.log`. That is the deliberate trade against cold-starting Orca for a dead session, and it is in the limitations list rather than left to be discovered.

**2 — macOS CLI discovery.**

Fixed: `PATH` first, then `/usr/local/bin/orca`, then `~/.local/bin/orca`. Gated to macOS rather than all non-Windows, since Linux does not dispatch this path at all and I would only be asserting install locations I cannot check.

This one also exposed a fixture bug worth mentioning: `missingBinaries: ["orca"]` stopped meaning "no CLI anywhere" the moment the candidate list grew, so the "degrades quietly when the orca CLI is not installed" test was quietly exercising a *successful* switch. Fixed the fixture, not the assertion.

**3 — Antigravity stale key.**

In, now that the scope question is settled. `mergeOrcaPaneKey()` compares the reported `sourcePid` / `wtHwnd` against the stored pair and drops the key when they differ; the event-name rule stays for the producers that do signal. Only values present on both sides count — most events carry no process metadata at all, and reading absent as moved would blank the key on the very next event — and the comparison is string-based because producers disagree on the wire type of the pid. Covered: a pid change, a `wtHwnd`-only change, the omitted-identity sticky case, and `100` vs `"100"`.

**Something you did not raise, found in daily use rather than in the suite.**

The description's latency claim was too generous and I have corrected it. On Windows the *first* focus of a session can land on the wrong tab: the raise is immediate, the switch is not, and a cold `terminal list` was killed at exactly the 6 s timeout with Orca already in front showing whichever tab was last active. The next click meets a warm CLI and lands in ~0.5 s. The suite cannot see this at all — it mocks the CLI, so cold-start cost does not exist there.

I have documented it rather than fixed it here, because the two candidate fixes pull in opposite directions: a longer timeout trades a wrong tab for seconds of apparent nothing, while pre-resolving the handle when a pane-keyed session first registers fixes it properly but adds a mechanism to a diff you have already called large. Your call whether either belongs in this PR.

**Verification.** 6313 tests / 6279 pass / 16 fail / 18 skip, the 16 in the same seven files this branch does not touch. I measured the pre-change baseline over the identical file set, so the +7 is arithmetic rather than an assertion. The Windows focus dispatch is untouched this round — the only Windows-visible edit is the candidate-list restructure, which returns the same list it did before — so the earlier live verification still stands. The stale-key rule is cross-platform but covered by tests only; I have not contrived a live terminal-move for it.

macOS is still mock-only. The offer to smoke-test on your Mac is very welcome, and the four cases from my last comment stand — with the CLI now findable without a login shell, the "Orca not running" case in particular should behave differently from before.

rullerzhou-afk:
Thanks again for the detailed follow-up and for addressing the earlier findings so carefully. I took you up on the macOS testing offer and tested exact head `4c5bcd622201f6ea10d8d2ac91204876caea5a3d` with Orca 1.4.161 on a real Mac.

The local Orca path worked: starting from the wrong active tab, Clawd's production focus path resolved the exact `ORCA_PANE_KEY`, completed the pane switch, and brought the correct Orca terminal forward. The installed app and bundled CLI were also verified on Apple Silicon.

I have prepared a follow-up branch with the remaining hardening and cross-producer coverage:

- Branch: `fix/pr759-orca-followups`
- Commit: `91081029b395b70c012706e8869c51ea9601f03d`
- Direct Send remains fail-closed unless an exact pane match succeeds.
- All focus-capable hook producers consistently preserve the Orca pane identity.
- Secure Orca-managed SSH sessions can request their local pane without treating a remote PID as local.
- macOS CLI discovery covers Apple Silicon Homebrew and the CLI bundled in `Orca.app`.
- Remote sessions remain excluded from Telegram Direct Send, and the Session HUD remains local-only.

Verification on the follow-up branch:

- focused Orca/focus/Direct Send suite: 180/180 passed;
- full suite: 6,349 total, 6,331 passed, 0 failed, 18 platform skips;
- real macOS local-pane smoke: passed.

One boundary remains: in a same-machine Remote SSH loopback smoke, the hook and exact pane key were correct and `orca terminal switch` returned success, but Orca's visible UI stayed on the local worktree. I am therefore not treating Remote SSH's visible switch as fully validated.

The Windows-specific changes on this follow-up branch still need a real-machine smoke. I will run that with the maintainer's Windows machine when it is available and report the result separately, so I am not calling the combined branch merge-ready yet.

Thank you again, KaiC—your implementation and detailed explanations made the remaining work much easier to isolate and verify.

rullerzhou-afk:
@KaiC5504 Thank you for the thoughtful Orca terminal-focus contribution and for carefully hardening the pane identity, focus ordering, and fail-closed Direct Send paths. We completed real-device validation on Windows and macOS, and the exact-pane behavior worked as intended. Your detailed commit history and regression coverage made this much easier to review and verify. Merged in 97a2a6a — much appreciated! 🙏

PR #764, opened 2026-07-28
fix(focus): match cmux workspaces by workspaceId so session focus reaches the right workspace

YOOGOMJA:
## Summary
- Session focus into cmux never selects the workspace: cmux is brought to the front, but the previously selected workspace stays active.
- `findCmuxPanelMatch` reads the workspace UUID from `ws.id`, while cmux writes it as `ws.workspaceId`, so every lookup returns `cmux-workspace-not-found`.
- Reads `workspaceId`, and corrects the cmux test fixtures to the schema cmux actually writes.

## Problem context
I hit this by clicking a session in the HUD: cmux came to the front, but I landed on the workspace I had left, not the one running that session.

That click runs `scheduleCmuxWorkspaceSwitch`, which resolves the terminal TTY and matches it against `~/Library/Application Support/cmux/session-*.json` to find the owning workspace and panel. On cmux 0.64.17 the match never succeeds, 20 out of 20 times in my `focus-debug.log`:

```
focus result branch=mac-open reason=ok bundle=cmux.app
focus result branch=cmux reason=cmux-workspace-not-found
```

A workspace entry in the session file carries `workspaceId`, not `id`:

```json
{
  "workspaceId": "A7804655-ABD6-409C-B59B-EE270C6E0DCA",
  "focusedPanelId": "DD445FB9-...",
  "panels": [ { "id": "DD445FB9-...", "ttyName": "ttys000", "type": "terminal" } ]
}
```

`typeof ws?.id === "string" ? ws.id : null` is therefore always `null`, and the loop `continue`s past every workspace. As a result, neither `focus-panel` nor the `select-workspace` fallback is ever reached. Panels are unaffected. cmux does key those by `id`, which is likely where the assumption came from.

This has been present since cmux support was added in #259 (issue #248) and was never caught, because `test/focus-cmux.test.js` builds its fixtures with `id` as well. The tests describe a payload shape cmux does not produce, so they stayed green while the runtime path was broken.

## What changed
- `src/focus.js`: read the workspace UUID from `ws.workspaceId`.
- `test/focus-cmux.test.js`: correct the five workspace fixtures to `workspaceId`, matching real session files.

## Validation
- Clicking in the HUD: I loaded this patch into the installed 0.13.0 app and clicked sessions in the session HUD. `focus-debug.log` recorded 9 `cmux-workspace-selected` results. That string appears nowhere in the log before the patch, where 20 of 20 cmux evaluations ended in `cmux-workspace-not-found`. Five failures remained, all for one pane that cmux itself no longer reports, which is #767 rather than this key mismatch.
- Live session file: the patched matcher resolves `ttys000` to workspace `A7804655-ABD6-409C-B59B-EE270C6E0DCA` and panel `DD445FB9-94E6-4590-B1A6-CFCA3D89BA06`, the workspace that hosts that terminal.
- `node --test test/focus-cmux.test.js`: fixtures corrected but code unchanged gives 4 failed, 4 passed, so the bug reproduces. After the code change, 8 passed, 0 failed. The unmodified `main` suite reports 8 passed because its fixtures use `id`, the key this bug never reaches.
- `npm test`: 5,957 passed, 13 failed, 18 skipped, identical to the `origin/main` baseline. Those 13 come from this environment and are unrelated to focus.
- `git diff --check`

## Platform note
Verified on macOS 26.5.2 with cmux 0.64.17 and Clawd 0.13.0. The click test ran against the installed release with this patch swapped in, not a fresh `electron-builder` package, and the app was restored afterwards.

## Scope control
After this fix, clicking a cmux-hosted session selects the owning workspace and focuses the exact panel instead of only raising the app.

This PR changes only the workspace UUID read and the matching fixtures. Untouched:

- TTY resolution and session-file discovery order
- the `focus-panel` to `select-workspace` to AppleScript fallback chain
- other terminal hosts, and the Windows and Linux paths

`scheduleCmuxWorkspaceSwitch` still returns early unless a process in `pidChain` has a `cmux` command name, the same host check used by the iTerm2, Ghostty, tmux, and Superset branches.

## Deliberate trade-offs (open to changing)
- No `id` fallback: every workspace entry I inspected carries `workspaceId`, and I found no evidence of a build that wrote `id`. Say the word if you know of one and I will add it.
- A workspace miss also loses the panel id, so the AppleScript `focus terminal id` fallback never runs even though it needs no workspace. Focus dies entirely instead of degrading, which is worth fixing separately.
- Reading the mapping from cmux instead of its session file would drop the private-schema dependency: `cmux debug-terminals` already reports `tty=ttys003 ... workspace=workspace:8 pane=pane:14`, and #248 proposed `CMUX_WORKSPACE_ID` via `ps eww`. Happy to open a follow-up if you want either.

Refs #259, #248

### DIFF
--- a/src/focus.js
+++ b/src/focus.js
@@ -1243,10 +1243,7 @@ function findCmuxPanelMatch(sessionData, ttyName) {
     const tm = win?.tabManager;
     if (!tm) continue;
     for (const ws of tm.workspaces ?? []) {
-      // cmux persists the workspace UUID as `workspaceId`; older payloads used `id`.
-      const workspaceId = typeof ws?.workspaceId === "string" ? ws.workspaceId
-        : typeof ws?.id === "string" ? ws.id
-        : null;
+      const workspaceId = typeof ws?.workspaceId === "string" ? ws.workspaceId : null;
       if (!workspaceId) continue;
       for (const p of ws.panels ?? []) {
         if (p?.ttyName === ttyName && typeof p.id === "string" && p.id) {
--- a/src/focus.js
+++ b/src/focus.js
@@ -1243,7 +1243,10 @@ function findCmuxPanelMatch(sessionData, ttyName) {
     const tm = win?.tabManager;
     if (!tm) continue;
     for (const ws of tm.workspaces ?? []) {
-      const workspaceId = typeof ws?.id === "string" ? ws.id : null;
+      // cmux persists the workspace UUID as `workspaceId`; older payloads used `id`.
+      const workspaceId = typeof ws?.workspaceId === "string" ? ws.workspaceId
+        : typeof ws?.id === "string" ? ws.id
+        : null;
       if (!workspaceId) continue;
       for (const p of ws.panels ?? []) {
         if (p?.ttyName === ttyName && typeof p.id === "string" && p.id) {
--- a/test/focus-cmux.test.js
+++ b/test/focus-cmux.test.js
@@ -55,7 +55,7 @@ describe("cmux panel focus (macOS)", () => {
   it("should call cmux focus-panel with matched panel UUID", (t, done) => {
     const panelId = "18AA1EB5-3055-445C-B780-60C88B21341B";
     const { tmpDir, cleanup: cleanupFile } = writeMockSessionFile([{
-      id: "ws-uuid-1",
+      workspaceId: "ws-uuid-1",
       panels: [{ id: panelId, ttyName: "ttys007", type: "terminal" }]
     }]);
     const origHome = process.env.HOME;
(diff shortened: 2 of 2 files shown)


rullerzhou-afk:
Thank you, @YOOGOMJA, for tracking this down so carefully and for the exceptionally thorough macOS/cmux validation. The workspaceId schema mismatch was subtle, and fixing both the runtime matcher and the fixtures gives us a much more trustworthy regression test. I've merged the PR. I also really appreciate you separating the remaining unmapped-pane behavior into #767 with clear evidence so it can be handled independently. Thanks again for the excellent contribution!

PR #795, opened 2026-08-02
feat(roam): add axis-constrained roam mode (#686)

### LINKED ISSUE #686: claude螃蟹可以增加仅可水平和竖直移动吗

weed33834:
## Summary

Add a new **Constrain to axis** setting that restricts the pet's idle wander movement to horizontal or vertical only, preventing diagonal movement.

## Problem context

Issue #686 asks: can the crab move only horizontally and vertically? When free roam is enabled, the pet currently wanders in any direction (including diagonally). Some users prefer a more "grid-aligned" movement pattern that looks cleaner on screen.

## What changed

- **prefs.js**: Added `roamConstrainAxis` boolean pref (default: `false`, no behavior change for existing users)
- **roam.js**: Modified `pickRandomTarget()` to constrain movement when the flag is on
  - Randomly picks horizontal (same Y, random X) or vertical (same X, random Y) each roam
  - Falls through to the regular random pick if the constrained target is too close
  - Exposed `setConstrainAxis()` to allow runtime toggle
- **main.js**: Wired up the pref to `_roam.setConstrainAxis()` with subscription
- **settings-actions.js**: Added the action mapping
- **settings-tab-general.js**: Added toggle switch in the Behavior section
- **settings-i18n.js**: Added translations for all 5 supported languages (en, zh, zh-TW, ko, ja)

## How it looks

New toggle in Settings → Behavior:
- "Constrain to axis" / "沿轴移动" / "沿軸移動" / "축 이동 제한" / "軸移動に制限"
- Description: "Only move horizontally or vertically, not diagonally. Works best with Free roam enabled."

## Testing

- All 32 roam tests pass
- All 164 prefs tests pass
- All 195 settings-actions/validators tests pass
- Syntax check clean on all modified files

Closes #686


rullerzhou-afk (changes requested):
Thanks for adding this setting. The pref, Settings action/UI, translations, startup initialization, and live subscription are wired coherently, and the default `false` preserves the existing free-roam behavior.

I reviewed exact head `a375a9b8d79fbb0da494e944dc8cb205cb17a0d4` with independent passes. I am requesting changes because the enabled mode still has deterministic paths that move diagonally, contrary to the UI promise.

## Required before merge

### 1. Keep every constrained-mode fallback constrained

`pickRandomTarget()` tries only one axis-aligned candidate. If that candidate is closer than `ROAM_MIN_DIST`, execution falls through to the existing two-dimensional random loop and then its corner fallback. Both coordinates can therefore change while `roamConstrainAxis` is enabled.

Deterministic reproduction: starting at `(400, 300)`, a too-close horizontal candidate followed by the current fallback produces `(1267, 670)`.

Please make the constrained branch own its complete retry/fallback behavior:

- generate axis-aligned candidates only;
- if the selected axis cannot provide a target at least 100px away, try the other axis;
- if neither axis has a valid target, return `null` and retry on a later roam cycle;
- never enter the normal two-dimensional picker or corner fallback while constrained mode is enabled.

### 2. Preserve the stationary coordinate through margins and final clamping

The horizontal branch clamps `bounds.y` into the inner 15% roam band, and the vertical branch similarly clamps `bounds.x`. If the user has placed the pet outside that inner band, the supposedly stationary coordinate changes too.

Deterministic reproduction: starting at `(400, 0)`, selecting a horizontal move produces `(1389, 162)`.

For every non-null constrained target, exactly one coordinate must remain equal to the walk's starting coordinate. Keep the stationary coordinate unchanged. If the final screen clamp cannot preserve that invariant, reject that candidate or try the other axis instead of correcting both axes.

### 3. Make enabling the live setting take effect immediately

`setConstrainAxis()` currently only flips the boolean. If the user enables the setting during an already-running diagonal roam, `animateTo()` keeps its previously captured diagonal target and can continue diagonally for several seconds. On the `false -> true` transition, please cancel/replan an active roam so the newly enabled restriction is immediately true on screen.

## Regression coverage

The existing 32 roam tests pass, but the PR adds no tests for `roamConstrainAxis` or `setConstrainAxis`. Please add deterministic coverage for:

- successful horizontal and vertical constrained moves;
- a first constrained candidate that is too close;
- a preferred axis with no valid target while the other axis is valid;
- neither axis having a valid target (`null`/later retry, never two-dimensional fallback);
- starting outside each inner margin, including a non-zero or negative work-area origin;
- enabling the setting during an active unconstrained roam;
- `roamConstrainAxis: false` retaining the existing free-direction behavior.

Please also add focused wiring coverage for the pref default/validation/round-trip, boolean Settings action validation, and the General-tab switch update/broadcast path.

No Settings hierarchy redesign is needed in this PR. We can handle making this switch visually subordinate to Free roam as a separate follow-up; this PR only needs to make the promised runtime behavior reliable and covered.

rullerzhou-afk (changes requested):
Thanks for addressing the three issues from the previous review. I re-reviewed current exact head `98aacf147b20a4b190b9c4cbe16b2e3736bf0801`. The constrained picker now owns its fallback path, preserves the stationary coordinate while selecting a target, and stops an in-progress diagonal roam when the setting is enabled.

There is still one axis-invariant blocker after target selection.

## Preserve the stationary coordinate through the final screen clamp

`pickRandomTarget()` now returns an axis-aligned target, but `animateTo()` subsequently passes both coordinates through `clampToScreenVisual()` and accepts both returned coordinates. If that final clamp corrects the stationary coordinate—for example, when the pet starts outside the normal rest-clamp region—the animation interpolates both axes and becomes diagonal again.

A deterministic harness reproduces this on the current head: start at `(400, -100)`, select a horizontal target, and model the final screen clamp correcting Y to `0`. The second applied animation frame is `(398, -98)`, so both X and Y have changed from the walk start.

The new tests do not cover this path because their default `clampToScreenVisual()` returns `{ x, y }` unchanged. Please make the final clamp axis-aware: preserve the chosen stationary coordinate, or reject/replan the target if the clamped result cannot keep the invariant. Add a regression test with a clamp stub that changes the stationary coordinate, and assert that every applied frame—not only the initially selected target—keeps exactly one coordinate equal to the walk's starting coordinate.

## Complete the requested wiring coverage

The previous review also requested focused coverage for the `roamConstrainAxis` pref default/validation/save-load round trip, its boolean Settings action, and the General-tab switch update/broadcast path. The follow-up commit adds only `test/roam.test.js` coverage, so these wiring checks are still absent.

The existing focused suites pass on this head (41/41 roam tests and 632/632 affected prefs/Settings/i18n tests), but neither suite exercises the remaining final-clamp invariant or directly covers the new Settings key. Once those cases are fixed and covered, this should be ready for another pass.

rullerzhou-afk (approved):
Re-reviewed exact head de5efa313245bc89d816b60c6afbac10f9dfaaa3. The axis invariant now survives the final screen clamp, the requested Settings/prefs wiring coverage is present, and the current-main merge test plus full suite are green. Approved.

PR #805, opened 2026-08-04
fix(cursor): preserve Windows hook paths and report safe chat titles

mantertius:
## Summary

Cursor's Windows PowerShell hook launcher can lose quotes around spaced Node or Clawd installation paths when the hook is wrapped in `cmd /s /c`. Call Node directly with PowerShell's `&` operator, recognize and migrate older encoded commands, and remove duplicate Clawd-owned entries while preserving third-party hooks and existing entry settings.

Send the native Cursor chat name through the existing `session_title` contract. Read only the matching conversation's bounded metadata from current or legacy SQLite layouts, using platform-specific standard profile paths. Missing, locked, corrupt, or unsupported databases fall back to the existing safe first-line prompt policy. SQLite remains optional on Node 22.12; remote hooks do not read the local database.

Silence only Node's SQLite experimental notice during the optional import: Cursor marks any stderr as a hook error even when stdout and exit status are valid. Other warnings remain enabled.

## Validation

- 264 related tests pass after merging current main, including installer idempotency/migration, actual Windows PowerShell stdin bridges, title privacy/layouts, hook contracts, manifest closure, and shared state routing.
- Node 22.12 tested both without SQLite (fallback) and with `--experimental-sqlite`.
- GitHub Actions passed all six Windows/macOS/Linux × project-Node/Node-22.12 compatibility jobs, plus the repository asset audit, on `610bdec95a9ea4d6ec1e33450e1c7da7d8ff1a7f`: [compatibility run](https://github.com/rullerzhou-afk/clawd-on-desk/actions/runs/34497672386), [asset audit](https://github.com/rullerzhou-afk/clawd-on-desk/actions/runs/34497672331).
- The Windows command-quoting/pipe fixture now has its own 5s read budget after hosted-runner diagnostics showed no bytes at the production 400ms deadline; the successful Node 22 run delivered the legacy bridge payload in 426ms. Production timing is unchanged, and all six jobs separately exercise the 11 stdin reader boundary tests, including the 400ms default. The existing production deadline can still drop input delivered too late under load.
- Real Windows Cursor 3.19.19 Agent chats used the exact repaired hooks under a `Clawd on Desk` path, alongside a spaced Node path. Prompt, tool, thought, and stop records succeeded in Cursor's Execution Log after the SQLite notice fix.
- Renamed the real chat to `PR805 真机验收`; Clawd Dashboard displayed that exact name and transitioned from running to completed.
- Full-suite checking also found 15 failures reproduced on unchanged main (symlink and DOM cases); the full suite is not reported as green. macOS/Linux automated compatibility passed on their native CI operating systems, without Cursor GUI acceptance on those platforms.


rullerzhou-afk (changes requested):
Thanks for tackling the Windows path-with-spaces failure. Switching the Cursor command to an encoded PowerShell invocation is a reasonable direction, but this revision introduces one deterministic startup regression and the session-title path needs compatibility/privacy hardening before merge.

## Required changes

### 1. Encoded commands break Windows registration idempotency

`registerCursorHooks()` still recognizes an owned entry with:

```js
entry.command.includes(MARKER)
```

After this PR, `cursor-hook.js` exists only inside the UTF-16LE base64 payload, so the outer command no longer contains the plaintext marker. The first legacy `cmd` -> encoded migration works, but every later startup/repair fails to recognize the encoded entry and appends another hook for each of the 11 events.

I reproduced this with the PR's command shape: the second registration reports another 11 additions and every event has two entries; a third registration grows them to three. Since enabled Cursor integrations are synced at Clawd startup, this causes repeated state events and unbounded `hooks.json` growth.

Please use the already imported `commandMatchesMarker(entry.command, MARKER)` here. It would also be useful to collapse extra Clawd-owned entries so Repair self-heals duplicates created by this revision.

Please add Win32 regression coverage for:

- fresh registration run twice: second result is `added: 0`, `updated: 0`, `skipped: 11`, the file is unchanged, and every event still has exactly one Clawd hook;
- a stale encoded command is updated rather than appended beside;
- uninstall removes encoded/duplicate Clawd entries while preserving third-party hooks.

The existing idempotency test only covers Linux, and the new Windows/legacy migration tests each run registration once, so they miss this regression.

### 2. Cover the Cursor database schemas and platform paths we support

The new test constructs only a normalized `composerHeaders` table, which proves that assumed schema but not compatibility with existing Cursor installs. On a real Windows Cursor 2.2.44 database I verified that `sqlite_master` contains only `ItemTable` and `cursorDiskKV`; there is no `composerHeaders` table, while `composerData:<id>` rows exist and some contain `name`/`subtitle`. In that layout the query always throws `no such table`, the empty catch hides it, and Composer names never work.

Please either define and enforce a minimum Cursor version for this feature, or detect/fallback across the known layouts (normalized `composerHeaders`, `ItemTable['composer.composerHeaders']`, and `cursorDiskKV['composerData:' + id]`) and test them with realistic sanitized fixtures.

`composerHeadersDbPath()` is also Windows-only: without `APPDATA` it builds `$HOME/AppData/Roaming/...`, which is wrong on macOS and Linux. Please add platform-aware paths and tests (`Library/Application Support` on macOS; `XDG_CONFIG_HOME`/`.config` on Linux), or explicitly gate and document the feature as Windows-only.

### 3. Do not turn common credentials into session titles

The prompt fallback creates a new propagation path from raw prompt content to Clawd's session title/HUD and optional companion surfaces. Its regex misses common bare credentials such as `sk-...`, `ghp_...` / `github_pat_...`, `AKIA...`, and long high-entropy secrets. `hooks/clawd-hook.js` already has a stronger `PROMPT_TITLE_SECRET_RE`; please reuse the same policy (or extract a shared helper) and add leak/false-positive tests.

## Compatibility follow-ups

- The project declares Node `>=22.12.0`, but `node:sqlite` still requires `--experimental-sqlite` on 22.12 and is unflagged only from 22.13. At the supported floor this silently falls back, while the test's `catch { return; }` is reported as a pass without exercising the feature. Please make the runtime requirement/fallback explicit and use a real skip or minimum-version CI coverage.
- The new hook helper should be accounted for under the repository's hook-helper/`HOOK_FILES` contract. It is packaged locally by `hooks/**/*`, but the current remote manifest closure test cannot see it because `cursor-hook.js` is not in that subset.
- Please complete the real Cursor Hooks-panel smoke tests listed in the PR and run the full suite; this branch currently reports no GitHub checks.

The `/state` -> state store -> HUD/Dashboard `session_title` consumption chain itself looks correct, and the legacy command migration/uninstall direction is otherwise sound.

rullerzhou-afk (approved):

PR #806, opened 2026-08-04
fix(focus): raise Cursor/Code when window title lacks project folder

mantertius:
## Summary
- Clicking a Cursor Agent session in the HUD/Dashboard currently dispatches focus but ends in `editor-parent-no-title-match` (no-op).
- Cursor Glass windows often use titles like `Cursor Agents` instead of the project folder name (`siga-horas`), so title matching never hits.
- Fall back to the editor process visible window / `MainWindowHandle` for `Code` / `Cursor` parents — same idea as Windows Terminal's pid-window fallback.

## Evidence
From `%APPDATA%\clawd-on-desk\focus-debug.log`:
```
agent=cursor-agent sourcePid=28040 cwdTail=...\siga-horas
reason=editor-parent-no-title-match status=unconfirmed
```
Process 28040's `MainWindowTitle` is `Cursor Agents`.

## Test plan
- [x] `node --test test/focus-windows.test.js`
- [ ] Manual (Windows): click Cursor Agent row in Session HUD → Cursor comes to front
- [ ] Manual: VS Code / Claude-in-editor session with mismatched title still raises the IDE

### DIFF
--- a/src/focus.js
+++ b/src/focus.js
@@ -289,6 +289,10 @@ function makeFocusCmd(sourcePid, cwdCandidates, focusCacheKey = null, wtHwnd = n
                 }
             }
         } elseif ($editorProcessNames -contains $proc.ProcessName) {
+            # Cursor / VS Code: prefer a window whose title contains the project
+            # folder. Glass / Agents windows often title themselves "Cursor Agents"
+            # (or a chat name) with no cwd — falling back to the process's visible
+            # window / MainWindowHandle is what users expect from a HUD click.
             $matches = @([WinFocus]::FindByPidTitles([uint32]$curPid, [string[]]$cacheTitleNames))
             if ($matches.Count -eq 1) {
                 [WinFocus]::Focus($matches[0])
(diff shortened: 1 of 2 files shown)


rullerzhou-afk (changes requested):
Thanks for working on this Cursor/Code focus fallback. Raising the IDE when the project folder is absent from the window title is a useful improvement, and the new PID/MainWindowHandle branches look appropriate for HUD and Dashboard clicks.

There is one safety boundary that needs to stay fail-closed before this can merge: please keep the four new editor fallback reasons out of WINDOWS_FOCUS_POSITIVE_REASONS. These branches run specifically after the project-title match failed, or when no title candidates exist, so foreground HWND equality only proves that a Cursor/Code window was raised. It does not prove that the mapped conversation, terminal tab, or input control is active.

focusDashboardSession passes this confirmed result to Telegram Direct Send, whose Windows adapter then sends Ctrl+V. The /focus-tab request cannot provide that confirmation here because it is fire-and-forget, its 200/404 result is ignored, and it only targets integrated terminals by PID rather than a Cursor Agent composer. This could paste a reply into another chat or the active code editor.

Could you please retain the new raise behavior but leave editor-parent-pid-window, editor-parent-main-window, editor-parent-pid-window-no-title, and editor-parent-main-window-no-title unconfirmed, as we already do for wt-parent-pid-window? It would also be good to add coverage showing that these reasons remain confirmed=false for Direct Send while HUD/Dashboard still dispatch the window raise.

Thank you again for improving the no-title Cursor case.

rullerzhou-afk:
Thank you, @mantertius, for identifying this Cursor Agents window-title mismatch, providing the reproduction, and proposing the PID-window fallback. We carried your idea forward in #1008, which is now merged, and credited your contribution in its description.

The merged implementation raises a unique visible Cursor/Code window, preserves ambiguity when multiple windows exist, and keeps the fallback unconfirmed so Direct Send cannot treat an app raise as a confirmed chat/input target. We validated the Cursor HUD/Dashboard paths, minimized-window recovery, title-match priority, and a real Claude Code session under VS Code on Windows.

Closing this PR as superseded by #1008. Thanks again for contributing the diagnosis and the original fix direction!

mantertius:
Thanks @rullerzhou-afk for picking this up and taking it further. Glad the diagnosis and direction were useful. Sorry for not being able to take the original PR all the way through, life got in the way, but it's great to see the idea land properly.

#1008 looks like a solid implementation! The PID-window fallback and the ambiguity guard are the right calls.

No concerns on the supersede. Thanks for the shout-out in the description. 🚀

PR #807, opened 2026-08-05
feat(dashboard): persist Sessions/Dashboard window bounds

### LINKED ISSUE #801: Sessions/Dashboard 窗口位置/大小未持久化（与 Settings 持久化保持一致）
问题概述

会话面板（Sessions / Dashboard）窗口没有将用户调整的窗口位置和大小持久化到偏好设置（prefs）。因此在用户移动或调整 Dashboard 后关闭并重新打开时，窗口会回到基于 pet 窗口或 Settings 窗口的计算默认位置/尺寸，而不是恢复为用户上次保存的几何。

重现步骤

1. 打开 Clawd → 打开会话面板（Sessions / Dashboard）。
2. 将会话面板拖动到任意位置并调整大小（例如放到屏幕边缘或改变为非默认宽高）。
3. 关闭会话面板并重新打开会话面板。

预期结果

- 会话面板应恢复到上一次关闭时的位置和大小（跨重启或重新打开保持一致）。

实际结果

- 会话面板每次打开都回到默认位置/尺寸（通常居中或基于 pet/Settings 的锚定逻辑），不会记住上次用户放置的几何。

已检查

- 仓库中已有 Settings 窗口的持久化实现（src/settings-window.js）和对应的 prefs 字段 settingsWindowBounds / 控制器路径 / 测试覆盖。Settings 的实现包括 getSavedBounds/onSaveBounds 注入、move/resize 事件上的 debounce 写回以及恢复时的 clamp/验证逻辑。
- Dashboard 的实现（src/dashboard.js）目前使用 computeInitialBounds() / computeSettingsAnchoredBounds() 根据 pet 或 Settings 窗口或居中来计算初始几何，但没有读取或写入 prefs（仓库中无 dashboardWindowBounds 字段）。
- 因此行为看起来是“仪表盘未实现持久化”，建议与 Settings 的实现对齐。

修复建议

- 在 prefs 中新增 dashboardWindowBounds 字段（nullable），并提供 normalize/validate（参照 settingsWindowBounds）。
- 在 src/dashboard.js 中：
  - 在初始计算时优先读取已保存的 bounds（通过注入的 ctx.getSavedBounds 或直接通过 settings controller），并做 workArea clamp；当 saved bounds 无效时回退到现有的 pet/anchored/center 逻辑。
  - 在 Dashboard 的 move/resize/close 事件上以 debounce 方式写回最新 normal bounds（通过注入的 ctx.onSaveBounds 或使用 settings controller applyUpdate("dashboardWindowBounds", bounds)），避免高频磁盘写入。
  - 保持 options.source === "settings" 时的 Settings-anchored 放置逻辑，但当用户手动移动后应覆盖并保存 saved 值。
- 添加相应单元测试以验证 normalize/clamp/保存行为（参考 test/settings-window.test.js 与 test/settings-controller.test.js）。

关联问题

- 参考已关闭的 Settings 相关 issue: #765（Settings 窗口位置/大小未持久化），该 issue 描述了 Settings 的缺失，Settings 已修复并有实现作为参考。

优先级/影响

- 用户可见体验：中等。会话面板是用户频繁打开的窗口，不能记住位置会降低可用性，尤其在多显示器或特定布局（把 Dashboard 放在某屏固定位置）下。修复工作量较小（复用已有 Settings 持久化模式）。

KaiC5504:
Fixes #801.

The Sessions/Dashboard window recomputed its placement from the pet or Settings window on every open, so a user-chosen position/size never survived a close. This persists normal-state geometry the same way the Settings window does since the #765 fix:

- New nullable `dashboardWindowBounds` pref (schema v13 → v14, no backfill — absent/null keeps today's computed placement), validated/normalized with the same helpers as `settingsWindowBounds`.
- Restore path: saved bounds win over pet-centering, clamped to the nearest work area with scaled minimums resolved from the saved rect's own display; invalid saved bounds fall back to the existing logic.
- Save path: debounced (500 ms) saves on move/resize of normal (not maximized) bounds, flush on `close` while native geometry is still queryable, dedupe against the last saved rectangle.
- `options.source === "settings"` anchored placement is kept, and programmatic re-anchoring (including any text-scale growth it triggers) rebases the persistence baseline, so it is never written over the user's saved standalone position — only a real user move after anchoring persists, matching the behavior requested in the issue.
- Constructor frame-quantization correction and the untouched-close-no-write baseline mirror the Settings window, and constructor minimums are capped to the clamped placement so a work area smaller than the scaled minimum can't force the window to overflow the display.
- Geometry reads are hardened like the sibling: degenerate work areas fall back to a sane rect, a throwing pet-bounds lookup can't kill the window open, and the move text-scale timer is cleared on close.
- `dashboardWindowBounds` joins `RENDERER_INERT_SETTINGS_KEYS` so bounds saves don't rebuild an open Settings tab.

Note: the persistence helpers are intentionally duplicated from `settings-window.js` rather than extracted, matching the existing duplication of `isUsableBounds`/`clampBoundsToWorkArea` between these two files. Happy to extract a shared module instead if you prefer.

Tested: full suite green (13 new dashboard tests, each verified to fail against the pre-fix code); live-verified on Windows — move/resize/close/reopen and full app restart restore the window, Settings-anchored opens still attach to Settings without clobbering the saved standalone position, a manual drag after anchoring persists, and maximize → close → reopen restores the normal rectangle.

### DIFF
--- a/src/dashboard.js
+++ b/src/dashboard.js
@@ -246,15 +246,16 @@ module.exports = function initDashboard(ctx) {
     const placement = getDashboardPlacement(options);
     if (isUsableBounds(placement.bounds) && typeof dashboardWindow.setBounds === "function") {
       dashboardWindow.setBounds(placement.bounds);
+      // The anchored placement can land the window on a display with a
+      // different textScale; re-zoom right away (memoized — cheap no-op when
+      // nothing changed). This can grow the window to that display's scaled
+      // minimum, so it must run before the baseline capture below.
+      applyTextScaleToWindow();
       // Programmatic anchoring is not user geometry: rebase the persistence
       // baseline (and drop any pending save) so re-anchoring never overwrites
       // the user's saved standalone position.
       clearSaveBoundsTimer();
       lastSavedBounds = getNormalWindowBounds(dashboardWindow) || placement.bounds;
-      // The anchored placement can land the window on a display with a
-      // different textScale; re-zoom right away (memoized — cheap no-op when
-      // nothing changed).
-      applyTextScaleToWindow();
     }
   }
 
--- a/src/dashboard.js
+++ b/src/dashboard.js
@@ -396,6 +396,13 @@ module.exports = function initDashboard(ctx) {
   function showDashboard(options = {}) {
     if (dashboardWindow && !dashboardWindow.isDestroyed()) {
       if (dashboardWindow.isMinimized()) dashboardWindow.restore();
+      // A debounce still pending on an open window holds the user's own last
+      // move. The re-anchor below drops that timer and rebases the baseline,
+      // so without flushing first the placement is lost for good — the close
+      // flush cannot recover it once the baseline matches the anchor.
+      if (options.source === "settings" && saveBoundsTimer) {
+        persistWindowBoundsNow(dashboardWindow);
+      }
       applySettingsPlacement(options);
       dashboardWindow.show();
       scheduleSettingsPlacementSync(options);
--- a/src/dashboard.js
+++ b/src/dashboard.js
@@ -8,6 +8,7 @@ const DEFAULT_WIDTH = 480;
 const DEFAULT_HEIGHT = 600;
 const MIN_WIDTH = 320;
 const MIN_HEIGHT = 400;
+const BOUNDS_SAVE_DEBOUNCE_MS = 500;
 const LIGHT_BACKGROUND = "#f5f5f7";
 const DARK_BACKGROUND = "#1c1c1f";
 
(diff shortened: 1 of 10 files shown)


rullerzhou-afk (changes requested):
Thanks for the thorough implementation and strong test coverage. The overall approach looks solid, but I found one persistence race that can still silently lose a real user placement.

Reproduction on exact head fe7c37b:

1. Start with saved standalone bounds P.
2. Open Dashboard standalone and move it to U, which schedules the 500 ms bounds save.
3. Before that debounce fires, open Sessions again from an already-open Settings window.
4. applySettingsPlacement() replaces the current bounds with the Settings anchor, then clears the pending save and rebases lastSavedBounds.
5. Close and reopen Dashboard standalone.

No save is performed for U, so the window restores the old P. The close flush cannot recover U because the current anchored geometry already matches the rebased baseline. I reproduced this deterministically both on the PR head and on the current-main integration result.

Could you please flush the pending user bounds before applying a programmatic Settings re-anchor, then keep the existing anchor/rebase behavior? Guarding that flush on an active saveBoundsTimer would avoid turning unrelated programmatic geometry into a user save. A regression covering manual move -> pending debounce -> Settings re-anchor -> standalone reopen would close the gap.

Everything else in the scoped review looked good, and the relevant test suites passed. Happy to re-review once this timing case is addressed.

KaiC5504:
Fixed in a9288475.

`showDashboard()` now flushes the pending save before the Settings re-anchor, guarded on an active `saveBoundsTimer` as you suggested. The anchor/rebase path itself is unchanged.

One deviation worth flagging: the flush sits at the `showDashboard()` call site rather than inside `applySettingsPlacement()`. That function also runs from `ready-to-show` on a freshly created window and from the 0/80 ms placement syncs, and neither can hold user geometry — an unshown window cannot be dragged, and a drag inside the 80 ms sync window is visually undone by the re-anchor anyway. A flush there could only promote post-rebase native frame drift into a user save, which is what the rebase exists to prevent. Happy to move it inside `applySettingsPlacement()` if you would rather have the guarantee in one place.

Regression test: "saves a still-debounced user move before the settings re-anchor drops it" in `test/dashboard.test.js`. It walks your exact chain — standalone move -> pending debounce -> Settings re-anchor -> standalone reopen — and also asserts the anchor itself is still not persisted. Verified failing on fe7c37b (it restored the stale saved rect, matching your report) and passing after the fix.

`test/dashboard.test.js` 33/33. Full suite 6900 tests, 3 failures, all pre-existing environment noise on my machine: my editor exports `CODEX_HOME`, which leaks into `codex-install` and `doctor-installer-defaults` expectations; those files pass 31/31 with it unset. On-device behavioral check is still pending on my side — I will follow up here once I have run it.

KaiC5504:
Following up on the on-device check: verified on Windows 11 at a9288475. Dragging the standalone Sessions window, re-opening it from an already-open Settings window inside the 500 ms debounce, closing it, then re-opening it standalone now restores the dragged position instead of the previously saved one.

rullerzhou-afk (approved):
Thank you for the thoughtful implementation and for following up quickly on the debounce/re-anchor race. I re-reviewed exact head a9288475 and independently verified the corrected persistence path on macOS, including reopen/restart behavior; your Windows 11 check also gives us the complementary platform evidence. The requested change is addressed and the full CI matrix is green. Approved — thanks for the contribution! We'll handle the remaining non-blocking edge-case coverage in a separate maintainer follow-up.

PR #810, opened 2026-08-05
feat: optional roam fence (limit where free roam wanders)

anthonyonazure:
Free roam currently uses the whole work area. This adds an opt-in fence: if `~/.clawd/roam-area.json` exists with `{ enabled, left, top, right, bottom }` (0..1 fractions of the work area), wander targets are clamped to that rectangle — e.g. keep the pet in the bottom-right quadrant. The file is read per target pick, so edits apply live without a restart, and the minimum hop distance scales down so small fences still produce targets. No file, no behavior change.

One of three independent PRs (roam fence / outlaw theme / dance tools) — take any subset, none depend on each other.


rullerzhou-afk (changes requested):
Hi @anthonyonazure — thanks for this! An opt-in roam fence is a genuinely useful idea, and expressing the area as work-area fractions makes sense.

I reviewed exact head `8e68653892d146b97ebfbd532ebcc70fd22d0a69` with deterministic target-selection and window-geometry repros. I’m requesting changes because the branch currently breaks its no-config compatibility promise, can place the actual window outside the fence, and has a semantic conflict with the axis-constrained roam logic now on `main`.

### 1. Please rebase onto current `main` and explicitly resolve the fence/axis semantics

This is not only a textual conflict. `main` now contains `roamConstrainAxis`, whose picker and final clamp preserve one coordinate exactly.

A mechanical integration can still pass all 48 existing roam tests while failing both of these cases:

- Axis mode continues using the fixed 100px minimum, so a small fence that works in free-direction mode may never produce an axis-aligned target.
- If the stationary coordinate starts outside the fence, preserving it leaves the final window outside the fence. If both coordinates start outside, no single-axis move can place the window fully inside.

Please define and test the combined behavior. A conservative rule would be:

- If the start is inside the fence, either axis may be selected.
- If only one coordinate is outside, that coordinate must be the moving axis.
- If both are outside, return no target for that round rather than silently moving diagonally.
- After screen clamping, validate both full-window fence containment and the axis invariant.

Tests should cover starts inside, X-only outside, Y-only outside, and both outside, asserting the final window and every applied animation frame.

### 2. Preserve the existing behavior when no active fence is applied

The adaptive `minDist` calculation currently runs unconditionally, including when:

- the file does not exist;
- `enabled` is false;
- the fence covers the full area;
- the file cannot be parsed.

Repro: with a 400×300 work area and a 250×170 pet at `(60,45)`, the parent commit returns no target because the farthest candidate is only 50px away. This PR lowers the threshold to 24px and starts a short walk without any fence file.

Please use the adaptive threshold only when a valid, enabled fence actually shrinks the candidate interval. Add regression tests pinning the missing, disabled, and full-range cases to the parent behavior.

### 3. Plan the walk using one frozen size

The picker subtracts `bounds.width/height`, but `animateTo()` may use a different `getEffectiveCurrentPixelSize()` for keep-size/mixed-DPI behavior.

Repro: in a 1000×1000 work area with a `.4–.6` fence, picker bounds of 100×100 and an effective animation size of 200×200 allow a target around `(499,499)`. The actual window then extends to approximately 699px, outside the 600px fence edge; the screen clamp does not know about the fence.

Please resolve one size snapshot before target selection and use it consistently for:

- fence geometry;
- random and fallback targets;
- final screen clamping;
- every animation frame.

The final full window rectangle should be revalidated against the fence.

### 4. Move parsing and validation out of the picker and cover the edge cases

The current empty `catch {}` fails open: a partially saved or malformed file temporarily restores full-area roaming. It also accepts coercible values through `Number()`, silently rejects reversed intervals, fails on a BOM, and rejects a valid corridor whose width or height exactly equals the pet size because of `xMax <= xMin`.

Our preferred architecture is a validated `roamFence` preference flowing through the existing prefs/controller/store path. If the external file is intentionally retained, please use a small injected loader with:

- strict object, boolean and finite-number validation;
- `0 <= left < right <= 1` and equivalent vertical validation;
- BOM handling;
- last-known-good behavior for transient read/parse failures;
- stable `ENOENT` meaning disabled;
- cached in-memory state rather than synchronous disk I/O inside target selection;
- a documented format and live-update timing.

For exact-fit corridors, only `xMax < xMin` or `yMax < yMin` is geometrically invalid; a zero-length interval on one axis can still allow movement on the other.

Please add focused fence tests for missing/disabled/full-range files, small fences, malformed and partial saves, exact-fit corridors, effective-size mismatches, fallback targets, negative work-area origins, and the fence/axis combinations above.

The existing roam tests remaining green is required, but they currently contain no fence coverage and do not detect these failures.

Thanks again — the feature itself is a nice addition; it just needs these invariants defined and locked down before it can merge safely.

anthonyonazure:
Thanks for the detailed review. All four points are addressed; the branch is rebased onto current main (ce387b8) and the fence is reworked on top of the axis-constrained roam.

**1. Fence and axis semantics.** Adopted the conservative rule as you specified: a start fully inside the fence may move on either axis; exactly one coordinate outside forces that coordinate to be the moving axis; both outside returns no target for that round (no diagonal moves). The stationary-axis restore runs before a final containment check, so both the axis invariant and full-window fence containment are enforced on the applied walk. Axis mode also gets a per-axis scaled minimum hop, min(100, max(24, range/2)), engaged only when the fence shrinks the interval, so a small fence that works in free-direction mode now produces axis-aligned targets too. Tests cover all four start cases (inside, X-only outside, Y-only outside, both outside) and assert the final window plus every applied animation frame.

**2. No-config compatibility.** The adaptive threshold now engages only when a valid, enabled fence actually shrinks the candidate interval. Missing file, enabled:false, a full-range rectangle, and invalid content all leave every interval and threshold bit-identical to the parent commit. Your 400x300 work area with a 250x170 pet at (60,45) repro is pinned as a regression test: no walk starts, with no file, with a disabled fence, and with a full-range fence.

**3. One frozen size.** pickRandomTarget now resolves a single size snapshot (keep-size effective size when exposed, else start bounds) before target selection and carries it on the returned target; fence geometry, random and fallback targets, screen clamping, and every animation frame use that one snapshot. After clampToScreenVisual and the stationary-axis restore, the final full-window rectangle is revalidated against the fence and the round is skipped if the clamp pushed it out. Your 1000x1000 work area, .4-.6 fence, 100px picker bounds vs 200px effective size repro now resolves to the single exact-fit target (400,400) and the real window stays inside the 600px edge; that exact scenario is a test.

**4. Parsing and validation.** Moved out of the picker into src/roam-fence.js, a small injectable loader with strict validation (typeof checks on the boolean and every edge, no Number() coercion of strings), BOM stripping, last-known-good behavior for malformed or partially saved content and transient read errors, and stable ENOENT-means-disabled. State is cached in memory; roam kicks an async refresh when it schedules a walk, so target selection never performs disk I/O. Format and live-update timing (an edit applies within one roam pause, about 4 to 8 seconds) are documented in the loader header. Exact-fit corridors are accepted: with an active fence only a negative pixel interval is rejected, and an interval collapsed to a point on one axis still allows movement on the other.

On the preferred prefs/controller/store architecture: I kept the external file deliberately, since the point of the feature is that outside tools can rewrite the rectangle live while the app runs without going through the settings store. The empty-catch fail-open you flagged is gone either way. If you'd still rather have this as a validated roamFence preference, I'm happy to convert.

Tests: 11 loader unit tests and 15 roam integration tests added (fence/axis start cases asserting every frame, small fences, exact-fit corridors, effective-size mismatch, clamp-escape skip, negative work-area origins, malformed and partial saves, live-update wiring). All 48 pre-existing roam tests and the full suite pass unchanged.

rullerzhou-afk (changes requested):
Hi @anthonyonazure — thank you for the thorough rework. The new loader, strict parsing, last-known-good handling, frozen-size fence planning, and the added fence/axis tests are all substantial improvements. I re-reviewed exact head `1b10a3b1e66e4ed20bf533e5eb8d7dc70986b07e`; the 74 focused roam/loader tests pass.

I found a few remaining edge cases that the new tests do not cover, so I’m requesting one more revision.

### 1. Keep an active fence even when it does not shrink the historical target band

`src/roam.js:145-153` currently sets `fenceRect = null` whenever the fence does not shrink the existing 15%-margin candidate intervals. That is safe for the adaptive-distance decision, but it is not equivalent to “no fence”: the pet’s starting coordinate may still be outside that rectangle.

Deterministic repro: 1920×1080 work area, 120×120 pet, active `.1–.9` fence, axis mode, start `(0,300)`, vertical axis selected. The historical target band is already inside the fence, so `fenceShrinks` is false and the fence is dropped. The walk finishes at `(0,734)` although the fence’s left edge is 192px.

Please retain the active `fenceRect` for start classification and final containment, and use the separate `fenceShrinks` boolean only to control the adaptive minimum hop. A regression test should use a non-full fence that encloses the old target band while the start lies outside it.

### 2. Preserve the parent picker geometry when no active fence exists

`src/roam.js:98-120` now uses the effective animation size to calculate target intervals unconditionally. The parent uses live bounds for picking and only uses the effective size during animation, so this changes the no-file path.

Repro: no fence/loader, axis mode, 400×300 work area, start `(100,100)`, live bounds 100×100, effective size 200×200. The parent starts a valid horizontal walk; this head finds no target on either axis and remains still. A second larger-work-area repro also selects a different target with identical RNG.

Please keep the parent bounds-based picker when there is no active, non-full fence, while using the frozen effective size for actual fence containment. The parent-compatibility test needs `bounds !== effectiveSize`; the current pins use matching sizes and therefore compare two copies of the new path.

### 3. Do not fail open during initial load or a transient ENOENT

The loader starts as inactive, and a single `ENOENT` immediately resets an existing valid fence to inactive (`src/roam-fence.js:76, 88-90`). That leaves two gaps:

- A replace-style save that briefly removes the path produces `active → inactive/full-area → active on the next refresh`, allowing one unrestricted walk.
- If the file exists but the first read is malformed, incomplete, or fails, the initial inactive state is also treated as permission to roam over the full area.

Please distinguish confirmed “missing/disabled” from “not loaded/error”. For example, retain the last-known-good fence across an isolated ENOENT and confirm/retry before disabling; before the first successful read, skip that roam round unless the loader has confirmed ENOENT or valid `enabled:false`. Tests should cover valid → transient ENOENT → valid and invalid-on-first-load.

Because this file is the only user interface for the feature, please also add the format/live-update contract to a user-facing guide and emit a deduplicated/rate-limited warning for invalid content. A small regular-file/size guard would also prevent a FIFO or accidental huge file from leaving the coalesced refresh pending indefinitely.

This is very close, and the core design is in much better shape now. If the remaining edge cases are more work than you intended to take on, no pressure at all — just say the word and I’m happy to finish them on top of your rebased commit while preserving your authorship and contribution. Thanks again for the careful follow-up.

rullerzhou-afk (changes requested):
Hi @anthonyonazure — thank you again for the remarkably quick follow-up. I re-reviewed exact head `3db8e121cee5bfc3b639003396cafc9bf076900e`. The two round-2 regressions are genuinely fixed: the non-shrinking fence is now retained, and the no-fence picker matches the parent when bounds and effective size differ. The expanded focused suite is also green (`85/85`).

There are a few small, localized cases left before I can approve:

### 1. A forced return into the fence must not use the ordinary minimum-hop threshold

When one coordinate is outside, `forcedAxis` correctly chooses that moving axis, but `tryAxis()` still applies the normal 24/100px minimum. If the pet is only slightly outside, every valid recovery target can therefore be rejected forever.

Repro: 1920×1080 work area, 120×120 pet, exact-fit X fence `[768,888]`, start `(767,450)`. X is only 1px outside and Y is valid, so the only legal recovery is X=768. The 24px minimum rejects it; after 20 seconds and repeated retries there are still zero applied frames and the pet remains outside.

Please let a forced containment-recovery move accept any non-zero legal displacement; keep the ordinary minimum hop for normal wandering after the start is already contained. Add the 1px-outside exact-fit case as a regression test.

### 2. Coalesced refreshes need one trailing read

`src/roam-fence.js:120-122` returns the current promise when a refresh is pending, so a newer refresh request can be lost completely.

Production-scheduling repro: a slow read captures the old `.1–.9` fence; while it is pending, the file changes to `.4–.6` and `scheduleNextRoam()` requests another refresh. That request is coalesced away. Eight seconds later the cache is still `.1–.9` and the walk targets `(672,672,120×120)`, outside the new `[400,600]` fence.

Please mark a refresh request received while pending and perform one trailing read after the current read settles (or otherwise make the scheduled walk wait for a freshness-guaranteed refresh). A delayed-read test should assert that the second state wins.

### 3. Non-ENOENT errors must break the consecutive-ENOENT streak

The implementation says two *consecutive* ENOENT results confirm removal, but a non-ENOENT error does not reset `enoentSeenWhileActive`.

Repro: `valid → ENOENT → EACCES → ENOENT` disables the active fence, even though the two ENOENT results are not consecutive. Please reset the streak on every non-ENOENT outcome, including read errors, and cover this sequence. A deduplicated warning for persistent read errors would also make the paused/stale state diagnosable.

### 4. Two finishing details

- The new non-shrinking-fence regression uses `Math.random() = 0.3`, which already selects the correct horizontal axis. Restoring the previous `fenceRect = null` bug still leaves the roam tests green. Please use a value such as `0.9` that prefers vertical, then prove the fence forces horizontal recovery.
- `src/roam-fence.js` references `docs/guides/roam-fence.md`, and the commit message says it was added, but that file is not present in this head. Please include the guide.

These are all contained finishing fixes; the overall design is now in good shape. If you would rather not spend another round on them, I’m very happy to take it from here and add one maintainer finishing commit on top of your work. Your original commits and attribution will remain intact. Just say the word — either way, thank you for being so responsive and careful with the feedback.

anthonyonazure:
All four round-3 items are addressed on the new head. Thanks for the 1px repro in particular, that one was a genuine stranding bug.

**1. Recovery hops.** A containment-recovery move (start rect outside an active fence) now accepts any non-zero legal displacement, in both the forced-axis path and free-direction mode; the ordinary 24/100px minimums still apply the moment the start is contained. Your exact-fit [768,888] corridor with the pet at (767,450) is a regression test: the pet walks the single legal pixel home, horizontally only.

**2. Trailing refresh.** A refresh request that arrives while a read is in flight now marks exactly one trailing read after the current read settles, so the newest file content always wins; the returned promise settles only after the trailing read. Your slow-read-then-edit scenario is a test (old fence captured mid-read, file replaced, second refresh requested: final state is the new fence). The coalescing test now pins "initial read plus at most one trailing read" for N concurrent requests.

**3. Streak reset.** Every non-ENOENT outcome, including read errors, breaks the consecutive-ENOENT streak. valid -> ENOENT -> EACCES -> ENOENT keeps the fence (tested), while a genuine consecutive pair still confirms removal. Persistent read errors now emit one deduplicated warning so the paused/stale state is diagnosable.

**4. Finishing details.** The non-shrinking-fence regression now mocks random=0.9, which prefers the vertical axis, so the test fails if fence retention regresses and only passes because the fence forces horizontal recovery. And you were right about the guide: docs/** is ignore-listed with per-file whitelists in .gitignore, so docs/guides/roam-fence.md was silently excluded from the round-2 commit. It's whitelisted and in-tree now.

Focused roam + loader suites and the full test suite are green with the new cases. Appreciate the offer to finish maintainer-side; happy to keep iterating here as long as it's useful to you either way.

One extra commit from a final self-review pass: the free-direction mode had the same 1px-stranding class as the forced-axis path. The fix already covered it, but no test bound on it, so there's now a free-direction regression (fully collapsed fence, pet 1px outside, walks the single legal pixel home), plus a defensive try/catch around the scheduling-time refresh kick.

rullerzhou-afk (changes requested):
Hi Anthony — thank you again for the careful round-3 follow-up. I re-reviewed exact head `f6c5a366cd7e779085dbf69c58619538499ff54c` with another independent adversarial pass.

The previous items are genuinely fixed: the 1px containment recovery now works in both axis-constrained and free-direction modes, refresh requests received during an in-flight read get a trailing read, non-ENOENT errors break the ENOENT streak, the adverse-RNG fence-retention test is binding, and the guide is present in the exact tree. The focused roam/loader suite is green (`90/90`), and a same-environment full run was also green (`6993` passed, `0` failed, `21` skipped).

Unfortunately, I found two remaining user-facing cases that still need changes before approval.

### 1. A valid fence in the outer 15% of the work area becomes a permanent no-roam zone

The active fence is always intersected with the historical 15% target band (`src/roam.js:117-164`). If the user deliberately places the fence wholly near an edge, that intersection can be empty even though the fence itself easily fits the pet.

Deterministic repro:

- work area: `1920x1080`
- pet: `120x120`
- active fence: `left=.9, right=1, top=.4, bottom=.6` (a 192px-wide valid strip)
- start already fully inside the fence
- free-direction roam, first 8s round plus four 4s retries

Result: zero applied frames forever. Here the fence requires `x >= 1728`, while the old margin band requires `x <= 1512`, so `xMin > xMax` and every round returns no target. The guide's “strip above the dock” example has the same failure class; a `.3-.7 x .87-1` strip that is tall enough for the pet also produces zero frames.

Please make an explicit active fence take precedence over the historical margin when their intersection is empty (or otherwise define a target domain that lets a geometrically valid edge fence work). Please add deterministic tests for a right-edge strip and a dock-adjacent strip, with the pet both inside and outside, asserting the final full window and every applied frame. The no-fence path should keep the historical margin behavior unchanged.

The guide also currently says a fence narrower or shorter than the pet only stops movement “on that axis.” With full-window containment, if either dimension cannot fit, no fully contained target exists and the current code stops the entire roam. Please either define and implement a different partial-containment product rule, or document the actual whole-roam behavior clearly; I would prefer keeping whole-window containment and correcting the guide.

### 2. Axis mode with both coordinates outside the fence freezes permanently, not only “for that round”

`src/roam.js:223-231` returns no target when both axes are outside. Because the window never moves, the next round receives the exact same geometry and returns no target again. Advancing the first 8s round plus four 4s retries still produces zero window writes.

The existing “both coordinates outside” test currently locks that permanent freeze in as the expected result. Removing only the early return is not sufficient either: the final full-fence revalidation in `animateTo()` rejects a one-axis partial recovery because the stationary axis remains outside.

One part of this is on us: our earlier conservative suggestion said to return no target “for that round” rather than move diagonally, but it did not define how a later round could ever recover. Sorry we did not call out that missing recovery phase earlier.

Please add a convergent recovery path. The least surprising option is two staged single-axis moves: recover one outside axis first, then the other on the next round, preserving the axis-only invariant on every frame. The regression should advance both stages, assert that each stage changes exactly one coordinate, and assert final full-window containment. If you prefer a one-time diagonal recovery, please make that product exception explicit and test it, but silent permanent freezing should not remain the behavior.

### Finishing notes (not separate approval blockers once the behavior/contract is explicit)

- **Live-update timing:** `refresh()` runs when the pause is armed, not when the target is picked. If the file changes one second into the pause, the already-pending walk still uses the old fence and the following walk uses the new one. That is internally coherent, but the guide and source comment currently promise that edits apply “within one roam pause (~4–8s).” Please change the contract to say that an edit applies after the currently pending walk. I am not asking for an async pick-time refresh in this PR; doing that safely would require an additional planning guard, post-await eligibility checks, and stale-completion handling.
- **File guard:** the regular-file/64KiB check is `stat(path)` followed by a separate `readFile(path)`. A path replacement between them can still swap in a FIFO and wedge the shared pending promise, or swap in an oversized valid JSON file and bypass the cap. This is a low-probability local robustness case, so I will not hold the PR on a full cross-platform file-handle rewrite, but the “FIFO can never wedge” comment is stronger than the implementation. Either read and bound the same opened handle (`fstat`, at most `MAX+1`, `finally close`) or narrow the comment and track the hardening separately.

The feature and the rework are both substantial, and the old no-fence behavior held across the differential cases I ran. These findings are limited to active-fence geometry and recovery, not a request to redesign the rest of the loader.

If you would rather stop iterating after this many review rounds, the maintainer finishing offer still stands: I am happy to add the two focused behavior fixes on top of your commits while preserving your original work and attribution. Either way, thank you for the unusually responsive and thoughtful follow-ups.

anthonyonazure:
Both blockers and both finishing notes are addressed on the new head.

**1. Edge fences.** The fence now takes precedence over the historical 15% margin band on any axis where their intersection is empty: interior fences keep the intersected band exactly as before, but a valid fence placed wholly in the outer band gets its own containment interval instead of a permanent no-roam zone. Your right-edge strip (.9-1 x .4-.6) and a dock-adjacent strip (.3-.7 x .87-1) are deterministic tests, each with the pet starting inside and outside, asserting every applied frame and the final window. The no-fence path is untouched, and the adaptive-hop gate now keys off "fence altered the band" rather than strictly shrank it, so edge fences also get correctly scaled minimum hops.

On the containment contract: kept whole-window containment as you preferred, and corrected the guide. It now states plainly that a fence smaller than the pet in either dimension has no valid position, so roaming stops entirely until the fence is enlarged, disabled, or removed; that behavior is pinned by a test. The misleading "on that axis" sentence is gone (the corridor case, where one dimension is exactly the pet's size, is still described and still works).

**2. Staged both-outside recovery.** You're right that "no target for that round" had no recovery phase; thanks for owning the earlier suggestion, but the missing convergence is on me for not noticing either. Both-outside now recovers in two staged single-axis walks: X first (the target is marked partial, and the final revalidation checks containment on the moving axis only), then the next round sees only Y outside and finishes with full-window revalidation. The freeze-lock-in test is replaced by a staged regression that advances both stages, asserts each stage changes exactly one coordinate on every applied frame, and asserts final full-window containment.

**Live-update contract.** Wording fixed in both the guide and the loader header: an edit applies to the next walk planned after it; a walk whose pause was already armed keeps the previous fence. No pick-time async refresh attempted, agreed that it needs its own guards and doesn't belong in this PR.

**File guard.** Took the handle option: the production reader is now a single non-blocking open (a FIFO can't hang the open), fstat on that handle (no stat/read swap window), a bounded read of at most MAX+1 bytes (an oversized file can't bypass the cap regardless of what stat said), and close in a finally. Guard violations behave like invalid content: last known good kept, one deduplicated warning. The injected-deps test seam is unchanged.

Focused suite and the full run are green with the new cases. Happy to keep going if the next pass finds more, and the finishing-commit offer is appreciated either way.

rullerzhou-afk:
Hi Anthony — a quick maintainer update, including why we took over this last stretch.

After four review rounds, the remaining work had shifted from the core fence idea into repo-specific hardening, Settings integration, and real macOS interaction testing. Rather than send you through another increasingly product-specific iteration, we decided to finish that last mile maintainer-side on your branch. Your original commits are unchanged and remain fully attributed; our work is contained in two separate commits on top:

- `e817d418` hardens the bounded file reader and closes the remaining recovery/full-range contract edge cases.
- `74ffce69` adds a visual picker under **Settings → General → Free roam**, where users can draw, move, resize, save, or disable the activity area.

The reason for adding the picker is product discoverability: during real-app testing we realized that most users would reasonably expect “choose an activity area” to be something they can draw in Settings, rather than a JSON file they must author by hand. The external `~/.clawd/roam-area.json` contract and your loader design remain the backend, so this extends rather than replaces your approach.

Human QA also caught an intermittent drag rollback caused by a delayed pointer-capture event; that is fixed in the picker commit. The focused fence/settings/picker matrix is green (`440/440`), and the macOS Electron preview passed draw, resize, repeated move, save, disable, and roam-behavior checks. CI is rerunning on the pushed head now, and we will re-check the exact head before merge.

Thank you again for the idea and for working through every review round so thoughtfully. The fence design and the bulk of the movement/loader work are yours; these maintainer commits are the final hardening and user-facing layer on top.

rullerzhou-afk (approved):
Final maintainer re-review on exact head d387013c35044815853719b6ab2b8ac95d1c63b0: all previously blocking findings are resolved. The focused activity-area matrix passes 461/461; the packaged macOS arm64 app passed native audit and real UI smoke for draw, move, resize, Escape cancellation, save, and custom-area removal. GitHub unit tests, repository audit, and all five package targets are green on this exact SHA. Anthony's original commits and attribution remain intact; maintainer commits finish the review findings and product UI. Approved.

PR #811, opened 2026-08-05
feat: Clawd Outlaw theme (cowboy hat, toggleable cigarette, bender idle)

anthonyonazure:
A new optional theme with some attitude: every stock sprite restamped with a pixel cowboy hat (creased crown, wide curled brim) plus a toggleable cigarette (pulsing ember, drifting smoke). The hat yields to any built-in headgear — hard hat while building, delivery cap while carrying — and comes off during sleep. Includes a 15-second bender idle animation: a bored Clawd drinks, smokes, sways, and passes out with floating Zzz.

`tools/outlaw` has the generator (accessories are defined once and injected into each sprite's animated body group, so they track every pose) and small CLIs for the cigarette toggle and roam-fence config. Running live on macOS; screenshots in the fork.

One of three independent PRs — take any subset.


rullerzhou-afk:
Hi Anthony — we really love the Outlaw theme, especially the bender idle animation. It has a lot of personality and we’d like to keep the concept and your authorship intact.

Before merging, we think the cigarette needs a small per-animation art pass: it doesn’t suit every pose, and in a few moving sprites it doesn’t stay naturally anchored to Clawd.

Rather than ask you to spend time fine-tuning dozens of animations, would you be comfortable with us adding a separate maintainer polish commit directly to this PR branch? We would keep your original commits untouched, limit our changes to per-sprite cigarette visibility and positioning, and show you the updated preview before merging.

If you would prefer to make the adjustments yourself, or would rather we work in a separate branch, that is completely fine too.

anthonyonazure:
Please do — a maintainer polish commit directly on this branch works great for me, and thank you for keeping the authorship intact. Per-sprite cigarette visibility and positioning is exactly the kind of art pass I'd rather have expert eyes on. Looking forward to the preview. One note in case it's useful: the smoke/ember animations were recently reworked to SMIL because CSS keyframe animations don't run when a sprite is rendered through an  channel, so any repositioned cigarette layers should keep the <animate>/<animateTransform> children with the ember and smoke rects.

anthonyonazure:
Pushed one commit before your polish pass: the smoke/ember rework mentioned above. The PR's sprites still carried the CSS-keyframe smoke, which never renders through an  channel (base state opacity:0, animation never runs — the same root cause as the cherry fix in the previous commit). Smoke and ember flicker are now SMIL (<animate>/<animateTransform> children inside the od-cig group), verified rendering and moving in img mode. Your per-sprite positioning pass can move the whole od-cig group freely; just keep the animate children with their rects and everything keeps working.

rullerzhou-afk:
Hi Anthony — thank you again for this wonderful contribution, and for trusting us with the maintainer polish pass. We’ve now completed and merged the final integration in #957. The shipped version keeps your cowboy-hat and cigarette artwork, preserves the original commits and authorship, and brings the bender animation into Clawd as a conditional idle easter egg. We added per-animation visibility, positioning, and attachment handling so the accessories fit each pose naturally. We especially loved the bender idea — it gives Clawd a huge amount of personality. Because #957 included your original commits, GitHub has marked this PR as merged as well. Thank you for the creativity, the thoughtful SMIL follow-up, and the smooth collaboration!

PR #812, opened 2026-08-05
feat: dance mode tools (party CLI + chiptune composer)

anthonyonazure:
`tools/dance`: `clawd-dance` shuffles any mp3s dropped in `~/.clawd/dance/tracks/` while holding the pet's headphones-groove state through the local state server (timed parties supported, Ctrl-C returns him to work). `make_song.py` is a pure-stdlib chiptune composer that renders a fully arranged 72-bar track (builds, two drops, breakdown, sidechain pump) in seconds; `make_track.py` is the minimal 8-bar loop version. No app code changes; audio files deliberately left out to keep the repo light.

One of three independent PRs — take any subset.

### DIFF
--- a/tools/dance/README.md
+++ b/tools/dance/README.md
@@ -1,0 +1,17 @@
+# Clawd Dance
+
+Soundtrack system for Clawd. `clawd-dance.sh` puts Clawd in his headphones-groove
+state (via the local state server on 127.0.0.1:23333) and shuffles the album in
+`~/.clawd/dance/tracks/`; `clawd-dance chip [min]` plays the homemade chiptune
+instead; a minutes argument auto-stops the party.
+
+- `tracks/` — instrumental album generated with Google Lyria 3
+  (downtempo, rave, soulful, cinematic ambient, and a West Texas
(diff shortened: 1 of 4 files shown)


rullerzhou-afk:
Thanks for putting this together. The idea of a music-reactive pet mode is fun, but after reviewing the current head, we do not think this implementation is ready to become a maintained Clawd feature, and the gap is broader than a small follow-up.

At the moment this is a macOS-only helper that runs afplay while independently posting a juggling state. Clawd does not know which track is playing or react to the audio itself. It is not integrated with Settings or a theme capability contract, its state is attributed to Claude Code, and both generators write to a hard-coded /Users/anthony path, so they fail for other users. With no playable track, the shell loop can also keep retrying a missing file.

A real music mode would first need a defined audio-source model, per-theme behavior and fallback rules, isolated lifecycle/session handling, actual audio or beat behavior, and clear cross-platform and licensing boundaries. That should begin as a product/design discussion rather than grow inside this PR.

I am closing this PR, but the underlying idea is welcome as a future Issue or Discussion once that scope is defined. These standalone scripts may also work better as an external experiment. Thanks again for the contribution.

PR #822, opened 2026-08-07
Add Brazilian Portuguese (pt-BR) locale

arismarioneves:
Registers **Brazilian Portuguese (`pt-BR`)** as a selectable UI language, with full key *and placeholder* parity against `en` on every localized surface.

Scope is deliberately narrow: **desktop locale only**. No README, no `docs/`, no PWA — following the one-topic-per-PR preference you set out on #528. Rebased onto current `main`.

### What's covered

| Surface | Keys |
|---|---|
| `src/i18n.js` | 459 |
| `src/settings-i18n.js` | 858, including the 19 function-valued ones |
| `src/bubble-renderer.js` — `BUBBLE_STRINGS` | 41 |
| `src/settings-ipc.js` — sound picker, agent discovery picker, theme removal | 3 dialogs |
| `src/settings-animation-overrides-main.js` — export/import dialog | 1 dialog |
| `src/codex-pet-main.js` — import and removal dialogs | 2 dialogs |

Plus the tables that silently fall back to English when a locale is missing: update bubble mode labels, Telegram completion notifications, the Telegram `/status` diagnostic card, and session-range pluralisation in the animation overrides tab — the last one needs its own branch because Portuguese inflects *sessão/sessões*.

For the function-valued keys, only the string literals changed; signatures and logic are identical to `en`.

### Registration

`SUPPORTED_LANGS`, the `lang` pref enum, the `settings-actions` validator (without it the picker silently reverts the choice), the General tab option list and label map, the permission automation confirmation window, and the tutorial language picker. `langPortugueseBrazil` was added to all six locales and, like the other language names, stays in its own language so a user who can't read the current UI language can still find it.

### Locale detection

`mapLocaleToLang` auto-selects pt-BR only for the `pt-BR` tag (`pt_BR` resolves through the existing normalization). Bare `pt` and other Portuguese variants fall back to English and can select it manually.

### Translation conventions

Hybrid, aimed at how Brazilian developers actually speak: terms with an established Portuguese equivalent are translated (session → *sessão*, theme → *tema*, shortcut → *atalho*, agent → *agente*, bubble → *balão*), while terms that are never translated in practice stay in English (hook, plugin, statusline, prompt, log, build). Product names are untouched.

### Not included, on purpose

`src/theme-metadata.js` has a `name: { en, zh, "zh-TW" }` label with no `ko` or `ja`. Adding only `pt-BR` there would deepen an existing inconsistency rather than fix it — happy to fill in all the missing locales in a separate PR if you want it closed.

### Tests

`npm test` → **7016 passing, 1 failing**. The failure is `test/focus-superset.test.js` ("returns the most recently opened workspace for a path"), which fails identically on `main` at ce387b8 in this environment (Windows) with no changes applied — it's unrelated to this PR.

Portuguese count grammar is covered by focused assertions in `test/i18n.test.js`: one for the entries that are functions and can inflect, one for the entries whose contract is a plain string and must therefore read correctly at 1 and at N. I checked they fail when the plural-only wording is put back.

Test changes were unavoidable: several assertions compared locale occurrence counts against a hardcoded `5`, and two `SUPPORTED_LANGS` deep-equals had the language list inline. Both now expect six. `test/prefs.test.js` also gained `pt-BR`, `pt` and `pt-PT` mapping cases.

I use Clawd daily with a Brazilian team, so I'm glad to keep reviewing the pt-BR wording as new strings land.


rullerzhou-afk (changes requested):
Thanks for keeping this PR narrowly scoped and for covering the locale registration and placeholder parity. I reviewed the current head (`e73a828`). The structural integration looks careful, but I found two issues to address before merge:

1. Several count-dependent Portuguese strings produce incorrect singular grammar. For example, a count of 1 currently renders forms such as `1 ativas`, `outras 1 ativas`, `1 problemas`, `1 precisam de atenção`, `1 ignorados`, and `1 novos` / `1 atualizados` / `1 removidos` in the Codex Pet refresh summary. Please make these strings singular-aware where functions are available, or use number-invariant wording where the locale contract must remain a string. Please also add focused pt-BR assertions for singular and plural counts so key/placeholder parity cannot hide this class of error.

2. The PR description says this is item 4 from the split requested on #528. Item 4 there was specifically the Spanish desktop locale, submitted by a different contributor; it was not a Brazilian Portuguese request. Please correct or remove that provenance sentence so the project history is accurate.

Once those are addressed, I will re-review the updated head.

arismarioneves:
Both fixed in `b61ee55`. Thanks for catching the grammar one — key and placeholder parity pass on `1 ativas`, so the checks I had could never have surfaced it.

**1. Singular grammar.** Every case you listed was real, and auditing for it turned up one more: `doctorAgentSummaryNeedsAttention` is appended after a name list that can hold a single agent, so it rendered `Claude Code precisam de atenção`.

Where the locale entry is a function, it now branches on the count:

| | count 1 | count 4 |
|---|---|---|
| `doctorAgentSummaryAttention` | `1 precisa de atenção` | `4 precisam de atenção` |
| `doctorAgentSummarySkipped` | `1 ignorado` | `4 ignorados` |
| `toastCodexPetsRefreshOk` | `1 novo, 1 atualizado, 1 sem mudança, 1 removido, 1 inválido` | `2 novos, 2 atualizados, …` |

Where the contract must stay a plain string, the wording is number-invariant instead:

| | before | now |
|---|---|---|
| `dashboardCount`, `sessionHudActive` | `1 ativas` | `1 em atividade` |
| `sessionHudOtherActive` | `outras 1 ativas` | `mais 1 em atividade` |
| `doctorIssueCount` | `1 problemas` | `1 problema(s)` |
| `toastAgentInstallHintPartial` | `1 instaladas, 1 falharam` | `1 instalada(s), 1 com falha` |
| `toastAgentCleanupHintPartial` | `1 removidas, 1 falharam` | `1 removida(s), 1 com falha` |
| `toastAgentInstallHintPartialSkipped` | `1 instaladas` | `1 instalada(s)` |
| `doctorAgentSummaryNeedsAttention` | `precisam de atenção` | `precisa(m) de atenção` |

`(s)` / `(m)` matches the convention already in the file — `en` uses `day(s)`, and `aboutCleanupSuccess` uses `item(ns)` / `integração(ões)`.

Two pt-BR tests in `test/i18n.test.js` cover this: one for the function-valued entries, one for the plain-string ones, each asserting the rendered text at a count of 1 and of N. I verified they fail when the plural-only wording is put back, so they aren't decorative.

**2. Provenance.** You're right, and I was wrong — item 4 on #528 was the Spanish desktop locale from @Zamaniego, not a Portuguese request. Nobody asked for pt-BR; I opened this because my team uses Clawd daily in Brazil. The description now refers only to the one-topic-per-PR preference you set out there, which is what actually shaped the scope. Sorry for muddying the history.

`npm test` on the updated head: 7016 passing. The single failure is `test/focus-superset.test.js`, which fails identically on `main` at ce387b8 in this Windows environment with no changes applied.

One open question from the description, in case it's easier to answer here: `mapLocaleToLang` currently routes every `pt-*` tag to `pt-BR`, so pt-PT and pt-AO users get Portuguese rather than English. Happy to narrow it to `pt-BR` only if you'd rather not imply European Portuguese support.

rullerzhou-afk:
Thanks — I verified `b61ee55`, and the two requested fixes plus the focused count tests look good.

For locale detection, please narrow automatic selection to `pt-BR` only (while still accepting normalized forms such as `pt_BR` through the existing normalization), rather than mapping bare `pt` and every `pt-*` locale to Brazilian Portuguese. The shipped locale is explicitly regional; users of other Portuguese variants can still select Português (Brasil) manually, but automatic detection should only select the variant we actually ship. Please update the mapping tests accordingly.

I will re-review once that small change lands.

arismarioneves:
Done in `37b5c48`. Auto-detection now claims only the `pt-BR` tag; `pt_BR` still resolves through the existing normalization, and bare `pt` / `pt-PT` / `pt-AO` fall back to English. Mapping tests updated to match.

`npm test`: 7018 passing, with the same unrelated `focus-superset` failure that reproduces on `main`.

rullerzhou-afk (approved):
Reviewed the current head (`37b5c48`). The requested singular/plural fixes, provenance correction, focused tests, and pt-BR-only auto-detection are all resolved. I also verified a clean merge against the current main and the directly related test suite passes. Thanks for the careful follow-through!

PR #836, opened 2026-08-09
feat: add native Slack notifications (webhook-first) (#832)

### LINKED ISSUE #832: [Feature Request] Add Slack integration for mobile/remote notifications
Amazing tool!

### Summary
Currently, `clawd-on-desk` supports Telegram and Feishu (Lark) for sending notification alerts and status updates. Adding **Slack** support would complete the core messaging platform trio, making it much more convenient for developers who rely on Slack as their primary workplace or personal workspace communication tool.

### Proposed Solution
Follow the same integration pattern used for Telegram / Feishu:
1. **Settings UI**:
   * Add a **Slack** configuration card under Settings -> Notifications / Mobile.
   * Allow users to input a **Slack Webhook URL** (for basic channel alerts) or **Bot Token (`xoxb-...`) & Channel ID** (for rich Block Kit messages/interactive notifications).
2. **Notification Service**:
   * Implement a Slack notification handler (e.g., `src/notifications/slack.ts`).
   * Trigger status updates, permission alerts, and task completion events to Slack via the Slack Incoming Webhook / Web API.
3. **Documentation**:
   * Add a setup guide in `docs/` explaining how to create a Slack App / Webhook URL.

### I'd like to contribute! 🙋‍♂️
I am happy to work on a Pull Request for this feature. Before I start, I'd love to get feedback from the maintainers on preferred implementation details (e.g., whether to start with Webhooks first or full Slack Bot Integration).

我已经在本地完成了一个初步版本的实现，但我也非常希望能先听听你的想法与建议，获得你的同意后再提交 Pull Request。

wang4433:
## Summary

Adds a native **Slack** notification channel alongside the existing Telegram / Feishu (Lark) integrations, as requested in #832.

Answering the open question in the issue (Webhooks vs. full Bot): this PR is **Webhooks-first** — the lightweight, zero-new-dependency path — with an optional Bot Token + Channel ID transport (`chat.postMessage`) also wired in. Full **Socket Mode interactive approval** (Allow/Deny resolved from Slack) is intentionally left as a follow-up, since it needs an app-level token + a persistent WebSocket client (mirroring the Feishu approach). The client already leaves a `supportsApproval = false` seam for that.

Closes #832

## What's included

- **Settings UI** — a Slack card in the remote-approval/notifications tab (next to Telegram/Feishu): Webhook URL, optional Bot Token + Channel ID, enable toggle, per-event toggles (done / error / permission), output-mode toggle, and a "Send test" button
- **Notifications** — task **done**, **errors/interruptions**, and **permission requests** are pushed as rich Block Kit messages with emoji. A webhook is one-way, so permission requests are *announced* ("approve in the desktop app")
- **Transport** — Incoming Webhook (primary) or `chat.postMessage` when a bot token + channel id are set; uses the built-in `fetch`, **no new dependencies**
- **Config & security** — secrets stored in a `0600` env file (`slack-notify.env`), never in plain prefs; webhook host pinned to `hooks.slack.com`; all outbound text run through the shared `secret-redact`; graceful degradation (never throws) when unconfigured or offline
- **i18n** — `slackNotify*` strings across en / zh / zh-TW / ko / ja / pt-BR
- **Docs** — README (EN + zh-CN) updated

## Files

New: `src/slack-notify-settings.js`, `src/slack-message-format.js`, `src/slack-notify-client.js` (+ 3 test files).
Wired into: `src/prefs.js`, `src/main.js`, `src/permission.js`, `src/settings-actions.js`, `src/settings-tab-telegram-approval.js`, `src/settings-i18n.js`, `README*.md`.

## Testing

- **30 new unit tests** (config normalize/validate/readiness + secret env I/O + masking, Block Kit formatting + redaction + truncation, client dedupe + per-event gating + graceful error handling) — all green
- Ran the **full existing suite** and diffed against a clean baseline: **no new regressions**
- Smoke-tested against a **live Slack Incoming Webhook** — all four calls (test / done / error / permission) returned `ok`

## Architecture notes

- Completion pings ride the existing `broadcastSessionSnapshot` fanout (same dedupe as the Telegram companion: `id:rawEvent:at`), so they're fire-and-forget on the sync broadcast path and never block it
- Permission announcements hook `addPendingPermission` (guarded, once per entry) and fire independently of whether an interactive channel is connected
- The Slack notifier is stateless — unlike the Feishu client there's no connection to manage; it reads current prefs/secrets on each send


rullerzhou-afk (changes requested):
Thanks @wang4433 — I reviewed the current #836 head (`8658f735`) and ran the three new Slack test files locally; all 30 tests passed. Webhook-first and one-way notifications are a good direction for phase 1.

I found two blockers that need to be addressed before merge:

1. **Auto-approved permissions can still produce a false Slack alert.** `addPendingPermission()` announces the request to Slack before the automation checks run, so Slack may say “approval needed” even though Clawd immediately approved it. The notification needs to happen only after the request is known to require a real user decision, with regression coverage for global/session automation and DND.

2. **Outbound redaction is incomplete.** Assistant output is redacted, but the session title, metadata, and top-level fallback `text` are not consistently passed through secret redaction and Slack text escaping. Since the title may derive from the user’s prompt, every user/agent-derived field—including fallback text—needs the same protection. Please add tests covering secrets and Slack mention syntax such as `<!channel>`.

A few additional changes are needed:

- Add `redirect: "error"` to outbound Slack requests.
- Strengthen webhook validation tests with a suffix-confusion case such as `evil-slack.com`.
- The field is presented as a Bot Token, but its validator currently also accepts `xoxp-` and `xoxe-`. For this phase, please either keep the transport webhook-only or constrain and document the supported bot-token form and required scopes clearly.
- Add a Slack setup guide with a privacy warning: messages go to a channel and may include the session title, folder, and host, so a private channel should be recommended.
- Update the Remote Approval page copy so it is unmistakable that Slack is notification-only and cannot approve requests in this version.

Thanks again for the substantial first cut. Once these correctness and privacy issues are addressed, I’ll re-review the updated head.

wang4433:
Thanks for the thorough review — all seven points are addressed in `40b822c5`.

## 1. Auto-approved permissions no longer produce a false Slack alert

You were right about the root cause: `addPendingPermission()` only means "the route accepted this request", and permission automation can consume it on the very next statement. Since Slack is one-way, no later message corrects the bogus "approval needed" ping.

The announce moved out of `addPendingPermission` to the two points where a human decision is actually known to be pending:

- **`showPermissionBubble`**, immediately after `maybeAutoApprovePermission` declines. The route's DND / per-agent / headless gates have already run by then, so this single chokepoint covers global automation, session overrides, and DND at once.
- **`maybeStartRemoteApproval`**, after `started` is confirmed — remote-only entries (bubbles disabled) never reach `showPermissionBubble`, and this point is past `maybeAutoResolveSessionPermission`. The existing `_slackPermissionAnnounced` guard keeps it idempotent for entries that already announced.

`announceSlackPermission` also re-checks `ctx.doNotDisturb` and pending-list membership. That's redundant given the call sites, but it means a future caller can't silently reintroduce the bug.

Regression coverage in the new `test/permission-slack-announce.test.js` (15 tests): global `unattended` and `auto-tools` auto-approval, session-override auto-allow, session override that *fails* the live gate (must still announce), DND, headless, passive codex notifications, plan reviews, announce-once-only, a throwing notifier, and the three remote-only cases.

**Why two call sites rather than one:** I looked at consolidating into a single helper called from the route, but the bubble path and the remote-only path diverge before automation finishes — `tryRemoteOnlyApproval` runs session automation itself and returns early, while the bubble path runs global automation inside `showPermissionBubble`. A single call site would have to sit after both, which is exactly where these two already are.

## 2. Outbound redaction and escaping now cover every user/agent-derived field

Split into two sanitizers, because Slack treats the two text types differently:

- `redactPlain` for `plain_text` header blocks — redaction only; escaping there would surface a literal `&amp;`.
- `redactMrkdwn` for `mrkdwn` sections **and the top-level fallback `text`**, which Slack parses as mrkdwn by default. That field was the real gap: it carried the session title unredacted and unescaped.

Applied to session title, folder, host, agent id, short session id, tool name, detail/summary, and both fallback strings. Assistant output inside the code fence is escaped too — a fence is not a mention-proof container.

Also added `clipMrkdwn`, since truncating escaped text can land inside an entity (`…&am`) and render as garbage; it drops the partial entity.

Tests cover secrets and `<!channel>` / `<!here>` / `<@U123>` in the title, metadata, permission fields, assistant output, and the fallback text, plus a truncation case that guarantees no dangling entity.

## 3–5. Smaller items

- **`redirect: "error"`** on both transports. Tests assert it for the webhook *and* `chat.postMessage`, since a redirect would move the `Authorization` header off the pinned host.
- **Webhook validation** — nine suffix-confusion cases, including `evil-slack.com`, `hooks-slack.com`, `hooks.slack.com.evil.com`, `hooks.slack.com@evil.com`, and the real host hidden in path/query/fragment.
- **Bot token constrained to `xoxb-`.** `xoxp-`/`xoxe-` are rejected, with the required scopes (`chat:write`, plus `chat:write.public` only for public channels the bot hasn't joined) documented in the validator, the settings hint, and the guide.

**Why constrain rather than drop the bot transport:** dropping it would leave workspaces that disallow Incoming Webhooks with no path at all, and the transport is already implemented and tested. Constraining the prefix makes the field do what its label says, which was your actual objection.

## 6–7. Docs and page copy

New `docs/guides/slack-notifications.md` covering setup for both transports, required scopes, secret storage, what is sent and when, and troubleshooting. It leads with the privacy warning — a table of every field that reaches the channel (session title derived from the user's prompt, folder name, host name), a private-channel recommendation, and a note that redaction is pattern-based and not a guarantee. Linked from `setup-guide.md` and both READMEs.

The Slack settings card now opens with a notice row above the setup steps: **"Notifications only — Slack cannot approve"** plus the channel-visibility warning, in all six locales. It's accent-tinted rather than error-red, since this is a standing property of the channel and not a misconfiguration. The enable and permission-request toggles reinforce it. Both READMEs now lead the Slack bullet with **notification-only**.

## Testing

- 25 new tests, all passing.
- Full local suite: identical failure set before and after this commit (27 pre-existing failures on my machine — hook installers and statusline adapters for agents I don't have installed, plus one caused by having the app running). `diff` of the two failure sets is empty, so no regressions. `test/mobile-preview-server.test.js` hangs locally on both base and head and was excluded.
- Verified end-to-end against a live Incoming Webhook: test, completion, and permission messages all delivered, and a deliberately hostile session title containing a token and `<!channel>` posted as `<redacted:token>` with the broadcast inert.
- Verified the auto-approve fix in the running app: with **Permission handling → Auto-approve**, a Bash request executed (the target file was created) and Slack stayed silent; with **Ask every time**, the same request produced both a bubble and a Slack card.

Socket Mode interactive approval remains out of scope for this PR, and `supportsApproval = false` is still the seam for it.

shengmai-justin:
Three commits are now on the branch. `40b822c5` is @wang4433's answer to your review; the two on top are mine and are **outside the scope of what you asked for** — they can be dropped independently without touching the review fix.

---

## Real-machine verification of blocker 1

`AGENTS.md` asks for real Claude Code verification on permission-path changes rather than hand-written `curl` payloads, so I ran the announce fix against a live session. Two phases, because "no Slack message" on its own proves nothing — it could just mean Slack was never wired up.

**Phase A — automation on (`auto-tools`)**

```
17:35:53  /permission hit | DND=false pending=0
17:35:53  showing bubble: tool=Write session=… stack=1
17:35:53  permission automation: mode=auto-tools action=auto-allow intent=tool-approval
17:35:53  response: {"hookSpecificOutput":{"decision":{"behavior":"allow"}}}
```

Real Claude Code raised **7 `PermissionRequest` events** (Bash ×5, Write ×1, Read ×1). All 7 hit `action=auto-allow`. **0 Slack permission cards.** On the pre-fix code each of those would have produced an "approval needed" ping for a request no human ever saw.

**Phase B — automation off**

Same Write request. The log line for the automation chokepoint is absent (the entry passed through it), a bubble is built, and Slack receives the card immediately.

The control was a Settings → *Send test* message before both phases, confirming the webhook was live throughout. So the Phase A silence is the fix working, not a dead integration.

## Two things a real workspace taught us that unit tests could not

Both were found by sending to an actual channel and looking at the result, and both had passing unit tests asserting the *wrong* behaviour beforehand:

1. **Top-level `text` is not a silent fallback once `attachments` are present.** Slack renders it as the message body *above* the attachment, so sending both prints the title twice. The summary belongs in the attachment's `fallback`.
2. **`fallback` is documented as plain text, but Slack still parses control sequences there.** A probe carrying `<https://example.com|LINKED>` arrived clickable. It therefore has to keep the escaped string — mention syntax in a session title must not survive into the notification either.

A `plain_text` **header**, by contrast, really is inert: `<!channel>` renders as literal characters and notifies nobody. That asymmetry is why the escaping policy differs by block type, and it is now pinned by a test with the reasoning in a comment.

## Commits added on top (not part of the review)

**`372b3eb9` — scannable cards, and a frame per message.** Every permission card opened with the same `⏳ Permission needed`, and repeated the agent and tool as `Tool:`/`Agent:` fields below, so a busy channel was unreadable and the actual reason sat third. The header now carries `⏳ claude-code · Bash` and appears nowhere else; the description gets the body; the folder moves next to the hint. Four blocks become three, and push previews distinguish requests too. Messages are also wrapped in a coloured attachment — Block Kit draws no border, so cards ran into each other; green done, red interrupted, amber waiting on you, blue informational.

**`72eae98d` — optional `@`-mention.** With a Slack member id configured, Clawd `@`-mentions it on every message so notifications reach you in a channel you are not watching. Empty by default. This is the one place the formatter deliberately does **not** escape its output — an escaped mention is inert text that notifies nobody — so the validator is the security boundary rather than a formatting nicety: a strict `^[UW][A-Z0-9]{6,20}$` allowlist, checked on save, again in `normalizeSlackNotify`, and once more before `<@ID>` is built. Tests cover eleven hostile inputs including `<!channel>`, `U123>ping<!channel` and `U01234567|evil`.

Worth weighing: if the channel is private and only you read it — which the guide recommends — setting that channel to notify on all messages does the same job with no configuration. The mention earns its keep in a shared channel.

## One change I made to a test in `40b822c5` — please look at this

`permission announcements redact and escape every agent-derived field` asserted that `<@U123>` in a tool name appears nowhere in the payload. `372b3eb9` moves the tool and agent out of an mrkdwn body field and into the `plain_text` header, where escaping does not apply, so that assertion no longer holds as written.

I did not delete it. It now asserts the invariant per block type: secrets must be gone everywhere, mention syntax must be escaped in every mrkdwn-parsed field (section, context, fallback), and the header — verified inert above — is redacted but not escaped. **This is a security assertion whose shape changed, so it deserves a second opinion.** If you would rather the header stay escaped, the redesign commit is the one to drop.

## Also worth knowing

- The completion card's title is the user's own prompt text verbatim. The privacy warning in the guide was written from reasoning; it is now observed fact, and it argues for the private-channel recommendation.
- Bot token: left exactly as @wang4433 had it (`xoxb-` only). I had prototyped removing the transport entirely for this phase, but that is his call and his commit, not mine to change.
- Full suite: **7074 passing**. The 4 failures are `after-pack-koffi.test.js` and reproduce identically on a clean `main` on this machine — macOS resolves `/var` through a symlink to `/private/var`, which that test's path check rejects. Unrelated to this PR; it should pass on the Linux runners.

rullerzhou-afk (changes requested):
Thank you both for the careful follow-up. I re-reviewed exact head `72eae98d` merged onto current main `5caf171b`.

The earlier findings around automation/DND permission timing, exact webhook host pinning, `redirect: "error"`, xoxb-only tokens, mrkdwn escaping, and the notification-only product boundary are addressed. I also reran the related merge-tree tests: 68/68 pass.

There are still production-path gaps that the current tests do not cover. To avoid another round of piecemeal feedback, please address this consolidated list in one pass:

### 1. Protect webhook secrets and disable link unfurling

A complete `https://hooks.slack.com/services/...` value still survives when it appears in a session title, assistant output, or permission detail, and reaches the final wire payload. Slack treats the whole webhook URL as a secret.

Please add both:
- a high-confidence Slack webhook URL rule to the shared redactor; and
- a last-mile scrub of the final outbound body against the currently configured webhook URL and bot token.

Also set `unfurl_links: false` and `unfurl_media: false` on both transports. Slack unfurls links by default for Incoming Webhooks and `chat.postMessage`, and explicitly calls out LLM-derived URLs as an exfiltration risk:
- https://docs.slack.dev/messaging/sending-messages-using-incoming-webhooks/
- https://docs.slack.dev/messaging/unfurling-links-in-messages/
- https://docs.slack.dev/concepts/security/

Please test the final serialized/fetch body, covering title, fallback, folder/host/agent metadata, assistant output, and permission detail.

### 2. Make delivery bounded and retry-safe

Multiple completions in one snapshot are currently sent fully in parallel. The dedupe key is committed before delivery, while HTTP 429 is classified but `Retry-After` is never read. A rate-limited or transiently failed notification is therefore lost permanently, and replaying the same snapshot does not retry it.

Please add a bounded per-destination FIFO, distinguish in-flight/pending from successfully sent events, honor `Retry-After` for capped 429 retries, and use bounded backoff for retryable network/5xx failures. Permanent 4xx responses must not loop, and queue/retry limits should be explicit.

There is also a startup-recovery edge: the recovered working snapshot is broadcast only to Dashboard/HUD, not to the Slack client. The first later Stop therefore becomes Slack's first snapshot and is swallowed by the unconditional priming branch. Please explicitly prime the notifier with the recovered snapshot, or otherwise distinguish pre-start historical completions from events that occur after startup. Tests should prove both "old done is not backfilled" and "recovered working -> new Stop is delivered."

Slack's posting and 429 behavior is documented here:
https://docs.slack.dev/apis/web-api/rate-limits/

### 3. Fix credential lifecycle and Settings state

The backend supports clearing a field with an explicit empty string, but the UI sends only non-empty fields and rejects both inputs being empty. Because a valid webhook always wins transport resolution, once a webhook has been saved the user cannot remove it or switch to the bot transport without manually editing `slack-notify.env`.

Please provide explicit per-field Remove/Clear actions (and preferably an explicit/resolved transport indicator), and cover webhook -> bot, bot -> webhook, clear-one, clear-all, and restart persistence.

Please also:
- validate non-empty webhook/token values before writing them;
- separate credentials-present, transport-configured, enabled, and ready states;
- allow Send Test while continuous notifications are disabled, as long as the transport is valid;
- surface stable localized errors instead of reporting every failure as "Slack rejected the message";
- show the saved mask for bot-only configuration; and
- mask the webhook input as a secret too.

### 4. Make permission notifications interaction- and route-aware

For `AskUserQuestion`, the Slack path currently runs the ordinary approval summary builder. The actual questions are omitted, the body falls back to "No description available", and the hint says "Approve or deny" even though the user must answer a question.

A `remoteOnly` entry also says to act in the desktop app, although by construction there is no desktop bubble and the usable action is in Telegram/Feishu.

Please carry route-owned interaction kind and action target into the Slack payload, reuse the elicitation payload/question data where appropriate, and distinguish at least:
- local permission -> approve/deny in the desktop app;
- human question -> answer the question;
- remote-only -> use the active remote approval channel.

Please add a real Claude Code validation for ordinary PermissionRequest and AskUserQuestion, as required for changes touching the Claude permission payload path. Also define and test the intended behavior when the local bubble fails to construct/load, since an Incoming Webhook message cannot later be retracted.

### 5. Complete the documentation and privacy disclosure

`docs/guides/slack-notifications.md` is referenced by both READMEs, the setup guide, and source comments, but is missing from the tree.

Please add the guide and make the English/Chinese setup path coherent. It should document both transports, scopes/channel membership, credential revocation vs local clearing, transport switching, troubleshooting, and the full outbound data boundary: session title, folder, host, permission summary/detail, and optional full assistant output. It should state that redaction is best-effort and that Slack retention/admin/channel access applies after delivery.

Please also correct the unconditional `0600` wording: POSIX files are `0600`; Windows relies on the current user's AppData ACL rather than Unix mode bits. Finally, clarify whether DND intentionally suppresses only permission Slack messages while completion notifications continue, because the current README wording says all hook events are silenced.

The full-output formatter should also report truncation after escaping/final clipping; currently `&<>` expansion can cause additional silent clipping while the label still says the output is complete.

### 6. Keep the two extra feature commits independently reviewable

The scannable permission header in `372b3eb9` is a useful improvement, and the mention validator in `72eae98d` is careful. However, both commits are explicitly outside the original review scope and are coupled through attachment-nested blocks.

Please preserve the contributors' authorship and split these two changes into follow-up PRs. If you prefer to keep them here, the remaining evidence becomes a merge requirement: real Slack desktop and mobile push behavior, VoiceOver/NVDA coverage for attachment-nested blocks without top-level `text`, and confirmed mention delivery for both webhook and bot transports. Slack documents secondary attachments as legacy and recommends layout blocks:
- https://docs.slack.dev/legacy/legacy-messaging/legacy-secondary-message-attachments/
- https://docs.slack.dev/reference/methods/chat.postMessage/

### Verification boundary

Automated tests can prove formatting, gating, queue/retry state, credential lifecycle, and final wire bodies. They cannot prove real Slack rendering or notification delivery. Please include a short real-workspace result for both transports covering test/done/error/permission, output off/full, clear/switch/restart, revoked credentials, a burst/429 case, mobile push, and actual mention delivery if the mention commit remains.

No additional feature expansion is requested here; the goal is to make the existing one-way Slack channel safe, switchable, and reliably testable before merge.

shengmai-justin:
All six items are addressed. Head is `19d1bba8`, merged with current `main` (`1deb50ca`) so the branch is clean again.

| # | Item | Commit |
|---|---|---|
| 1 | webhook redaction + disable unfurling | `82f2ee0b` |
| 2 | bounded, retry-safe delivery | `466f2667` |
| 3 | credential lifecycle + Settings state | `05b4772e` |
| 4 | interaction- and route-aware cards | `19d1bba8` |
| 5 | documentation + truncation label | `ab32572e` |
| 6 | split the two extra commits | see below |

## 1 — the webhook could leave in a message

Confirmed: the shared redactor had eight patterns and none matched a URL, so a `hooks.slack.com/services/...` string in a session title or assistant output reached the wire verbatim — published into the very channel it unlocks. Two layers now, because they fail differently: a rule in `secret-redact.js` (Telegram and Feishu inherit it), and a last-mile scrub of the serialised body against the *currently configured* credentials, which is the only layer that knows what the live secrets are and sees the complete payload. Both are load-bearing — removing either turns tests red, and with the redactor rule removed the scrub still catches the end-to-end case.

`unfurl_links: false` / `unfurl_media: false` on both transports. Tests assert on the final serialized fetch body across title, fallback, folder/host/agent metadata, full assistant output and permission detail, plus the direct `sendMessage` path where the formatter is bypassed entirely.

## 2 — delivery

The dedupe key was committed before the send, so a 429 or a blip lost the notification permanently and replaying the snapshot skipped it. Now a single bounded FIFO drained one message at a time. `lastNotified` means *settled*; in-flight state is tracked separately, so a repeat snapshot cannot double-enqueue and a failure is no longer mistaken for a send. 429 honours `Retry-After` (clamped); network/timeout/5xx get capped backoff; 401/403/404 are not retried. Queue overflow drops the oldest and logs it.

The startup-recovery edge was exactly as you described — the recovered snapshot reached Dashboard and HUD but not the notifier. `prime()` records what is already history without sending it; both halves are tested ("old done is not backfilled", "recovered working → new Stop is delivered").

## 3 — credentials

The root cause was one flag doing three jobs. `describeTransport()` separates credentialsPresent / transport / enabled / ready, and reports each credential individually so a bot-token-only setup no longer looks unconfigured. Each saved credential now has a masked line and a Remove button; removing the webhook is how you switch to the token. Values are validated before they are written — a malformed webhook would otherwise sit on disk masking a working token. Send Test needs a usable transport rather than the enable switch, and failures surface a stable code the renderer localizes. The webhook input is a password field.

Driving the real Settings window caught something the unit tests could not: `transport` was still read from `readiness()`, which reports none whenever sending is off, so a disabled-but-configured channel returned `transportConfigured: true` next to `transport: null`. Fixed, with a test pinning that they cannot disagree.

## 4 — questions, and the un-retractable message

The bug was worse than the hint wording. `buildRemoteApprovalSummary` looks for a description, summary or reason; an `AskUserQuestion` has none of those — it has `questions`. So it always fell through to `approvalSummaryUnavailable`, and Slack received "No description available" with the questions absent, under a card telling the reader to approve something whose `capabilities.allowDeny` is `false`.

The announce now carries `kind` and `actionTarget`. An answerable elicitation reuses `buildRemoteElicitationPayload` — the same data Telegram and Feishu render — so the card shows the questions and their options and says *answer*. `maybeStartRemoteApproval` is explicitly gated on `remoteOnly` rather than relying on the once-flag: ordinary bubbled entries reach that call too, and it produced the right answer before only because `showPermissionBubble` happened to run first.

Question and option text is redacted, escaped and clamped here. `buildRemoteElicitationPayload` passes `toolInput.questions` through untouched — each channel sanitises on its own side — so without this, agent text would have bypassed every protection added in item 1. Capped at five questions and five options with a "+N more" marker, matching the other channels.

**On the bubble failing after the heads-up has gone out:** a webhook message cannot be edited or deleted, and the request falls back to the agent's own prompt, so "approve in the desktop app" silently becomes untrue. We cannot retract, so we correct — `failPermissionBubble` sends a short follow-up saying the request went back to the terminal, only for entries that actually announced, at most once. My first instinct was to announce later instead, after the bubble exists; that turns out to be untestable, since `new BrowserWindow` throws without Electron and every announce test would become unreachable.

### Real Claude Code validation

Run against a live session with an isolated profile.

```
/permission hit | DND=false pending=0
ELICITATION: tool=AskUserQuestion session=...
/permission hit | DND=false pending=1
showing bubble: tool=Bash session=... suggestions=1 stack=2
```

Both card types confirmed in a real workspace:

- **❓ Answer needed** — the agent-supplied question header, the question text, every option as a bullet, and "Answer in the desktop app". No tool field, no approve/deny wording.
- **⏳ Permission needed** — `claude-code requests Bash`, the tool/agent/folder fields, the description, and "Approve or deny in the desktop app".

Earlier in the same branch, the announce-timing fix was validated the same way: 7 real `PermissionRequest` events under `auto-tools`, all `auto-allow`, **0 Slack cards**; with automation off, the card appears. A test message before both phases served as the control.

## 5 — documentation

The guide was missing because `.gitignore` excludes `docs/**` and re-includes individual guides by name, so `git add` silently skipped a new file — no error, clean `git status`. Both attempts hit the same trap. The one-line allowlist entry is the actual fix.

Three claims corrected: the unconditional `0600` (true on POSIX; Windows relies on the AppData ACL), the DND wording, and credential removal. On DND I verified rather than assumed — it is checked nine times in `server-route-permission.js` and drops the request before the announce, while `state.js` uses it only for animation, and the notifier has no DND awareness at all. So completions do keep arriving, which is the right behaviour (DND suppresses what asks you to act, not the "it's done" signal you left to wait for) but the README said the opposite.

The truncation label is also fixed. `prepareAssistantOutput` decided "truncated" on the raw text while escaping runs afterwards and expands: 2500 `<` characters fit the raw budget and become 10000 escaped, so the final clip dropped 7100 characters while the label still said the output was complete.

## 6 — the two extra commits

Split off as you asked, authorship preserved: **`wang4433:feat/slack-card-and-mention`** (`372b3eb9` scannable cards + framing, `72eae98d` opt-in mention). **They will come back as separate follow-up PRs, not here** — this PR is now only the review fixes.

Your point about attachments being legacy is a fair objection to the framing half specifically, and there is no Block Kit equivalent for the colour bar, so that one deserves its own discussion rather than riding along here. The scannable-header change and the mention validator have no such issue and may be worth splitting apart from each other when they are re-proposed.

## Two Slack behaviours worth recording

Both were found by sending to a real workspace, and both had passing unit tests asserting the *wrong* thing beforehand:

1. Top-level `text` is **not** a silent fallback once `attachments` are present — Slack renders it as the message body above the attachment, so the title printed twice.
2. Attachment `fallback` **is** parsed despite being documented as plain text; a `<url|label>` probe arrived clickable. It therefore keeps the escaped string, so mention syntax in a session title cannot reach the notification either.

A `plain_text` header, by contrast, really is inert: `<!channel>` renders literally and notifies nobody. That asymmetry is why the escaping policy differs by block type, and it is pinned by a test with the reasoning in a comment.

## What is verified live, and what is not

Being explicit rather than implying the whole verification boundary is covered:

| | Status |
|---|---|
| test / done / error / permission | verified — webhook |
| ordinary `PermissionRequest` + `AskUserQuestion`, real Claude Code | verified |
| assistant output off / full | verified — webhook |
| clear / switch | verified through the real Settings IPC |
| restart persistence | unit test only |
| **bot transport** | **not exercised live** |
| **revoked credentials** | **not done** |
| **burst / 429** | unit tested only |
| **mobile push** | **not done** |
| mention delivery | n/a — that commit moved to the follow-up branch |

The remaining rows need a Slack app with `chat:write` and a channel id, which I do not have set up yet. I would rather say so than let the list read as complete.

## Suite

7637 passing. One failure — `still shows the basename for ordinary QwenWork project cwds` — reproduces on pristine `main` on its own (7539 passing, same single failure), so it came with the merge rather than from it. The four `after-pack-koffi` failures reported earlier are gone; `main` fixed them.

rullerzhou-afk:
Maintainer fixes are now at 0c4b167f. Targeted regression tests pass 680/680; the full npm test run reports 7,760 passed, 0 failed, and 35 skipped. Real Slack, Electron, and Claude Code permission round-trip testing remains pending. Please validate the latest head.

shengmai-justin:
Validated `0c4b167f` against a real Slack workspace — both transports, driven through the Electron app, with live Claude Code permission round-trips. `npm test` on this branch merged with current `main` (`f0745407`): **7796 tests, 0 failed**; the merge itself is clean, no conflicts.

## Coverage

| | webhook | bot |
|---|---|---|
| credential save, `0600` env file, zero secret bytes in `clawd-prefs.json` | ✅ | ✅ |
| configured-but-disabled state (`transport` set while `ready:false`) | ✅ | ✅ |
| Send Test | ✅ | ✅ |
| completion done / error, output off | ✅ | ✅ |
| no title duplication, mobile push (`fallback` field) | ✅ | — |
| permission round-trip: real bubble → `behavior=allow matched=true` | ✅ | ✅ |
| restart persistence + no backfill of pre-restart completions | ✅ | — |
| queue cap, oldest dropped, every drop logged | ✅ 60 | ✅ 120 |
| credential failure → permanent, not retried, precise error class | — | ✅ |
| **429 + Retry-After** | not reachable | ✅ |
| transport switch (clear webhook → bot) | ✅ | |

Error/interrupted cards are **harness-driven**: a real `PostToolUseFailure` doesn't produce one (correctly — the tool fails mid-turn, the agent recovers, the turn ends `Stop`), so reaching that card needs `StopFailure`/`ApiError`, which I couldn't force by hand. Everything else above is app-driven.

## The 429 case — your fix was load-bearing

120 completions at once on the bot transport:

```
50 messages delivered · 63 HTTP calls · 13 genuine 429s · Retry-After: 1s ×13
```

All 13 were honoured and all 50 arrived. Slack answers 429 with a body of `{ok:false, error:"ratelimited"}`, and the pre-`324c3ed4` code parsed the body before checking the status:

```js
if (parsed && parsed.ok === false) return { errorClass: `slack-${parsed.error}` }; // "slack-ratelimited"
if (!res.ok) return { errorClass: classifyHttpStatus(res.status), retryAfterMs };  // unreachable
```

`slack-ratelimited` isn't retryable, so those 13 would have been classed permanent and dropped with `Retry-After` discarded. Worth noting webhooks never rate-limited (50 in 6.2s, zero 429) — this row is only obtainable on the bot transport.

Your split error classes also paid off on first contact: a missing scope and an uninvited bot surfaced as `slack-missing_scope` and `slack-not_in_channel`, each hitting its own localized string, each correctly permanent.

## Two defects — both pre-existing, not from your commits

**1. `outputMode: "full"` silently degrades to title-only.** Six consecutive real cards carried no `Assistant output` block while prefs and status both read `full`. Narrowed by elimination: feeding a synthetic entry that carries `assistantLastOutput` through the real client produces the block correctly, and the hook's own `buildStateBody("Stop", …)` returns `assistant_last_output` against the live transcript — so the field is empty on the real snapshot, somewhere between the state route and the snapshot entry. Intermittent: one earlier card did carry its output. `telegram-companion`'s `completionOutputMode: "full"` reads the same field.

**2. The session id on every card is a constant.** `slack-message-format.js` does `shortId(entry.id)` → first 6 chars, but `entry.id` is the namespaced key `s1.bG9jYWw.<uuid>`, so every local session renders `#s1.bG9`. With several sessions notifying, the disambiguator disambiguates nothing. `entry.rawSessionId` is already on the snapshot (`state-session-snapshot.js:144`); the desktop bubble uses it and shows something useful. `telegram-companion.js` has the identical line.

## Two notes on the delivery queue

Both reproduced with a probe against the real client; cut these if you'd rather keep the thread short.

**Any `slackNotify` preference change discards queued notifications.** `slackNotifyConfigRevision` is bumped by `subscribeKey("slackNotify")` and by every secrets write, and a mismatch cancels the item permanently (`settle()` records `lastNotified`). Toggling `outputMode` mid-drain dropped **4 of 6** queued completions as `cancelled (configuration changed)`. `destinationKey` already answers the cross-channel question precisely.

**Permission heads-up now shares the completion FIFO.** `notifyPermissionRequest` moved from a direct `sendMessage` to `enqueueAutomatic`, so the most latency-critical message queues behind the least. Under a 429 with `Retry-After: 30`, a permission ping enqueued at t=0 reached the wire at **t=180s**; with `isStillRelevant`, the realistic outcome is cancelled rather than late.

## Not covered

Revoked-webhook (`not-found`) specifically — though `slack-missing_scope` and `slack-not_in_channel` exercised the same permanent-failure path on real completions. Mention delivery is still on `feat/slack-card-and-mention` and out of scope here.

rullerzhou-afk:
Thanks again for the original Slack notification work here, @wang4433, and thanks to @shengmai-justin for the follow-up hardening work on this branch.

This PR is now in `main` via the recovery/integration PR #909, which was merged with a regular merge commit so the authored commits from this branch remain in history.

Merged via #909:
- merge commit: e4961f2568d5fdb5e3365597e6b050fae76f61ae
- recovered head: 6595ee4895b645bc90a99076580b4fb29db2d72f

Note: we did not run a fresh real Slack workspace smoke on the recovered #909 branch before merge, so the merged docs/release notes keep that boundary explicit.

rullerzhou-afk:
Follow-up after the #909 recovery merge: a fresh post-merge real Slack smoke was completed on `main`.

Verified the recommended Incoming Webhook path with both the Clawd send-test path and the local Clawd runtime `/state` completion path. Bot-token transport, mobile push, and permission-card smoke were not run in this pass.

PR #841, opened 2026-08-09
fix(opencode): show real session title in HUD and update it after OpenCode renames the session

### LINKED ISSUE #829: OpenCode: use the actual session title in HUD/dashboard instead of project folder name
Image

## Summary

Clawd currently appears to display a project/repository-derived name for OpenCode sessions rather than the actual OpenCode session title.

OpenCode Desktop already exposes and displays a per-session title. In my current session, OpenCode shows:

`Hy-MT2-1.8B prep for A/B test`

but Clawd does not use that session title in its HUD/session list.

## Expected behavior

When an OpenCode session has a title, Clawd's HUD/dashboard should display that actual OpenCode session title.

The project/repository/folder name can still be kept as secondary metadata if useful.

## Why this is different from #796

#796 concerns an incorrect or stale `cwd` / project path being associated with an OpenCode session.

This request is specifically about the display name: even when the `cwd` is correct, Clawd should use OpenCode's actual **session title** rather than deriving the visible title from the working directory or repository name.

## Environment

- Windows 11
- OpenCode Desktop
- Clawd on Desk v0.14.0
- OpenCode plugin enabled

xiaoshidefeng:
## Summary

Fixes #829. The HUD/dashboard was showing the **project folder name** for OpenCode sessions instead of the real session title (e.g. "Hy-MT2-1.8B prep for A/B test"). It also never updated when OpenCode renames a session.

Two problems were fixed:

1. **The OpenCode plugin never forwarded the session title.** The family plugin (`hooks/opencode-family-plugin/core.mjs`) builds the `/state` POST body with only `state`, `session_id`, `event`, `agent_id`, `hook_source` — no `session_title`. The server-side fallback chain in `sessionDisplayTitle()` then fell through to `path.basename(cwd)`, which is the project folder name.

   OpenCode already exposes the per-session title via `event.properties.info.title` on `session.created` / `session.updated` events (the field shown in OpenCode's own UI). The plugin now extracts it and forwards it as `session_title`, following the exact same pattern already used for `info.directory` → `body.cwd`.

2. **OpenCode swaps a placeholder title for the real one a few seconds after creation, and the update never reached Clawd.** A new OpenCode session starts with a placeholder title (`"New session - <timestamp>"`), then `session.updated` replaces it with the real summary-based title. Since `session.updated` maps to no Clawd state transition, the HUD kept showing the placeholder forever.

   The plugin now detects a title change and forwards it as a **metadata-only** POST (`metadata_only: true` + the new `session_title`), the same channel Claude Code's statusline refresh uses. The server's `updateSessionMetadata` now accepts a `sessionTitle` and updates the stored title without disturbing the lifecycle state.

## Files changed

- `hooks/opencode-family-plugin/session-ids.mjs` — extract `info.title` in `getEventSessionInfo()`
- `hooks/opencode-family-plugin/core.mjs` — capture the session title into `_sessionTitleById` (mirrors `_sessionDirectoryById`), include it as `body.session_title` in `buildStateBody()`, and push a metadata-only POST when the title changes
- `src/state.js` — `updateSessionMetadata` accepts `sessionTitle` and updates it without touching lifecycle fields
- `src/server-route-state.js` — the `metadata_only` branch forwards `session_title` to `updateSessionMetadata` (not gated on the Claude telemetry flag, since it isn't Claude statusline data)
- `test/opencode-family-core.test.js` — cover title capture + per-instance isolation
- `test/opencode-plugin-session.test.js` — cover `info.title` extraction

## Verification

- Live-tested against **OpenCode Desktop 1.18.15**: a new session shows the placeholder title, then auto-updates to the real title (e.g. `"New session - …"` → `"轻松问候"`) in the HUD a few seconds later.
- All related tests pass: **410/410** across opencode-family, state, server-route-state, and title tests.

Screenshots to follow.

image

### DIFF
--- a/hooks/opencode-family-plugin/core.mjs
+++ b/hooks/opencode-family-plugin/core.mjs
@@ -250,6 +250,9 @@ export function createOpencodeFamilyPlugin(config) {
   // Authoritative session directory learned from session lifecycle info.
   // Shared by every directory handler returned from this factory product.
   const _sessionDirectoryById = new Map();
+  // Authoritative session title learned from session lifecycle info.
+  // Shared by every handler returned from this factory product.
+  const _sessionTitleById = new Map();
   // Compatibility latch: old hosts that never emit info.directory keep the
   // latest-init fallback. Once this host proves it emits bindable session info,
   // a map miss omits cwd rather than forwarding a potentially stale directory.
@@ -466,17 +469,47 @@ export function createOpencodeFamilyPlugin(config) {
     return { directory: null, source: "none" };
   }
 
+  function captureSessionTitle(event) {
+    if (!event) return null;
+    switch (event.type) {
+      case "session.created":
+      case "session.updated":
+        break;
+      default:
+        return null;
+    }
+
+    const metadata = getEventSessionInfo(event);
+    const eventSessionId = normalizeSessionId(metadata.eventSessionId);
+    const infoSessionId = normalizeSessionId(metadata.infoSessionId);
+    if (eventSessionId && infoSessionId && eventSessionId !== infoSessionId) {
+      return null;
+    }
+
+    const sessionId = infoSessionId || eventSessionId;
+    if (!sessionId) return null;
+    if (!metadata.title) return null;
+
+    _sessionTitleById.set(sessionId, metadata.title);
+    debugLog(`SESSION_TITLE capture session=${sessionId} title="${metadata.title}"`);
+    return sessionId;
+  }
+
   function cleanupSessionDirectory(event, phase) {
     if (!event || typeof event.type !== "string") return;
 
     if (event.type === "server.instance.disposed" && phase === "before-send") {
       _sessionDirectoryById.clear();
+      _sessionTitleById.clear();
       return;
     }
 
     if (event.type === "session.deleted" && phase === "after-send") {
       const normalized = normalizeSessionId(getEventSessionId(event));
-      if (normalized) _sessionDirectoryById.delete(normalized);
+      if (normalized) {
+        _sessionDirectoryById.delete(normalized);
+        _sessionTitleById.delete(normalized);
+      }
     }
   }
 
@@ -573,6 +606,13 @@ export function createOpencodeFamilyPlugin(config) {
     if (isChildSessionId(clawdSessionId, _sessionParentById)) {
       body.headless = true;
     }
+    // Session title from OpenCode's own session-info title field. Mirrors the
+    // pattern used by other agents (clawd-hook, workbuddy-hook) that
+    // include session_title in their state POST body. The server reads
+    // this and stores it as sessionTitle, which sessionDisplayTitle()
+    // then uses before falling back to path.basename(cwd).
+    const sessionTitle = _sessionTitleById.get(clawdSessionId);
+    if (sessionTitle) body.session_title = sessionTitle;
     return body;
   }
 
@@ -675,10 +715,12 @@ export function createOpencodeFamilyPlugin(config) {
     buildStateBody,
     translateEvent,
     captureSessionDirectory,
+    captureSessionTitle,
     resolveSessionDirectory,
     cleanupSessionDirectory,
     get _sessionParentById() { return _sessionParentById; },
     get _sessionDirectoryById() { return _sessionDirectoryById; },
+    get _sessionTitleById() { return _sessionTitleById; },
     get _hostEmitsSessionInfo() { return _hostEmitsSessionInfo; },
     get _rootSessionId() { return _rootSessionId; },
     set _rootSessionId(v) { _rootSessionId = v; },
@@ -879,6 +921,7 @@ export function createOpencodeFamilyPlugin(config) {
           // not map to a Clawd state and info-only deleted events need its id.
           cleanupSessionDirectory(event, "before-send");
           captureSessionDirectory(event);
+          captureSessionTitle(event);
 
           if (sid && !_rootSessionId) {
             _rootSessionId = sid;
--- a/hooks/opencode-family-plugin/core.mjs
+++ b/hooks/opencode-family-plugin/core.mjs
@@ -502,7 +502,9 @@ export function createOpencodeFamilyPlugin(config) {
     // harmless — and it covers the "created with no title, titled later" case.
     if (prevTitle !== metadata.title) {
       _sessionTitleById.set(sessionId, metadata.title);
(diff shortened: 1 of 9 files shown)


rullerzhou-afk (changes requested):
Thanks — the direction is right, and I verified that the event-shape handling covers both the official `properties.info` form and the MiMo-style patch form. I am requesting changes at `cba97bf6`; please address the following in one pass:

1. **Blocking: move the new title test out of `test/opencode-family-core.test.js` and make it hermetic.** `captureSessionTitle()` now performs real I/O. The current test reads the real `~/.clawd/runtime.json`, POSTs to `127.0.0.1:23333-23337`, and appends to the real plugin log because this suite has no temp HOME or fake fetch. I confirmed four recorded `opencode:ses_t` / `My Title` requests from this test against a running Clawd, each accepted on port 23333 with status 204. Please move this coverage into `test/opencode-family-session-directory.test.js`, which already has temp-HOME, fake-fetch, fake-Bun, and `settlePosts()` infrastructure. Assert the complete metadata-only body, and assert that a blank title produces no fetch.

2. **Normalize and bound the title before storing or sending it.** `clawd-hook.js` and `copilot-hook.js` already normalize control characters/whitespace and cap titles at 80 characters. This plugin currently stores the raw title, sends it in the metadata update, and includes it in every later lifecycle body. I reproduced a 17,000-character title producing payloads over the 16 KiB server cap: `session.created` launched two five-port scans, and the next state launched another five-port scan. The server rejects before JSON parsing with a headerless 413, so the plugin distrusts the response, drops its cache, and repeats the scan while state updates remain dark. Apply the same bounded normalization at capture time and test both metadata and normal state bodies.

3. **Do not log title text.** `SESSION_TITLE capture ... title="..." (was "...")` persists user/LLM-derived content and allows embedded newlines/control characters to forge diagnostic log lines. Log only non-content metadata such as `session=<id> changed=true len=<n>`; add a temp-HOME assertion that the log does not contain the title body.

4. **Do not stamp `metadataUpdatedAt` for a title-only update.** That field is documented and tested as context/quota telemetry freshness. A rename currently makes stale telemetry appear fresh. The title does not need this stamp to broadcast: `sessionTitle` and `displayTitle` are already in `sessionSnapshotSignature`. I verified that removing the stamp would preserve the one title-change broadcast and same-title no-op behavior.

Please also add direct regression coverage for the new server/state branches:

- metadata-only `session_title` calls `updateSessionMetadata`, not `updateSession`;
- title remains allowed when the local Claude telemetry gate blocks context data;
- title + context in one POST becomes one metadata update, while an empty metadata payload causes no update;
- title-only state updates leave lifecycle state, `updatedAt`, recent events, badge, context usage/origin, and `metadataUpdatedAt` unchanged;
- same/normalized-equivalent title is a no-op, and an unknown session returns false without creating a ghost;
- the isolated plugin harness covers created/updated/dedup/deleted behavior plus official info-only and MiMo patch event shapes.

I also verified two follow-ups and am explicitly **not** asking you to expand this PR for them:

- Same-session `/state` POSTs are not causally ordered; a delayed lifecycle body carrying title A can arrive after metadata title B and restore A. This deserves a per-session FIFO/revision change with delayed-fetch tests, but it is a transport refactor and the next lifecycle event normally self-heals.
- `server.instance.disposed` is directory-scoped, while the shared factory clears process-wide directory/parent/title maps. The directory clear predates this PR, so directory-scoped ownership cleanup should fix all related maps together in a separate hardening change.

Metadata owner validation and Unicode-surrogate truncation can likewise remain separate follow-ups. Once the four items above and their regression tests are addressed, I do not consider FIFO or disposal cleanup a merge requirement for #841.

xiaoshidefeng:
Thanks for the thorough review! I've addressed all four blocking items plus the requested regression coverage in one pass.

**1. Hermetic title tests.** Moved the `captureSessionTitle` coverage out of `test/opencode-family-core.test.js` (which performs real I/O — my apologies for those stray POSTs) into `test/opencode-family-session-directory.test.js`, which already has temp-HOME, fake-fetch, fake-Bun, and `settlePosts()`. The suite asserts the **complete metadata-only body** — the process-tree enrichment fields that `postToClawd` injects are split off and sanity-checked separately (cwd = the session directory, agent_pid is numeric), and the remainder is `deepStrictEqual`'d against the exact title-push contract, so a missing or an unexpected field both fail. It also covers unchanged title → no fetch, blank title ignored, and bounded titles in both metadata and normal state bodies.

**2. Normalized + bounded titles.** Added `normalizeTitle` in `session-ids.mjs` (collapse control chars/whitespace, cap at 80 chars with ellipsis), mirroring `clawd-hook.js`/`copilot-hook.js`. Applied at capture time so it's bounded before storage, the metadata POST, and every later lifecycle body. Tests include a 200-char title that truncates to 80.

**3. No title text in logs.** The debug line now logs only `session=<id> changed=true len=<n>` — no user/LLM-derived content. Added a test asserting the log does not contain the title body.

**4. No `metadataUpdatedAt` stamp for title-only updates.** A rename no longer touches that telemetry-freshness field; the title still broadcasts because `sessionTitle`/`displayTitle` are in the snapshot signature.

**Regression tests added** for the server/state branches:
- metadata-only `session_title` routes to `updateSessionMetadata`, not `updateSession`
- title remains allowed when the local Claude telemetry gate blocks context data (using `claude-code` + gate-closed)
- title + context in one POST becomes one metadata update; an empty metadata payload causes no update
- title-only state updates leave lifecycle state, `updatedAt`, recent events, badge, context usage/origin, and `metadataUpdatedAt` unchanged
- same **and normalized-equivalent** title is a no-op (`"  Stable\t Title  "` collapses to the stored title — true, unchanged, no re-stamp); an unknown session returns false without creating a ghost
- the isolated plugin harness covers created/updated/dedup/deleted plus the official info-only and MiMo patch event shapes

All related tests pass: **422/422** (the two tightened test files alone: 286/286).

I've left the two non-blocking follow-ups (same-session FIFO ordering and `server.instance.disposed` directory-scoped cleanup) out of this PR as you suggested.

rullerzhou-afk (approved):
Re-reviewed current head 13b48629: the four blockers requested at cba97bf6 are addressed by 72f87095/13b48629, including hermetic coverage, bounded/sanitized titles, privacy-safe logging, and telemetry timestamp semantics. The FIFO and directory-scoped disposal follow-ups are isolated in #855. Approving the current #841 head.

PR #843, opened 2026-08-09
feat(agents): integrate QwenWork (千问办公) state-only hook agent

xiaoshidefeng:
## Summary

- Add QwenWork (千问办公) as a hook-only, state-only agent (Phase 1), modeled on the QoderWork integration. The pet reflects QwenWork lifecycle state, while permission decisions remain entirely in QwenWork's native flow: `PermissionRequest` / `PermissionDenied` are observation-only events and the hook always returns `{}`.
- Detect, install, and uninstall through `~/.QwenWorkCN/settings.json`, matching the desktop application's real user-data layout.
- Support the QwenWork desktop application on Windows and macOS only. QwenWork currently has no official Linux client, so this integration deliberately exposes neither a Linux process detector nor a WSL pairing/deployment entry.
- Use the portable Windows hook command form (unquoted forward-slash `node` token plus double-quoted arguments). PowerShell `-EncodedCommand` is recognized only to migrate Clawd-owned entries written by older builds; it is not the form Clawd writes.
- Wire QwenWork into the registry, Settings, Doctor, startup sync/repair, installation detection, assets, and marker-scoped install/uninstall cleanup. Third-party hooks are preserved.
- Add opt-in, privacy-preserving hook diagnostics. `CLAWD_QWENWORK_HOOK_DEBUG=1` records only event/status and bounded field-shape summaries; full payload logging additionally requires the explicit `CLAWD_QWENWORK_HOOK_DEBUG_RAW=1` opt-in. POSIX log files are constrained to regular, non-symlink targets with private permissions.
- Document the supported platforms, state-only permission boundary, installation lifecycle, and migration behavior.

## Verification

- **macOS:** the contributor's original real-machine verification confirmed installation detection, registration in `~/.QwenWorkCN/settings.json`, and an end-to-end hook POST (`posted: true`).
- **Windows real-machine E2E:** verified with the signed official QwenWorkCN 0.1.6 desktop app after logging in.
  - Settings **Install** registered all 10 lifecycle hooks and committed `integrationInstalled=true` / `enabled=true`.
  - A real QwenWork task created the requested file and emitted 18 hook invocations: SessionStart 1, UserPromptSubmit 1, PreToolUse 3, PermissionRequest 6, Notification 3, PostToolUse 3, and Stop 1. All 18 posted successfully to the live Clawd server.
  - No Clawd permission bubble appeared, as required for this state-only integration; the six real `PermissionRequest` events remained in QwenWork's native permission flow.
  - Settings **Uninstall** removed every Clawd marker and committed both preference flags to false. The pre-test Clawd preferences and QwenWork settings state were restored afterward.
  - Summary-only debug output contained no prompt text, requested filename, filesystem path, raw payload, or error text.
- **PR branch:** `npm test` — 7,309 tests; 7,274 passed, 0 failed, 35 skipped.
- **Merged with current `origin/main`:** `npm test` — 7,539 tests; 7,504 passed, 0 failed, 35 skipped.
- Focused QwenWork/cleanup/offline/cache/platform/snapshot/Doctor suites — 221 tests; 217 passed, 0 failed, 4 POSIX-only tests skipped on Windows.
- The four POSIX debug-log security cases passed separately under WSL on ext4. This validates POSIX filesystem behavior only and does **not** imply QwenWork Linux/WSL product support.
- `npm run audit:assets` — 0 errors; 1 repository-wide tracked-tree soft-budget warning also present on the base branch.

image


rullerzhou-afk:
Thank you for putting together the QwenWork integration and for doing the initial macOS/Windows investigation. I owe you an apology for how this maintainer follow-up happened.

I intended to ask Claude for an independent second review so that we could give you one consolidated set of findings. My handoff prompt was incorrectly scoped, and Claude implemented the proposed fixes on a local branch instead of stopping after the review. That was my prompting/workflow mistake, not a problem with your contribution.

Once I noticed, I did not want to push generated changes without treating them as maintainer-owned code. I reviewed the patch independently, reproduced the findings, found and fixed one additional POSIX debug-log target issue, and then ran the complete automated and real-machine validation. Since maintainer edits are enabled and the resulting patch held up, I have now fast-forwarded this PR branch by appending one maintainer commit, `f04ad4ce`; none of your existing commits were rewritten.

The follow-up closes the Settings/About uninstall path, updates the Windows offline/cache contracts, removes the unverified Linux/WSL entry, masks QwenWork internal workspace IDs, hardens opt-in debug logging, and documents the state-only permission boundary. It also corrects the PR description: the portable Windows command is not `-EncodedCommand` (that form is migration-only), and the previously reported four environmental test failures did not reproduce.

For the real Windows check, I installed and logged into the signed official QwenWorkCN 0.1.6 app, installed the integration through Clawd Settings, and ran an actual QwenWork file-creation task against a live Clawd server. All 18 emitted events posted successfully, including six real `PermissionRequest` events. No Clawd permission bubble appeared, which is the intended state-only behavior; QwenWork kept control of permission handling. Settings Uninstall then removed the Clawd markers and restored both preference flags. The PR branch and a clean merge with current `origin/main` both pass their full test suites with zero failures.

I am sorry for making a code change during what should initially have been a findings-only review, and for the surprise this may cause. Please review the appended commit and push back on anything you disagree with—we will treat any adjustment as a normal maintainer follow-up. The original integration gave this work a strong foundation; thank you again for contributing it.

rullerzhou-afk:
Final maintainer follow-up update: after the earlier f04ad4ce append, I ran another independent, findings-only review specifically to make sure our maintainer changes had not introduced regressions. That review found a few remaining edge cases in the maintainer patch, so I fixed them ourselves and appended 06e04771 as a normal fast-forward. No contributor commit was rewritten.

The final delta makes the registry capability authoritative before any generic /permission handling for state-only agents, adds the promised managed-uninstall backup for QwenWork, fails closed on malformed but parseable top-level or hooks JSON shapes instead of reporting a false install, keeps internal-workspace session titles private and distinguishable, and aligns the debug-permission documentation with the implementation.

This additional commit is maintainer-owned and follows from the review workflow mistake described in my previous comment. There is no request for you to rework your original contribution; please only review or push back if you would like to.

Final verification:

- targeted regression set: 391 tests, 0 failures, 4 POSIX-only skips
- final full suite after the last edge-case fixes: 7,319 tests, 7,284 passed, 0 failed, 35 skipped
- Windows real-config-directory smoke: 10 hooks installed, managed uninstall created a byte-identical pre-uninstall backup, retained a third-party hook, and removed every QwenWork marker
- the real ~/.QwenWorkCN/settings.json did not exist, so the smoke used uniquely named test-owned sibling files and removed them afterward; the real settings state was untouched
- two independent read-only Claude passes reported no remaining actionable finding in the final patch

Thank you again for the original QwenWork integration, and again I am sorry that a review request unexpectedly turned into maintainer code changes. We have kept all subsequent fixes and validation on the maintainer side.

PR #853, opened 2026-08-10
fix(opencode): avoid premature stale idle during long active work

### LINKED ISSUE #850: OpenCode: active agent can fall back to idle, releasing the sleep blocker
## Summary

With an OpenCode Desktop agent still actively running, Clawd can briefly show the correct busy/working state at the start of the task, then fall back to **idle** even though the OpenCode agent continues working.

This also breaks the existing **Block sleep while working** feature: once Clawd incorrectly considers the OpenCode session idle, the sleep blocker is no longer held, so Windows can enter Modern Standby during a long-running agent task.

## Reproduction

1. Use OpenCode Desktop with the Clawd OpenCode plugin enabled.
2. In Clawd, enable **Block sleep while working**.
3. Start a long-running OpenCode agent task.
4. Observe that Clawd initially becomes busy/working.
5. After some time, while OpenCode is visibly still running, Clawd returns to **idle**.
6. Run:

```powershell
powercfg /requests
```

Observed while OpenCode is still running:

```text
SYSTEM:
None.
```

7. If the machine is then left idle, Windows can enter Modern Standby even though the OpenCode agent is still active.

## Actual behavior

- OpenCode task starts -> Clawd may briefly show busy/working.
- OpenCode task continues -> Clawd may incorrectly return to idle.
- `Block sleep while working` therefore stops protecting the system.
- On a Modern Standby laptop, Windows subsequently entered standby due to idle timeout and the agent stopped progressing until wake/resume.

Observed Kernel-Power timeline:

```text
20:03:37  Event 506  Entering Modern Standby. Reason: Idle Timeout.
20:13:04  Event 507  Exiting Modern Standby. Reason: Input Keyboard.
```

The machine was plugged in and AC `Sleep after` was configured as `0` / Never.

## Expected behavior

While the root OpenCode session is genuinely active:

- Clawd should remain in an active working/thinking state.
- `Block sleep while working` should keep its `prevent-app-suspension` blocker held.
- Child/subtask/session lifecycle events should not make the root session appear idle prematurely.
- The blocker should only be released after the active OpenCode task/session is actually idle or finished.

## Related

This may be adjacent to earlier OpenCode lifecycle fixes:

- #366 - per-session state dedup / `sessionID` capture
- #401 - child/subtask `parentID` handling and headless marking

The current symptom suggests there may still be an OpenCode event/state transition that incorrectly resolves the active root session to idle.

## Environment

- Windows 11 ARM64
- Modern Standby: S0 Low Power Idle, network connected
- OpenCode Desktop
- Clawd on Desk v0.14.0
- OpenCode plugin enabled
- **Block sleep while working** enabled

PeterShanxin:
## Summary

OpenCode can legitimately spend more than five minutes in one silent tool/model segment. Clawd's generic five-minute working stale timeout incorrectly treated such a session as abandoned, forced it idle, and released `prevent-app-suspension`.

## Root cause

Live reproduction with:

```text
pwsh -NoProfile -Command "Start-Sleep -Seconds 420; Write-Output 'done'"
```

Before the fix:

- the tool started around 01:44:28;
- stale age crossed the generic `workingStaleMs=300000` threshold;
- at age `305656` ms Clawd returned `working-timeout`;
- the session was forced idle;
- `anyInProgress=0` and the power blocker stopped;
- the OpenCode tool continued running for about two more minutes.

Runtime evidence: https://github.com/rullerzhou-afk/clawd-on-desk/issues/850#issuecomment-5244153342

## Fix

Apply a bounded 20-minute stale floor to local, non-headless OpenCode working-like sessions, analogous to the existing local Codex protection for long silent active work.

- Process-death handling still takes precedence.
- Genuine OpenCode completion still transitions through `session.idle` / `Stop` immediately.
- Finite stale/session cleanup still provides a safety ceiling for abandoned work.
- Remote OpenCode, headless sessions, Codex, Claude, and other agents retain their existing policies.
- No tool-in-flight state machine or broader lifecycle redesign is introduced.

## Tests

Regression coverage includes:

- the exact pre-fix ~305-second OpenCode failure point;
- eventual stale retirement;
- dead agent/source process cleanup;
- genuine OpenCode completion;
- blocker-facing `isSessionInProgress` behavior;
- remote/headless OpenCode and other-agent isolation.

Final results: focused suites 530 passed, 0 failed, 0 skipped; broad `npm test` 7,521 passed, 0 failed, 35 platform-gated skips.

## Live verification

Windows 11 ARM64:

- reran the same 420-second silent tool;
- at about 5m30s OpenCode was still running and Clawd remained working;
- `powercfg /requests` showed Clawd's Electron request under `EXECUTION`;
- after the tool completed, Clawd returned idle normally.

## Scope

Broader OpenCode lifecycle hypotheses, tool-in-flight tracking, FIFO transport ordering, and unrelated issue work are intentionally out of scope.

Fixes #850

### DIFF
--- a/src/state-stale-cleanup.js
+++ b/src/state-stale-cleanup.js
@@ -23,6 +23,7 @@ function isLocalOpencodeWorkingLikeSession(session) {
   return !!session
     && session.agentId === "opencode"
     && !session.host
+    && !session.headless
     && isWorkingLikeState(session.state);
 }
 
--- a/src/state-stale-cleanup.js
+++ b/src/state-stale-cleanup.js
@@ -6,6 +6,7 @@ const SESSION_STALE_MS = 600000;
 const WORKING_STALE_MS = 300000;
 const DETACHED_IDLE_STALE_MS = 30000;
 const CODEX_LOCAL_WORKING_STALE_FLOOR_MS = 20 * 60 * 1000;
+const OPENCODE_LOCAL_WORKING_STALE_FLOOR_MS = 20 * 60 * 1000;
 
 function isWorkingLikeState(state) {
   return state === "working" || state === "juggling" || state === "thinking";
@@ -14,6 +15,13 @@ function isWorkingLikeState(state) {
 function isLocalCodexWorkingLikeSession(session) {
   return !!session
     && session.agentId === "codex"
+    && !session.host
+    && isWorkingLikeState(session.state);
+}
+
+function isLocalOpencodeWorkingLikeSession(session) {
+  return !!session
+    && session.agentId === "opencode"
     && !session.host
     && isWorkingLikeState(session.state);
 }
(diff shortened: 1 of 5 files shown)


rullerzhou-afk (approved):
Reviewed exact head 79fd0b0de0e467d522939319d603ce3d485f053f against current main, including the OpenCode FIFO and disposal changes from #855. The scoped stale-floor fix is correct: local interactive OpenCode stays active through the reproduced failure window, genuine Stop and SessionEnd plus agent exit remain authoritative, and remote, headless, and other-agent policies remain unchanged. The merged-tree OpenCode, FIFO, disposal, stale-cleanup, and state tests passed; the only local failure observed was a pre-existing unrelated QwenWork path assertion on macOS. Thank you for the careful reproduction and for providing the Windows 11 ARM64 live verification.

rullerzhou-afk:
Thanks again, @PeterShanxin — merged! Your preserved runtime evidence and the Windows 11 ARM64 420-second verification made the failure mode and the bounded fix unusually clear. This keeps local interactive OpenCode work active past the premature five-minute cutoff while preserving genuine completion and agent-exit handling. We really appreciate the careful, focused contribution.

PR #859, opened 2026-08-11
feat(opencode): show context-window usage in HUD

### LINKED ISSUE #830: OpenCode: show context-window usage in HUD/dashboard
Image

## Summary

Clawd currently shows context usage for supported Claude/Codex sessions, but no context percentage is shown for OpenCode sessions.

OpenCode itself already exposes this information in its session Context view. In my current session, OpenCode shows:

- Context limit: 200,000
- Total tokens: 111,600
- Usage: 56%
- Input tokens: 91,003
- Output tokens: 20,597

Clawd's HUD currently shows no context percentage for this OpenCode session.

## Expected behavior

Show OpenCode context-window usage in the same HUD/dashboard location used for agents that already support context usage.

At minimum:

- Context usage percentage

If available through the OpenCode integration/API, the Dashboard could additionally expose:

- Context limit
- Current/total token count

## Related

#230 requested context/token information in the HUD more generally.

This issue specifically tracks **OpenCode parity** for that capability.

## Environment

- Windows 11
- OpenCode Desktop
- Clawd on Desk v0.14.0
- OpenCode plugin enabled

PeterShanxin:
## Summary

OpenCode sessions now report context-window usage through Clawd's existing context metadata pipeline, providing parity with supported Claude/Codex sessions. The plugin forwards a metadata-only context update when an assistant step completes; Clawd stores it as session context metadata and the existing HUD / Dashboard render the percentage and used/limit detail.

## Source of truth

Real OpenCode SDK/runtime behavior was verified against the v1.1.25 and v1.18.18 SDK contracts (official `packages/opencode/src/plugin` and `packages/sdk/js` sources) and a live OpenCode Desktop 1.18.18 instance (Windows ARM64):

- `message.updated` events carry the session message on `event.properties.info`
- assistant-message token fields: `input`, `output`, `reasoning`, `cache.read`, `cache.write`
- there is **no** plugin-event `tokens.total` (`tokens.total` is an internal SessionV1 aggregate, not a message-token field)
- the default SDK client `provider.list()` returns the fields-style envelope `{ data: { all, default, connected }, request, response }`; a caller using `responseStyle: "data"` receives the payload directly

## Context usage semantics

The reported "used" figure mirrors OpenCode's own Context view exactly:

```
input + output + reasoning + cache.read + cache.write
```

The context limit is resolved from the active provider/model metadata:

- `providerID` + `modelID` from the assistant message
- `provider.list()` result, SDK envelope unwrapped (`data` object) then `all`/`providers` array inspected → `providers.find(provider.id === providerID).models[modelID].limit.context`
- unknown provider/model limits fail closed to `limit: null` (the `provider.options.limit` heuristic was removed — the upstream contract does not define it as a context window)
- per-client-instance 60s TTL cache; only successful positive limits are cached, so null/thrown lookups can recover on the next sample

## Delivery semantics (maintainer review round)

- **SDK envelope**: the fields-style `{ data: { all: [...] }, request, response }` result is unwrapped before inspecting `all`; fixtures cover both the SDK envelope and the `responseStyle:"data"` payload shape.
- **Version-safe session ID**: `message.updated` resolves `properties.sessionID → properties.info.sessionID → event.sessionID` and never uses `info.id` (the message ID) for a message event; lifecycle events keep their existing `info.id` fallback.
- **Async ordering**: a per-session sequence is assigned synchronously when the event arrives, fenced by a per-session generation; only the latest sequence of the current generation may publish, so stale `provider.list()` completions are discarded (token values are never used as an ordering test).
- **Delivery-aware dedup**: only a successfully delivered positive-limit sample becomes the dedup baseline; same `used` with a different model/limit re-posts, a null limit retries the lookup, and a failed delivery retries on replay.
- **Instance/generation scoping**: the limit cache is keyed by client instance and context state by instance token with a session generation fence; context state is cleared on session deletion / `server.instance.disposed`, so a lookup started by an old session generation cannot publish into a reopened session.
- **Metadata coalescing**: queued metadata snapshots merge per-field (`coalesced=metadata-fields`) so a session-title update and a context-usage update can no longer replace one another.
- **Scope**: context usage is explicitly OpenCode-only in the shared family core (`AGENT_ID === "opencode"` early-return) until a real MiMo compatibility fixture exists.

## Integration

The existing Clawd pipeline is reused end-to-end — no OpenCode-specific UI was added:

```
OpenCode message.updated
→ metadata-only /state POST (serialized per-session delivery, same channel as session-title metadata)
→ contextUsage metadata
→ session state metadata
→ session snapshot
→ existing HUD chip (percentage) + tooltip/dashboard (used / limit / percent)
```

## Lifecycle safety

Metadata-only context updates do not alter lifecycle state, `updatedAt`, recent-hook-event rings, completion badges, or stale cleanup. Unknown sessions are never created by the metadata POST (server-side designed drop), so no ghost sessions appear. Context usage rides the per-session serialized state channel introduced by the merged #855 (#841-style scaffolding, `metadata_only` short-circuits on the route).

## Tests

Coverage added/updated includes:

- real `properties.info` event shape
- exact component-sum token formula (incl. reasoning and cache read/write)
- exact providerID/modelID limit lookup
- SDK `{ data: { all: [...] }, request, response }` envelope unwrap
- v1.1.25 (`info.sessionID`) and current (`properties.sessionID`) `message.updated` session-ID shapes
- stale-lookup discard (sequence + generation fence)
- delivery-aware dedup: model change with same `used`, null-limit recovery, failed-delivery replay
- per-instance cache isolation and session-disposal/generation cleanup
- metadata field-merge coalescing (title + context coexist)
- missing/invalid limit behavior (fails closed to `limit: null`)
- metadata-only lifecycle isolation
- `source: "opencode"` normalization and `opencode-statusline` origin labeling (metadata and state paths)
- snapshot propagation
- existing-agent isolation (Claude/Codex behavior untouched)

Focused opencode suites (context usage, ordering/disposal, plugin session, route, snapshot, state): **518 passed / 0 failed**. Full suite on the final head: **7,952 passed / 0 failed / 35 skipped / 3 unrelated Windows PowerShell wrapper failures** (pre-existing, unrelated to this PR). Windows ARM64 package build: **PASS**. Native package audit: **0 errors**.

## Live verification

Windows 11 ARM64, OpenCode Desktop 1.18.18 with the packaged #830 plugin path in `plugin` config:

- completely fresh Desktop start with Clawd running: the packaged plugin initialized and reported during a real turn
- OpenCode Context view: Context Limit **1,000,000** / Total Tokens **44,745** / **4%** — Clawd's HUD showed the matching 4% context chip
- multiple OpenCode sessions show their real session titles together with their context percentages; title + context coexist correctly
- plugin log shows numeric `CTX used=… limit=…` values (no `limit=?`)
- controlled runtime probe: used = 24,272, limit = 1,000,000, correct session identity

## Scope

- no streaming token estimation
- no generic HUD redesign
- no #850 stale-timeout changes (merged via main)
- no duplicated #829 title implementation — title support is preserved as merged on main

Fixes #830


rullerzhou-afk (changes requested):
Thanks for the careful implementation and the detailed validation notes. I reviewed the exact current head, `957e1edaceb452f8d51de613e3ea1a347239875d`, against the official OpenCode v1.1.25/v1.18.18 contracts and the merge result with current main. I also ran deterministic probes with controlled promises rather than timing sleeps.

The token formula and the downstream `metadata_only` pipeline look correct. In particular, valid metadata does not create ghost sessions or alter lifecycle `updatedAt`/recent-event state, and the existing HUD/Dashboard consumers can render it without OpenCode-specific UI changes.

I cannot approve this head yet because the following issues remain.

### 1. The standard `provider.list()` SDK envelope is not unwrapped

The plugin runtime constructs a normal `createOpencodeClient()` without selecting `responseStyle: "data"`. In both v1.1.25 and v1.18.18, the default fields-style result is:

```js
{
  data: {
    all: [...],
    default: {...},
    connected: [...]
  },
  request,
  response
}
```

The current code checks `result.all`, `result.providers`, an array-valued `result.data`, or a raw array. It does not handle `result.data.all`.

A deterministic probe against the current resolver produced:

```text
SDK 1.1.25: resolveContextLimit -> null  (expected 1000000)
SDK 1.18.18: resolveContextLimit -> null  (expected 1000000)
```

The existing fake client models the HTTP response body rather than the SDK client result, so it does not catch this. Please add the real `{ data: { all: [...] }, request, response }` shape and unwrap the SDK layer before inspecting `all`.

Official references:

- [v1.1.25 plugin client construction](https://github.com/anomalyco/opencode/blob/v1.1.25/packages/opencode/src/plugin/index.ts#L24-L28)
- [v1.1.25 SDK fields-style return](https://github.com/anomalyco/opencode/blob/v1.1.25/packages/sdk/js/src/gen/client/client.gen.ts#L94-L145)
- [v1.18.18 plugin client construction](https://github.com/anomalyco/opencode/blob/v1.18.18/packages/opencode/src/plugin/index.ts#L141-L149)

### 2. `message.updated` needs a version-safe session-ID path

The current v1.18.18 internal runtime does provide `properties.sessionID`, so the existing fixture is valid for that runtime.

However, the PR explicitly cites v1.1.25. In v1.1.25, `message.updated.properties` contains only `info`, while the message's session ID is `info.sessionID`. The current fallback reads `info.id`, which is the message ID:

```text
getEventSessionId(no properties.sessionID): msg_01ABC
actual session ID in info.sessionID:        ses_real
```

That produces `opencode:msg_*`; the metadata update is then safely dropped because no such session exists.

Please resolve message events as:

```text
properties.sessionID -> properties.info.sessionID
```

and never use `info.id` as the session ID for a message event. Please add fixtures for both the v1.1.25 and current shapes.

### 3. Asynchronous limit lookups can reverse session order

For one session, I sent `used=100` followed by `used=200`, then completed the second `provider.list()` first. The actual POST order was:

```text
200 -> 100
```

The existing per-session FIFO cannot prevent this because each update enters that FIFO only after its independent lookup resolves. The dedup map already records `200`, so replaying `200` is skipped and the display can remain stuck at `100`.

Please assign a per-session generation/sequence when the event arrives and discard stale lookup completions. Token values must not be used as an ordering test because compaction can legitimately reduce usage.

### 4. Used-only dedup cannot recover or follow model changes

These cases currently fail:

- Same `used`, different model/limit: the new percentage is never posted.
- A lookup first returns `limit=null`, then becomes available after expiry: the unchanged sample never retries.
- `/state` delivery fails, then the same event is replayed: it remains deduplicated despite never being delivered.

Dedup should include the model/limit resolution state, and only a successfully delivered current-generation sample should become the stable dedup baseline. Transient lookup failures should not be cached as a normal `null` result for the full positive TTL.

### 5. Cache and lifecycle state need instance/generation scoping

The model-limit cache is shared by all directory instances but keyed only by `providerID::modelID`. Two projects can use the same IDs with different project-level model limits; the second instance currently reuses the first instance's limit without consulting its own client.

The context dedup map is also not cleared by session cleanup, and a lookup started by an old session generation can complete after that session ID has been reopened.

Please scope the cache to the client/directory instance and clear or generation-fence context state on session deletion/disposal.

### 6. Current main introduces a metadata coalescing collision

In the merge result with current main, queued metadata snapshots are coalesced solely by `metadata_only === true`.

With one POST in flight:

1. A session-title metadata update is queued.
2. A context-usage metadata update is queued.
3. The context update replaces the title update.

The deterministic result delivered only:

```text
session_title=undefined
context_usage={"used":3800,"limit":128000,"source":"opencode"}
```

The one-shot real title can therefore be lost permanently. Please coalesce by metadata payload kind, or merge the fields so title and context updates cannot replace one another.

### Live-verification evidence

There is also an evidence mismatch I would like to understand before relying on the Windows ARM64 result:

- The exact current head contains no debug-log line that emits `limit=`.
- The standard v1.1.25/v1.18.18 SDK client shape above resolves to `null` in the current code.

Could you provide the following sanitized evidence from the plugin process?

1. Exact `opencode --version` and the installed SDK/plugin package versions.
2. `Object.keys(await ctx.client.provider.list())`.
3. `Object.keys(result.data ?? {})`.
4. `Object.keys(event.properties)` for one real `message.updated`.
5. The source line that produced the reported numeric `limit=` log, or the diff between the code actually tested and this PR head.
6. A screenshot showing the percentage chip, rather than only the raw token-count fallback.

No prompts, provider credentials, or full session IDs are needed.

Please also either provide the upstream contract for `provider.options.limit` as a context-window fallback or remove that heuristic. Finally, because this shared core also allows MiMo Code metadata, please add an explicit MiMo compatibility test or narrow the claimed scope.

Once these paths and the corresponding regression tests are updated, I'll re-review the new exact head.

PeterShanxin:
Thanks for the detailed review. All six findings are addressed on the new head `399b54e0`, and the real packaged-plugin manual gate now passes with a completely fresh OpenCode Desktop start.

1. **SDK envelope** — `resolveContextLimit` now unwraps the fields-style SDK result (`{ data, request, response }`) via `unwrapProviderListResult` before inspecting `all`; the `responseStyle:"data"` payload path is covered too, with deterministic fixtures for both shapes (previously `null` per your probe — now resolves as expected).
2. **Version-safe session ID** — `getEventSessionId` resolves `message.updated` as `properties.sessionID → properties.info.sessionID → event.sessionID` and never `info.id`; lifecycle events keep their existing `info.id` fallback. Fixtures cover both the v1.1.25 and current shapes.
3. **Async ordering** — a per-session sequence is assigned synchronously at event arrival (`++state.sequence`) with a per-session generation fence; stale `provider.list()` completions are discarded (`CTX discard … reason=stale`) and token values are never used as an ordering test.
4. **Delivery-aware dedup** — the dedup baseline is a successfully delivered positive-limit sample only; same `used` with a different model/limit re-posts, a null limit retries the lookup, and a failed `/state` delivery retries on replay.
5. **Instance/generation scoping** — the limit cache is keyed per client instance and context state per instance token, with session-generation fencing; context state is cleared on session deletion / `server.instance.disposed`, so an old-generation lookup cannot publish into a reopened session.
6. **Metadata coalescing** — queued metadata snapshots now merge per-field (`coalesced=metadata-fields`) instead of last-wins, so a title update and a context update cannot replace one another.

Also: the `provider.options.limit` heuristic is removed — unknown limits fail closed to `limit: null`, since the upstream contract does not define it as a context window. The shared core is explicitly OpenCode-only for context usage (`AGENT_ID !== "opencode"` early-return) until a real MiMo compatibility fixture exists.

**Live verification (Windows 11 ARM64, OpenCode Desktop 1.18.18, packaged plugin in `plugin` config):**
- Completely fresh Desktop start with Clawd already running: the packaged plugin initialized and reported during a real turn.
- OpenCode Context view: Context Limit **1,000,000** / Total Tokens **44,745** / **4%** — Clawd's HUD showed the matching 4% context chip.
- Real session titles are preserved alongside the context percentages (title + context coexist); multiple sessions show their real titles and percentages.
- Plugin log shows numeric `CTX used=… limit=…` lines (no `limit=?`).
- Controlled runtime probe: used = 24,272, limit = 1,000,000, correct session identity.

Automated: focused opencode suites **518 passed / 0 failed**; full suite **7,952 passed / 0 failed / 35 skipped** plus 3 pre-existing unrelated Windows PowerShell wrapper failures; Windows ARM64 package build **PASS**; native package audit **0 errors**.

PR #859 body has been updated with the final results. Happy to re-review on your side.

The screenshot below is additional UI evidence that multiple real OpenCode session titles coexist with their context percentage chips. The exact 44,745 / 1,000,000 / 4% numeric match was observed in the Context panel during the same verification, but is not shown in this cropped screenshot.
image

rullerzhou-afk:
Thank you @PeterShanxin for contributing the OpenCode context-window usage feature and for iterating on #830. We took over a small maintainer follow-up directly on this PR in e48a759f; the feature scope and core implementation remain yours.

The follow-up separates server recognition from metadata acceptance, fails closed on malformed message identity/role, fences stale provider-limit cache writes, and adds deterministic race/protocol coverage plus sanitized validation logging.

Validation: targeted suites 588/588; real OpenCode 1.18.11 fresh-session, resumed-session, and disable/re-enable smoke all passed; the live Dashboard showed 16,267 / 128,000 (13%).

Thanks again for laying the groundwork and bringing this feature to Clawd!

rullerzhou-afk (approved):
Reviewed the exact current head e48a759f34d0c76cb0ecc9874dad4755c788e40b. The six issues from the earlier changes-requested review were resolved in subsequent contributor revisions, and the maintainer follow-up closes the remaining metadata-acceptance, malformed-identity, and stale-cache-write gaps.

Targeted suites pass 588/588, the required audit passes, and real OpenCode 1.18.11 fresh/resume/disable-re-enable smoke shows the live Dashboard context value 16,267 / 128,000 (13%).

Approved. Thank you again @PeterShanxin!

PR #865, opened 2026-08-12
fix(permission): stop prompting in Claude Code bypassPermissions sessions

Praesentia-YKM:
## Problem

Claude Code fires the `PermissionRequest` hook even in sessions started with
`--dangerously-skip-permissions`. Clawd's handler pops an approval bubble for those
requests, so a user who explicitly asked never to be asked still has to click through
every tool call.

## What this does

The hook payload already carries the session's mode in `permission_mode`. When that value
is `bypassPermissions` and the request is not a decision interaction, Clawd drops the
connection and lets Claude Code resolve the request natively, exactly like the existing
bubble-disabled paths in the same handler.

Clawd never answers `allow` or `deny` on the user's behalf here. It does not decide
anything, and it does not compensate for Claude Code's behavior. It reads a value it is
already being sent, and steps out of the way.

## Placement

The new branch sits after the two rules that must keep winning:

1. the headless auto-deny rule
2. the `PASSTHROUGH_TOOLS` auto-allow rule

Both still run first. Two regression tests pin that order, modeled on the ones that
already pin it for the subagent sub-gate (#451): moving the new branch above either rule
fails the suite.

`ExitPlanMode` and `AskUserQuestion` stay exempt through `isDecisionInteraction`. They are
UX flows rather than approvals, and Claude Code sends them regardless of permission mode.

## Two effects worth calling out

Both follow from the branch sitting above the bubble gates, and both are deliberate.

**Remote approval does not run for these requests.** No code in the Telegram / Lark path
changed, but a `bypassPermissions` request now returns before `tryRemoteOnlyApproval`.
That is the intent: the session's owner opted out of being asked, so keeping a remote
channel alive would reintroduce the same prompt on another device. If you would rather
these requests still reach remote approvers, that is a small move and I am happy to make
it.

**The check is not agent-scoped.** This block is shared with codebuddy, and `res.destroy()`
is already what it falls back to for every agent that reaches it, so an agent gate would
add a condition that cannot change today's outcome. Codex never reaches this branch: it
returns earlier in the handler under its own no-decision contract.

## Why only `bypassPermissions`

Claude Code's bundle enumerates six modes: `acceptEdits`, `auto`, `bypassPermissions`,
`default`, `dontAsk`, `plan`. Only `bypassPermissions` is described as bypassing all
permission checks. The neighbour that looks similar is not equivalent: `dontAsk` is "don't
prompt for permissions, deny if not pre-approved", so a bubble there is still the user's
only chance to approve something. Widening the check would change behavior the user never
asked to change, so this PR leaves the other five modes alone.

## Why `accepted()` rather than a `dropped-*` counter

`evaluateConnectionTest` in `doctor-hook-activity.js` reports `http-dropped` at warning
level as soon as any event carries a `dropped-*` outcome. Recording these as dropped would
make Doctor warn about a healthy bypass session. The two adjacent bubble-gate branches
record `accepted()` for the same reason.

## Evidence for `permission_mode`

`AGENTS.md` asks for at least one real Claude Code run behind any change touching the
`/permission` payload, so I captured one rather than writing a `curl` body by hand.

The setup was a scratch directory containing nothing but a `PermissionRequest` command
hook, pointed at a script that appends stdin to a file and prints `{}` — it records, it
never decides. A Claude Code 2.1.228 session started there with
`--dangerously-skip-permissions`, then one `Bash` call. The hook fired at all, which is
the first fact this PR rests on, and the body it wrote carries the second:

```json
{
  "session_id": "3d182efb-8a48-4bb2-973b-4546d1eaa90c",
  "transcript_path": "<local path>",
  "cwd": "<local path>",
  "prompt_id": "85d97bf5-7195-42a7-bad6-c227c1f1aea2",
  "permission_mode": "bypassPermissions",
  "effort": { "level": "xhigh" },
  "hook_event_name": "PermissionRequest",
  "tool_name": "Bash",
  "tool_input": { "command": "echo hi", "description": "Print hi" }
}
```

That body is now a test, with the two local paths anonymized the way the #451 subagent
fixture anonymizes its own. It runs through `handlePermissionPost` and asserts the
connection is destroyed with no pending permission, no bubble, and no remote approval
started. Worth noting: there is no `agent_id`, because this is a main-thread request
rather than the subagent request #451 captured.

The capture and the handler's behaviour on it, side by side:

live Claude Code payload and the resulting handler behaviour

## Scope

Three files, 147 added lines, one line changed. No new config option, no UI or translation
string, no new dependency.

`docs/project/agent-runtime-architecture.md` did not list `permission_mode` in the
`POST /permission` receive contract, so this PR adds it there along with a bullet for the
new gate under Permission Bubble.

Tests: `test/server-route-permission.test.js` passes 94/94. The full-suite failure count is
unchanged from the base commit (44 pre-existing failures on my machine, all in unrelated
telegram and native-pointer suites).

### DIFF
--- a/docs/project/agent-runtime-architecture.md
+++ b/docs/project/agent-runtime-architecture.md
@@ -271,13 +271,14 @@ CodeBuddy 的 PermissionRequest HTTP 所有权只认严格的本机 managed URL�
 - WorkBuddy 不进入 `/permission`：权限请求只以 Notification 驱动提醒，Allow / Deny 决策留在 WorkBuddy 原生 GUI
 - QwenWork 不进入 `/permission`：`PermissionRequest` / `PermissionDenied` 只被观察并映射成 `working`，hook stdout 恒为 `{}`，Clawd 不产生 allow/deny，也不在 permission automation eligibility 名单内
 - Codex 的 PermissionRequest 是 official command hook；hook 脚本挂起等待 `/permission`，再把 sanitized allow/deny JSON 写到 stdout
-- `POST /permission` 接收 `{ tool_name, tool_input, session_id, permission_suggestions }`；Codex 额外带 `turn_id`、`tool_input_description`、`tool_input_fingerprint`
+- `POST /permission` 接收 `{ tool_name, tool_input, session_id, permission_suggestions }`；Claude Code 与 Codex 还带 `permission_mode`（会话权限模式）；Codex 额外带 `turn_id`、`tool_input_description`、`tool_input_fingerprint`
 - 每个权限请求都会创建独立 `BrowserWindow`，多个 bubble 从右下向上堆叠
 - bubble 会通过 IPC `bubble-height` 回报真实高度，主进程据此重排
 - 支持 Allow / Deny / suggestion 决策，以及 `addRules` / `setMode` suggestion 类型
 - `permission-automation-policy.js` 的 off / auto-tools / unattended 与 `session-automation-coordinator.js` 的 per-session grant 会在 bubble 渲染前产生真实决定。auto-tools 对 Claude/Qwen 的未知 built-in（除有效 namespaced MCP）fail closed，但其他已知 adapter 对非空工具名不都使用逐工具 allowlist；unattended 在识别已知 decision tools 后仍有意对可作 Allow/Deny 的未知请求保留“handle every request”行为。新增 agent/tool/interaction 必须同时审查 policy 与 tests，不能笼统假设 unknown 一律 defer
 - Telegram 与飞书 / Lark 是和本地 bubble 并行的远程决策通道；关闭本地 bubble 不等于关闭远程审批。远程 client 超时、断连、未配置或启动失败不得产生决定或 deny：本地 bubble 存在时请求继续 pending；仅在 remote-only 且所有可用 client 都无决定时，整体请求才 no-decision 并让 agent 回原生 UI 重问
 - DND 只负责“不弹 bubble”，不替用户决定权限：opencode 与 MiMo Code 分支 silent drop，让 TUI 内置权限提示接管；Claude Code 分支 `res.destroy()`，让 CC 回到内置聊天/终端确认；Codex 分支返回 no-decision `{}`，让 Codex 原生审批接管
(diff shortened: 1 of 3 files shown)


rullerzhou-afk:
Thank you for investigating this and for including a real Claude Code 2.1.228 payload. The reported behavior was real, and the tests accurately capture that older payload.

Before merging, we repeated the validation with Claude Code 2.1.229, the PR build of Clawd, and active local/Telegram/Lark approval channels. The results changed the conclusion:

- An ordinary interactive `--dangerously-skip-permissions` Write completed without Claude emitting `PermissionRequest`, so this PR's new branch was never reached.
- `claude -p` likewise emitted no `PermissionRequest` in either bypass or default mode; Claude handled the allow/deny internally.
- Most importantly, we added an explicit project rule requiring confirmation for `Bash(git status:*)`, then ran that command in an interactive bypass session. Claude did emit `PermissionRequest` with `permission_mode: "bypassPermissions"`. This PR's early return matched it, suppressing the Clawd bubble and both Telegram/Lark approval cards. Claude's native prompt still appeared, and the command succeeded after choosing Allow once.

That last case is the blocker. The proposed predicate cannot distinguish the noisy ordinary request seen in 2.1.228 from a legitimate forced-ask request in 2.1.229. The payload provides no stable reason or version signal that would let Clawd safely separate them. Keeping this version-independent gate would therefore disable Clawd's local and remote approval paths for real requests that still require a decision.

There are secondary issues we could fix mechanically (scope the predicate to `claude-code` so CodeBuddy cannot inherit it, correct the documentation claim that bypass has no pending approvals, and expand the adapter/test matrix), but none of those resolves the protocol ambiguity above.

Because ordinary bypass requests no longer reproduce the problem on the current Claude Code version, and the remaining bypass requests can be legitimate, I don't think another incremental revision of this approach would be productive. My recommendation is to close this PR without merging. If ordinary bypass requests become reproducible again on a current Claude release, or Claude exposes a reliable signal distinguishing forced prompts from the old noisy request, this would be worth revisiting in a new PR.

Thanks again for the careful report, live capture, tests, and documentation. They were genuinely useful in isolating the upstream behavior change.

PR #866, opened 2026-08-12
fix: make Clawd accessory drag hitboxes size-aware

CheeseAgent:
## Summary

- stack the selected Clawd accessory above the built-in safety helmet during the 3+ session building animation
- derive the draggable hitbox from the currently selected accessory's actual aspect ratio, scale, offset, and per-animation motion padding
- keep the accessory contribution at exactly zero when no accessory is worn (the building pose's own box is widened separately, so its built-in helmet becomes draggable with or without a hat)
- refresh the hit window immediately after manual or seasonal accessory changes
- preserve the typing screen and built-in safety helmet drag regions

## Root cause

Clawd previously relied on fixed file-level hitboxes. Expanding those boxes for the tallest possible accessory made transparent regions draggable even when no accessory, or a shorter accessory, was selected. The building animation also placed the selected accessory at the normal head anchor, overlapping the built-in safety helmet.

## User impact

The 1-, 2-, and 3+-session working animations now remain draggable through their visible content. An equipped accessory contributes only its own calculated envelope, while selecting no accessory adds no transparent hat region.

## Validation

- `npm test`
- `npm run verify:electron`
- targeted accessory capability, hitbox geometry, theme schema, settings routing, holiday accessory, and pet-window runtime tests

---

## Maintainer follow-up

Picked this up rather than sending another review round — thanks for the Electron audit, it is what made the rest of this findable. Four commits on top (`d6299dc6..9bab1260`), merged with current `main`.

**Facing.** Geometry predicted the accessory's mirroring from mini edge XOR `flipAssets`, with the edge re-derived via `screen.getDisplayMatching`. That misses free roam (mirrors on walk heading, mini mode off) and the menu walk-in (mirrors seconds before the mini flag turns on). The renderer now reports the composed facing and geometry consumes it; the heuristic is gone. The rule moved to `src/pet-accessory-mirror.js` so both sides cannot drift.

**Envelopes.** The three `cloudling` entries were measured under conditions that do not hold:
- `cloudling-idle` / `cloudling-mini-idle` are pointer-driven, and a headless audit window never engages the pointer — the sampled figures were the breath-only subspace. Both are now analytic bounds from the SVGs' own constants. `mini-idle` was short by ~6.2 viewBox units per side, which is the drawn-but-not-draggable case this PR is about.
- `cloudling-mini-crabwalk`'s `frameFor()` is a pure function of the clock, so it is swept deterministically through a seek hook at a step coprime with both its periods; a period-dividing step steps clean over the contact pulse's peak.

**Also:** `syncHitWin` now reports `{applied, deferred}` (a drag in flight is not a failure); the payload fallback uses the non-committing builder; the envelope table gained a coverage guard; `validate-theme` stops printing a green accessory line for a theme whose capability is off; Linux CI runs the suite under xvfb so the audit stops skipping itself; `hitBoxPadding` is documented as a real contract.

### Verification

`npm test` on the merged tree: 8309 tests, 30 failures — identical by name to `origin/main`'s pre-existing DSH/installation-detector family, so zero regressions. The Electron audit passes on macOS, where it was deterministically red before. Real-machine spot check on macOS with Cloudling + wizard hat: the hat's outer edge is draggable in mini mode, and the hat stays grabbable during a left-heading roam.

### Known, not blocking

- `cloudling-typing.svg` has ~0.004 viewBox units of margin and is still wall-clock sampled; one real-machine sample showed a 1.2px top overhang, at the limit of that probe's own precision. Worth deriving an analytic bound for it.
- The menu walk-in reports the pre-mini facing for one IPC hop. Only wrong for a `flipAssets: true` theme at the left edge, and the only such built-in theme has no accessories.
- `validate-theme` derives the capability from the raw theme while the runtime derives it after the variant patch; no built-in theme has variants.
- No Windows or Linux hardware was exercised.


rullerzhou-afk (dismissed):
Thanks for the focused fix. I reviewed exact head `1adbf907` and am requesting changes because the new native hit-window path still has two blocking correctness/safety gaps and several runtime consistency gaps.

1. **Please contain accessory expansion before it reaches the native input window.** A schema-valid third-party theme can currently combine the existing broad `staticFrame` bounds with the new `hitBoxPadding` bounds and create a transparent input window far beyond any rendered pixels. One valid 100x100 fixture projected to roughly 2158x1199 px, which can intercept desktop clicks. This also affects older valid themes whose large `staticFrame` previously controlled rendering only. Please clamp the *accessory contribution* to the physical render-visible bounds (while preserving the existing base-hitbox semantics), tighten the schema as defense in depth, and add a maximum-valid external-theme integration test.

2. **`staticFrame` is not a valid oracle for animated `followTarget` content.** The renderer follows `getCTM()` every frame, while `resolveAccessoryAwareHitBox()` only unions the static frame plus authored padding. Chromium sampling found real misses: Cloudling idle extends about 4.8-4.9 theme units beyond the computed box on three sides; Clawd happy misses 12 units at the top; sleeping, yawn, debugger, and sweeping also escape their boxes; and headphones still miss about 0.77 units on each horizontal side with padding 1.5. Please audit every built-in dynamic descriptor and provide a measured motion envelope for each one, rather than adjusting only headphones.

3. **Apply the same mini mirroring semantics as the renderer.** Left-edge mini mirrors the whole facing stage, and `miniFlipAssets` may add a second flip, but main-process geometry has neither the edge nor effective-flip context. The asymmetric Clawd mini frame can still miss 3-8 px at large 5K/6K sizes. Please project the effective `edgeLeft XOR miniFlipAssets` transform before mini padding and seam clipping, and test both edges at a size large enough that the fixed 25 px padding cannot hide the error.

4. **Use one authoritative accessory payload.** Hit geometry currently re-resolves settings and the current date, independently from the payload last delivered to the renderer. Around midnight or after a clock/time-zone change, the visible accessory and hitbox can disagree. Please keep one main-process `currentAccessoryPayload` shared by renderer config, Settings, holiday refresh, theme reload, and geometry.

5. **Complete the accessory-change side effects and failure handling.** After a manual or holiday change, `syncHitWin()` does not reposition permission/update surfaces that depend on `getHitRectScreen()`. The holiday runtime also commits `lastDisplayKey` before `onAccessoryChange` succeeds, so a failed hitbox sync is permanently deduplicated. A single apply path should update the canonical payload, deliver it, sync the hit window, reposition floating surfaces, and retry a failed geometry sync without resending an unchanged renderer payload.

The current tests pass, but they mostly duplicate the static union formula or check that target IDs exist. Please add an independent Chromium/Electron test that loads the real SVGs, samples animated `getCTM()` values, and asserts that all seven accessory corner points remain inside the final screen hit rectangle, including mini padding, mirroring, and clipping. I would also use outward integer rounding (`floor` left/top and `ceil` right/bottom) and document/validate the new public `hitBoxPadding` field in the theme guide and validation CLI.

CheeseAgent:
Addressed the requested review items. Summary of the follow-up:

1. **Native hit-window containment / third-party safety**
   - Accessory-only hitbox expansion is now clamped to the effective render-visible viewBox before unioning with the existing base hitbox.
   - The base hitbox itself is left unchanged.
   - Tightened `hitBoxPadding` schema bounds and added a maximum-valid external-theme regression test so schema-valid metadata cannot create a giant transparent native input window.

2. **Measured animated motion envelopes**
   - Added an independent Electron/Chromium audit that loads the real built-in SVGs and samples the animated `getScreenCTM()` values.
   - Audited all dynamic accessory descriptors for Clawd and Cloudling, not just headphones, and added per-file measured envelopes where the real animation escapes the static projection.
   - The audit caught additional cases beyond the original report (including several mini/scripted states), which are now covered as well.

3. **Mini mirroring / clipping**
   - Main-process geometry now uses the renderer-equivalent `edgeLeft XOR miniFlipAssets` horizontal transform.
   - Mirroring is applied before mini padding / seam clipping.
   - The Electron audit projects into a 6000x6000 screen geometry and checks both mini edges, including the fixed 25px mini padding and seam clipping, so that padding cannot hide an asymmetric-frame error.

4. **Single authoritative accessory payload**
   - Added a main-process canonical accessory payload shared by renderer delivery, Settings changes, holiday refresh, theme reload, and hit geometry.
   - Geometry no longer independently re-resolves settings/date state around midnight or timezone changes.

5. **Apply side effects / retry behavior**
   - Accessory changes now update the canonical payload, deliver it, sync native hit geometry, and reposition hitbox-dependent floating surfaces.
   - `syncHitWin()` now provides a success/failure acknowledgement.
   - Holiday refresh does not permanently dedupe a failed geometry apply; it can retry geometry without resending an unchanged renderer payload.

Also switched native hit-window materialization to outward integer rounding and documented/validated the public `hitBoxPadding` field in the theme guide and validation CLI.

### Validation

I validated the final code state against a prospective merge with the current upstream `main`:

- upstream merge: clean
- `npm run verify:electron`: pass
- review-targeted suite: **188/188 pass, 0 fail, 0 skip**
- the suite includes the real-SVG Electron CTM audit, seven accessories, scripted Cloudling motion, 6000px mini geometry, both edge orientations, mirroring, padding, seam clipping, canonical-payload behavior, external-theme containment, and failed-geometry retry behavior

The current PR head is `be81faec`; the validated code head was `f1955646`. The commits after `f1955646` only removed the temporary CI workflow/result artifacts used for this validation, so the product/test code is unchanged.

I also ran the broader prospective-merge test suite. The remaining failures are in untouched DSH installation/PATH simulation tests; none of the files involved in those failures are changed by this PR.

rullerzhou-afk:
Thanks for sticking with this one — and for building the Electron audit. I want to be clear that it is the reason the rest of this was findable at all: an audit that loads the real SVGs and samples `getScreenCTM()` is exactly the right instrument, and I kept it wholesale. Every finding below came out of running *your* harness.

I've pushed four commits to this branch rather than sending another review round. That felt like the fairer trade after the wait, and the changes are more surgical than a description of them suggests: 24 files, +624/−89 against a merge with current `main`, of which the single largest deletion is 37 lines of the mini-edge guesswork in `pet-geometry-main.js`. Eight of the files you touched I did not modify at all, and `pet-accessory-hitbox.js` is ~93% yours.

What I changed, and why:

**Facing.** `mirrorX = miniMode && ((edge === "left") !== miniFlipAssets)` is missing a state-family term. The renderer's `.mini-left` stage and its asset-direction stage both wrap the accessory layer, and neither is gated on mini mode — free roam mirrors on walk heading alone, and the menu walk-in mirrors for the several seconds before the mini flag flips. Rather than reconstruct that predicate in main (which needs two pieces of state that do not currently exist), the renderer now reports the composed answer and geometry consumes it. `resolveMiniEdge` and its `getDisplayMatching` fallback are deleted.

**The envelope table.** This is the part worth reading if you take one thing from this. Three of the four `cloudling` entries were measured under conditions that cannot occur in production:

- `cloudling-idle` and `cloudling-mini-idle` are pointer-driven, and the audit's hidden window never engages the pointer. `mini-idle` in particular only ever had the probes applied to `cloudling-idle.svg`, so it recorded the breath-only subspace — `{0,0,0,0.336}` — while the real envelope is `6.205/4.238/6.205/2.281`. About 6 viewBox units of hat on each side, drawn but not draggable, whenever the cursor sat away from the pet. The audit and the table shared one blind spot, so it stayed green.
- `cloudling-mini-crabwalk`'s `frameFor()` is a pure function of the clock, so sampling it against the wall clock measures the machine. That is why your CI read `left=3.98, right=2.66` on a rectangle that is provably symmetric about its own pivot, and why the audit was deterministically red on macOS.

Both are now derived instead of sampled — analytic bounds from the SVGs' own constants for the pointer-driven pair, and a deterministic sweep through a seek hook for the walk. The audit also fails now if a pointer-driven visual measures no rotation, because "this animation barely moves" and "the probes never fired" look identical from outside.

I made the same class of mistake twice while doing this, for what it's worth: my first pass left the renderer→main report with no end-to-end coverage (deleting the call kept 138/138 green), and my sweep used a 20ms step that divides the 1.16s walk cycle exactly 58 times, so it revisited 58 phases forever and stepped over the peak it was meant to bound. Both were caught by review rounds, not by me. This is a genuinely hard thing to get right.

Verification: 8309 tests on the merged tree with 30 failures, identical by name to `main`'s pre-existing set. Your audit now passes on macOS. Real-machine check on macOS confirms the hat's outer edge is draggable in mini mode and stays grabbable during a left-heading walk.

I've listed the remaining known items in the PR description — none blocking, and `cloudling-typing.svg` is the one I would look at next if you're interested. Merging shortly. Thanks again for the work and the patience.

PR #884, opened 2026-08-15
refactor: animate settings groups with CSS Grid

YOIMIYA66:
## Summary

- Replace the shared Settings collapsible group's measured `max-height` animation with a CSS Grid `0fr` to `1fr` transition.
- Keep persistence, keyboard controls, ARIA state, `inert`, nested groups, dropdown overflow, scroll anchoring, and reduced-motion behavior intact.
- Remove the remaining `animateExpansion: false` call-site exceptions and use a local fade-in only for asynchronously inserted content.

## Problem context

The previous implementation read `scrollHeight`, wrote a measured CSS height, and waited for animation frames while tracking `expanding`, `collapsing`, and `resizing` states. That created a visible click delay and made dynamic Settings rows vulnerable to clipping or a second layout jump. Some groups disabled animation to avoid those symptoms, which made the shared interaction inconsistent.

## What changed

- Added a stable body-inner wrapper so the outer body owns the Grid track while the inner element owns content overflow, padding, and borders.
- Replaced height measurement and RAF-driven expansion with intrinsic Grid sizing and direct state updates.
- Centralized transition interaction locking. Rapid reversals invalidate stale generations; `transitionend`, `transitioncancel`, and a watchdog release `inert` safely.
- Derive the watchdog duration from the CSS transition custom property and add only a small event-loss buffer, so CSS and JavaScript do not carry duplicate animation durations.
- Preserve the language-picker overflow escape for both open and mounted menu lifecycles, and keep reduced-motion transitions and animations disabled.
- Mark newly revealed quota content for a short local fade-in without reintroducing group-level height measurement.

## Behavior after this PR

Settings groups begin opening or closing immediately using intrinsic layout. Hidden content remains non-interactive during the transition and becomes interactive after the current transition settles; stale callbacks cannot unlock a newer state. Async rows naturally expand the open group, nested groups remain independent, and persisted state / ARIA / keyboard behavior remain unchanged. Reduced-motion users receive an immediate state change without motion.

No Settings store/controller, IPC, preference schema, or business behavior was changed.

## Validation

- `node --test test/settings-renderer-browser-env.test.js` — 222 passed, 0 failed.
- `npm test` — passed on Windows.
- `node --check src/settings-ui-core.js` — passed.
- `node --check test/settings-renderer-browser-env.test.js` — passed.
- `git diff --check` — passed (only the repository's existing LF/CRLF warnings were reported).

## Platform note

Automated validation ran on Windows. This draft intentionally remains open for Electron GUI smoke testing of sound, quota, taskbar flash, Agent lists, nested remote approval menus, narrow windows, text scaling, light/dark themes, and reduced motion before it is marked ready.

## Scope control

This PR covers the shared Settings collapsible component only. Electron's native tray menu and buttons in independent windows are unchanged. The standard-button migration is delivered separately in the companion draft PR.

Part of #857

### DIFF
--- a/src/settings-tab-general.js
+++ b/src/settings-tab-general.js
@@ -717,7 +717,6 @@
       desc: t("rowQuotaRingGroupDesc"),
       defaultCollapsed: true,
       className: "quota-ring-collapsible",
-      animateExpansion: false,
       children: [optionList],
     });
     if (window.settingsAPI && typeof window.settingsAPI.getQuotaSourceCount === "function") {
@@ -726,12 +725,11 @@
           if (Number(count) <= 1) return;
           const revealMergeRow = () => {
             mergeRow.style.display = "";
-          };
-          if (typeof group.mutateCollapsibleBody === "function") {
-            group.mutateCollapsibleBody(revealMergeRow);
-          } else {
-            revealMergeRow();
+            if (!group.classList.contains("collapsed")) {
+              mergeRow.classList.add("collapsible-content-enter");
+            }
           }
+          revealMergeRow();
         })
         .catch(() => {});
     }
@@ -940,7 +938,6 @@
       desc: t("rowFlashDesc"),
       defaultCollapsed: true,
       className: "flash-collapsible",
-      animateExpansion: false,
       children: [buildOptionList("flash-option-list", [
         helpers.buildSwitchRow({
           key: "flashTaskbarOnComplete",
--- a/src/settings-ui-core.js
+++ b/src/settings-ui-core.js
@@ -47,6 +47,10 @@
   const COLLAPSED_GROUPS_STORAGE_KEY = "clawd.settings.collapsedGroups.v1";
   const NAVIGATION_STORAGE_KEY = "clawd.settings.navigation.v1";
   const MAX_PERSISTED_SCROLL_TOP = 10_000_000;
+  // Keep the watchdog tied to the CSS transition rather than duplicating its
+  // duration here. The buffer only covers a transitionend event that is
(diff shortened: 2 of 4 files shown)


YOIMIYA66:
Closing this draft because its implementation baseline is now too far behind `main` to update safely in place.

Current status:

- The PR head remains at `225b68cd`, while current `main` is `a7283581` (188 commits ahead of the PR head).
- The branch has real conflicts in `src/settings-ui-core.js`, `src/settings.css`, and `test/settings-renderer-browser-env.test.js`.
- Current `main` added contracts that the old implementation does not preserve, including imperative `group.expand()` used by the Feishu API Explorer fallback flow.
- Current `main` also contains newer picker `menu-mounted` lifecycle handling, stable scrollbar gutter behavior, focus preservation, responsive layout fixes, and additional shared collapsible call sites/tests.

The CSS Grid approach itself is still valid. Rather than resolving the old conflicts and risking regressions or reimplementing fixes already present on `main`, the shared-collapsible phase will be rebuilt from the latest `main`, preserving the current `expand()`, async-content, picker overflow, focus, scroll-anchor, ARIA/inert, persistence, and reduced-motion contracts.

Scope clarification: this PR and Issue #857 cover shared `buildCollapsibleGroup` consumers. Doctor disclosures and Animation Overrides `<details>` rows remain separate implementations and should be considered separately if a broader shared motion language is desired.

A replacement PR will reference #857 and this draft once the refreshed implementation is ready. Closing #884 as obsolete/superseded by that planned main-based replacement, not because the Grid direction was rejected.

PR #886, opened 2026-08-15
feat(traecode): add TraeCode (Trae CN) hook integration

### LINKED ISSUE #633: 支持 codebuddy 和 qoder，能不能也支持一下 trae

xiaoshidefeng:
## Summary

Add support for **TraeCode** (ByteDance's Trae CN IDE), fulfilling the request in issue #633 to "also support trae". Once enabled, the Clawd desktop pet shows and tracks state during TraeCode sessions.

- New `traecode` agent definition mapping the 6 hook events TraeCode supports (SessionStart / UserPromptSubmit / PreToolUse / PostToolUse / Stop / Notification)
- New `hooks/traecode-hook.js` hook script wired into `~/.trae-cn/hooks.json` (Claude Code-compatible format)
- New `hooks/traecode-install.js` installer: idempotent, migrates legacy configs, validates structure, and on Windows emits the PowerShell `&` call-operator prefix
- Session title: Trae stores it server-side, so Clawd derives it from the first prompt line and keeps the **first** title per session (server-side first-wins)
- TraeCode desktop idle-timeout cleanup so closed conversations retire even though Trae has no `SessionEnd`
- Usage hint on the TraeCode agent card + docs (README, setup guide, known limitations, runtime architecture) and CI coverage

## Scope (first release)

This release covers **Trae CN** only (`~/.trae-cn/hooks.json`, process names `Trae CN.exe` / `Trae CN`). The international Trae build (`~/.trae/hooks.json`, `Trae.exe`) is deliberately not registered.

## Manual setup required in Trae settings (important)

TraeCode hooks **must be enabled manually in Trae settings** — there is no programmatic bypass:

1. Open Trae **Settings > Hooks**
2. In the Configuration section, select the **Global** type
3. In Configured Hooks, click Create / Enable and confirm in the security warning panel

The hook run mode must be set to **Run Automatically Locally** so Clawd can receive TraeCode session events. Official docs: [Automate with hooks - TRAE](https://docs.trae.cn/ide_automate-actions-with-hooks) (international mirror: [docs.trae.ai](https://docs.trae.ai/ide/automate-actions-with-hooks)).

## State-only

The hook's stdout is always `{}` on every event — Clawd registers no `/permission` endpoint and produces no Allow / Deny; every approval stays in Trae's native permission flow. There is no `SessionEnd` event.

## Review feedback addressed

- **PreToolUse stdout** now emits `{}` (was top-level `{"decision":"allow"}`, which violates Trae's `hookSpecificOutput.permissionDecision` shape) — offline-contract test updated
- **Session ids are namespaced** as `traecode:<raw-id>`; events with a missing/blank id are answered immediately and never POSTed (no phantom `"default"` session)
- **Scoped to Trae CN** and removed the international process-name claims
- **Removed the insecure disk title lock**; the server keeps the first successful title per session (state.js first-wins), and first-POST-fails/second-POST-succeeds is covered by the design
- **Trae desktop idle-timeout cleanup** added (`traecode-desktop-idle-timeout`) with local/remote/headless boundary tests, mirroring Codex Desktop / ZCode
- **Hook lifecycle guarded behind `require.main`** — importing the module no longer reads stdin, arms timers, writes stdout, or exits; added an import-safety test
- **Installer hardening**: fails closed on non-object root/hooks, converts owned flat entries to the documented nested shape, drops the undocumented `shell:"powershell"` field
- **Docs**: package.json install/uninstall scripts, README, setup guide, known limitations, runtime architecture, AGENTS; Spanish `traecodeEnableHint` (upstream added the `es` locale); TraeCode tests added to required CI

## Testing

- ✅ TraeCode-focused suites pass (hook, installer, state first-wins, idle-timeout cleanup, registry)
- ✅ Verified working on **macOS** and **Windows** (covers `Trae CN.exe` process-name matching and the PowerShell `&` call operator)

Closes #633


rullerzhou-afk:
Thanks for putting together the TraeCode integration. I reviewed the current head (`2e883621`) across the hook protocol, installer/cleanup, runtime lifecycle, tests, CI, and the current `main`. The registry/gating/Settings/Doctor wiring is generally complete, and keeping the integration disabled/uninstalled by default is correct. However, I do not think this is ready to merge yet.

The following are merge blockers:

1. **The `PreToolUse` stdout violates the current Trae Hook contract and the integration's state-only boundary.** `hooks/traecode-hook.js:118-121` emits top-level `{"decision":"allow"}`. Trae's documented tool decision shape is `hookSpecificOutput.permissionDecision`; since this integration explicitly does not own permission decisions, it should emit `{}` for every event. `test/hook-adapter-offline-contract.test.js` currently locks in the incorrect output and needs to be updated as well.

2. **Session IDs are not namespaced.** `hooks/traecode-hook.js:179-192` forwards the raw Trae session ID into a server map keyed only by session ID, so it can collide with another agent's session. Missing IDs also fall back to the global string `default`, creating a phantom session. Please report `traecode:<raw-id>` and skip the POST when the ID is missing/blank.

3. **The advertised product scope does not match the installer.** The PR registers international `Trae.exe` processes and describes general TraeCode support, but installation/detection/cleanup only use `~/.trae-cn/hooks.json`. International Trae uses `~/.trae/hooks.json`, so international-only users will receive no events. Please either use one shared path resolver for both products across installer, detector, Doctor, Settings uninstall, and About cleanup, or explicitly scope the first release to Trae CN and remove the international claims/process entries.

4. **The title lock persists user prompt text insecurely.** `hooks/traecode-title-lock.js` writes the first prompt line as plaintext JSON under `os.tmpdir()`, without a private per-user directory or explicit POSIX `0600` mode. Uninstall does not remove it, and the nominal 30-day cleanup only runs during a later successful claim, so data may remain indefinitely. It also scans the entire system temp directory on the hook hot path. My strong preference is to remove the disk title lock and implement first-successful-title-wins in the Clawd server/state layer.

5. **The title is claimed before the POST succeeds.** If Clawd is offline for the first prompt, the marker is already occupied and later prompts omit `session_title`, so the title may never reach Clawd. Please cover first-POST-fails/second-POST-succeeds explicitly; removing the disk lock also resolves this.

6. **Trae sessions are never retired while the IDE remains alive.** Trae has no `SessionEnd`, and the hook attaches the long-lived IDE PID. Generic stale cleanup only moves the session to idle while that PID is alive, so closed conversations remain in HUD/Dashboard and can accumulate up to the session limit. Please add a Trae desktop idle-timeout cleanup path comparable to Codex Desktop/ZCode, with local/remote/headless boundary tests.

7. **Importing the production hook starts its real lifecycle.** `test/traecode-hook.test.js` requires the hook for `__test`, but module import immediately reads stdin, starts the 800 ms safety timer, writes stdout, and eventually calls `process.exit(0)`. The current 11 tests completed in about 480 ms in my isolated run, but a slower worker or future async test can be truncated and reported green. Please move execution into `run()` guarded by `require.main === module`, and add an import-safety test.

The installer also needs hardening before merge: validate that the root/hooks structures are plain objects, fail closed on unsupported schema versions, perform a real conversion of owned flat entries to the documented nested shape, and remove the undocumented `shell: "powershell"` field. Please also avoid `stdout.write()` followed immediately by `process.exit()`, and either remove `traecode-title-lock.js` or add the new helper to the repository's required hook dependency manifest.

Testing/CI notes:

- The focused Trae suite passed 80 tests locally.
- Full `npm test` ran 8,243 tests: 8,207 passed, 35 skipped, and one unrelated existing/environmental Codex Windows non-ASCII dispatcher case failed.
- The green required `unit` check only ran 187 fixed packaging tests. It did not run the Trae hook, installer, title-lock, Doctor, integration-sync, Settings, detector, or cleanup tests. These tests, including the Windows offline contract, need to run in required PR CI.
- This branch is currently seven commits behind `main`. After updating it, please add the new Spanish `traecodeEnableHint` and update `package.json` install/uninstall scripts, README, setup guide, runtime architecture, AGENTS, and known limitations. The documentation should clearly state Experimental/state-only scope, supported product paths/platforms, lack of permission/SessionEnd support, and the manual Trae Hooks enable step.

Please address the blockers above before merge. I am happy to re-review a follow-up revision.

xiaoshidefeng:
Thanks for the thorough review. All blockers are addressed in `b7358c47` (plus `891dd8ff` syncing to current main and `68ed0d87` for docs/CI/i18n) - please re-review.

1. **PreToolUse stdout** - The hook now emits `{}` for every event. Trae's documented decision shape is `hookSpecificOutput.permissionDecision`, which we don't own, so no top-level `decision` is emitted anymore. The offline-contract row is updated to `{}\n`.

2. **Session ID namespacing** - The hook now reports `traecode:<raw-id>` and skips the POST entirely when the id is missing/blank (no `"default"` phantom session). Covered by tests including an `expect-none` spawned-hook case.

3. **Product scope** - Scoped to Trae CN: international `Trae.exe` / `TraeCN.exe` process entries removed from both `agents/traecode.js` and the hook's resolver config; installer/detector/cleanup stay on `~/.trae-cn/hooks.json`. The PR description and docs state the CN-only scope explicitly.

4. **Title lock removed** - `traecode-title-lock.js` and its test are deleted. The hook forwards the prompt-derived title on every event; the server keeps the **first** title per traecode session (`resolveIncomingSessionTitle` in `state.js`, `FIRST_WINS_TITLE_AGENT_IDS`). No user prompt text is persisted to disk by the hook anymore.

5. **Title claimed before POST** - Resolved by the server-side design: since the hook always forwards the candidate and first-wins lives in the server's session store, a failed first POST simply means the next prompt's title becomes the first *successful* one. Covered by a state test (first title wins, empty never clears).

6. **Idle-timeout cleanup** - Added `traecode-desktop-idle-timeout` in `state-stale-cleanup.js`, mirroring Codex Desktop/ZCode: local non-headless idle traecode sessions are deleted at the configured `sessionStaleMs` cutoff even while the IDE PID is alive. Boundary tests cover remote/headless/working/other-agent sessions.

7. **Import safety** - The whole lifecycle (stdin read, safety timer, stdout, POST, exit) now lives in `main()` guarded by `require.main === module`. Importing for `__test` arms nothing. Added a spawn-based import-safety test. Exit now happens in the stdout write callback so a pipe-backed response flushes before termination.

**Installer hardening** - Fails closed on non-object root/`hooks`/event arrays; owned flat entries are converted to the documented nested shape (`{ matcher: "", hooks: [{ type, command }] }`); the undocumented `shell: "powershell"` field is no longer written and is removed on migration.

**Docs/CI/i18n** - `package.json` install/uninstall scripts, README / setup guide / known limitations / runtime architecture / AGENTS entries (Experimental, state-only, CN-only, no permission/SessionEnd, manual Trae Hooks enable step), Spanish `traecodeEnableHint`, and the TraeCode suites are now part of the required `unit` job's test list and trigger paths.

**Known limitation worth highlighting** - this integration hard-depends on the user enabling hooks inside Trae (Settings -> Hooks -> Enable, run mode *Run Automatically Locally*). The enable state lives in Trae's internal storage, so there is no programmatic way for Clawd to detect or drive it: until the user flips that switch, the integration is silently inert (hooks.json being registered proves nothing). The setup guide and the Settings agent card call the step out, but it remains the biggest onboarding pitfall for this agent. If a future Trae release exposes hook state on disk, wiring detection into Doctor would fix the silent gap.

**Verification** - End-to-end on both platforms this time: macOS (Trae CN + live Clawd, confirmed the full UserPromptSubmit -> Stop -> Notification chain landing with the `traecode:` namespaced session, first-prompt title, and idle-timeout cleanup firing in session-debug) and Windows.

rullerzhou-afk:
Thank you @xiaoshidefeng for contributing the TraeCode integration and for all the iteration that got it to this point! The core state-only integration and macOS Sandbox path gave us a solid base to finish the Windows validation.

We pushed maintainer follow-up commit `771fbbeb` after the final Windows real-device pass. It adds:

- a quote-free PowerShell `-EncodedCommand` wrapper so Trae CN's Windows Sandbox preserves Node paths containing spaces;
- encoded-command-aware migration, idempotency, Node-path preservation, and uninstall coverage for all six managed events;
- fail-closed handling for any explicit unsupported `hooks.json` version, while keeping the documented missing-version-as-v1 behavior;
- Settings and Doctor guidance aligned to **Sandbox** across all seven locales, plus updated setup, architecture, release, and maintainer documentation;
- regression coverage for the actual PowerShell 5.1 native argv boundary and protection against accidental wrapper overrides.

Final Windows verification used Trae CN 3.3.94 in Sandbox mode. One real session completed `SessionStart → UserPromptSubmit → PreToolUse(Read) → PostToolUse → Stop → Notification`; all six hooks exited 0, returned `{}`, produced no stderr, and the task replied `READY`. The focused suite passed 1030/1030, and GitHub's unit/audit/Windows x64 checks are green.

Thanks again for bringing TraeCode support to Clawd! 🦀

PR #889, opened 2026-08-16
fix(win): stop treating a maximized window as a fullscreen app

### LINKED ISSUE #871: 焦点丢失问题
我操作一下屏幕1  再到 屏幕2 上 点击输入框 焦点会丢失  比如百度的输入框这些
版本	Windows 10 专业版
版本号	22H2

KaiC5504:
## The bug

`isForegroundFullscreen()` decided from geometry alone: does the foreground window's rect cover `rcMonitor`? That test carries an assumption stated in its own comment — *"a maximized normal window covers the work area but leaves the taskbar strip, so its rect stops short of the full monitor."*

That only holds on a monitor that **reserves** a taskbar strip. A secondary display usually shows no taskbar, and an auto-hidden taskbar reserves a sliver that fits inside `FULLSCREEN_TOLERANCE_PX`. In both cases `rcWork == rcMonitor`, so an ordinary **maximized** window covers the monitor and gets classified as a fullscreen app.

From there the existing machinery does the damage: the ~1s focusable poll from #562 flips the hit window non-activating, and Electron's `setFocusable(false)` ends in `Focus(false)` on Windows, which deactivates the foreground app. Clicking an input box in a maximized browser on such a monitor loses focus — repeatedly, until the state settles, because `setHitWinFocusable` is gated on `isFocusable()` and so only fires on transitions.

Same failure family as #719 — a second blind spot in the same probe.

Each detail in the report maps to the mechanism:

| reported | mechanism |
|---|---|
| started at v0.11.0 | #562's focusable poll first shipped in v0.11.0 |
| screen 2 only, never screen 1 | screen 1 reserves the taskbar, so the rect stops short there |
| maximized breaks it, a small window doesn't | a windowed browser doesn't cover the monitor |
| loses focus several times, then settles | the call is change-gated, so it fires only on transitions |
| hiding the pet fixes it | a hidden hit window can't deactivate anything |

## The fix

Read the window style and treat a window that is **maximized AND still owns its caption** as a normal window.

This is deliberately narrower than "has a caption": it only overrides the verdict for maximized windows, so exclusive fullscreen, a borderless `WS_POPUP`, and F11 (which drops the caption) all keep the answer they had — the #538 / #562 stand-down behavior is unchanged.

The style binding is resolved in its own `try`/`catch`. `GetWindowLongPtrW` is a macro over `GetWindowLongW` on 32-bit Windows, so the export can be missing; folding it into the main block would turn a missing symbol into a constant-false probe and silently undo the #538 stand-down. A `0` read — its documented failure return — keeps the geometric answer, matching the existing class-read fallback directly above it.

## Verification

- `test/win-fullscreen-detect.test.js`: 18 → 29 tests. Full suite 8194 → 8205 tests, 8159 → 8170 pass, skipped unchanged. The regression test would fail if the exact bug came back.
- Checked against the real `user32.dll`, not only the fake koffi the unit tests use: the export binds with the signature used here, and a live window sweep confirmed maximized Chromium/Electron windows carry `WS_MAXIMIZE|WS_CAPTION` while every monitor-covering window on the machine was caption-less. The discriminator separates the two populations it claims to.
- Live A/B of the actual probe, before vs after, with the taskbar auto-hidden to produce `rcWork == rcMonitor` (the same geometry a taskbar-less second display gives). The maximized windows re-maximized from bottom `823` to `871` on an `864px` monitor and began covering it; the old code answered `fullscreen=true` for all three, the new code `false`. The lock screen, Windows Input Experience and the NVIDIA overlay stayed `true` in both.

## Residual risk

Verified on a single monitor, using auto-hide to synthesize `rcWork == rcMonitor`. The geometry is identical to the reporter's setup, but I have not tested a physical dual-monitor machine — that is the one check worth repeating on your side.

A fullscreen presentation that kept *both* a caption and the maximized state would now be treated as a normal window. None of the monitor-covering windows observed had that shape, and such a window is visually indistinguishable from a maximized one, but it is the behavior change this makes.

Fixes #871

### DIFF
--- a/src/win-fullscreen-detect.js
+++ b/src/win-fullscreen-detect.js
@@ -16,10 +16,11 @@
 // returns a probe that always answers false, so the watchdog keeps its current
 // behavior rather than ever hiding the pet because the FFI broke.
 
-// A maximized normal window covers the work area but leaves the taskbar strip,
-// so its rect stops short of the full monitor. A fullscreen app covers the
-// whole monitor (rcMonitor). A couple of px of slack absorbs DPI rounding
-// without mistaking a maximized window for fullscreen.
+// A maximized normal window covers the work area, so on a monitor that reserves
+// a taskbar strip its rect stops short of the full monitor while a fullscreen
+// app covers the whole monitor (rcMonitor). A couple of px of slack absorbs DPI
+// rounding. Geometry alone is NOT sufficient though — see isMaximizedNormalWindow
+// for the monitors where rcWork == rcMonitor and this test stops separating them.
 const FULLSCREEN_TOLERANCE_PX = 2;
 
 const MONITOR_DEFAULTTONEAREST = 2;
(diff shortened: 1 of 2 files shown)


rullerzhou-afk (approved):
Thanks, KaiC — this is an exceptionally well-scoped and carefully validated fix. The WS_MAXIMIZE + WS_CAPTION discriminator addresses #871 while preserving the fullscreen protections from #538/#562, and the fallback/regression coverage is excellent. I especially appreciate the single-monitor auto-hide A/B verification and the clearly documented residual risk. Approved — thank you for the timely, thoughtful contribution!

PR #892, opened 2026-08-17
fix(validate-theme): distinguish "could not validate" from "theme has errors"

chrono-meta:
## The problem

**`--assets` passed with no value exits 0.** The override is silently dropped, every check then runs against the default asset directory, and the tool tells the author their theme is valid. A validator that reports success for a command the user got wrong is the worst thing in this file, and it is reachable on `main` today — measured on `upstream/main` (`fb686b82`), run from a real checkout.

It stayed invisible because every non-success path in this script exited `1`, so nothing distinguished these three states:

- **validation could not run** — no theme directory given, `theme.json` missing, `theme.json` unreadable or not an object
- **validation ran and the theme is broken** — N schema errors
- **validation silently didn't do what was asked** — the `--assets` case above, which didn't even fail

A theme author who mistypes a path goes hunting for theme bugs that do not exist, and a wrapper script cannot tell "this theme failed validation" from "validation never happened".

## The change

The "could not validate" paths now exit `2`. `1` still means "validated, and it has errors"; `0` is unchanged. The contract is documented in the script header, in `README.md`, and in `docs/guides/guide-theme-creation.md`.

**The exit-code split and the defects are separable, and I'll take them apart if you prefer.** Splitting the code is what surfaced four places where the file did not honour the contract it was about to declare — but three of those are plain defects that stand on their own, with or without any contract:

| Input | Before | After | Stands alone? |
|---|---|---|---|
| `--assets` passed with no value | silently ignored, **exit 0** | exit **2** | **yes — false green** |
| `theme.json` parses to `null` / a number / a string / a boolean | uncaught `TypeError`, exit **1** | exit **2** | **yes — stack trace, not a message** |
| `--assets <dir>` where `<dir>` does not exist | exit **1** | exit **2** | **yes — reads as a theme verdict** |
| `theme.json` exists but is unreadable (e.g. it is a directory) | `Failed to parse theme.json: EISDIR` | `Failed to read or parse theme.json: EISDIR` | wording only |

The `null` row is the one that would have made the contract false three lines below where it is written: the crash exits `1`, which this PR defines as "validated and has errors" — while nothing was validated at all.

Two deliberate non-changes:

- A missing **default** `<theme>/assets` directory is still exit `1`. That is a theme defect, not a setup mistake. Only an explicit `--assets` override that does not resolve is treated as setup.
- `theme.json` parsing to `[]` is still exit `1`. An array is an object; it reaches the schema checks and fails them honestly. There is a test asserting this, so the guard cannot silently widen.

## Tests

`test/validate-theme-exit-codes.test.js` drives the real CLI with `spawnSync` and asserts `status`, rather than importing internals — the exit code is the observable thing downstream callers read.

Each guard was checked by reverting it and confirming that **exactly** the tests for that guard go red and no others:

| Guard reverted | Result |
|---|---|
| non-object `theme.json` | the 4 `null`/number/string/boolean cases red; the `[]` control stayed green |
| value-less `--assets` | 1 case red |
| non-existent `--assets` dir | 1 case red |
| read-vs-parse message | the 2 message-dependent cases red |

## About the pre-existing test failures

This branch is rebased on `upstream/main` (`bca6a8d3`), no conflicts.

I ran `node test/run-tests.js` to completion on `upstream/main` and on this branch, extracted the failing test names from both (ANSI stripped, `^\s*✖ ` lines, `sort -u`), and diffed the two sets:

- names present only on this branch: **none**
- names present only on `upstream/main`: **none**

The pre-existing failures are in the DSH integration lanes and are identical before and after. I have deliberately not touched them — they are unrelated to the exit-code contract, and I did not want to widen this PR.

I am quoting the set difference rather than a failure count on purpose: the runner emits nested markers, so different ways of counting give different totals. The set comparison does not depend on that.

## Not included

The four locale READMEs (`README.*.md`) also show the validate command but were left alone — happy to follow up with those, or to fold them in here if you'd prefer.

### DIFF
--- a/scripts/validate-theme.js
+++ b/scripts/validate-theme.js
@@ -53,17 +53,62 @@ const {
 const args = process.argv.slice(2);
 let themeDir = null;
 let assetsOverride = null;
-for (let i = 0; i < args.length; i++) {
-  if (args[i] === "--assets" && args[i + 1]) {
-    assetsOverride = args[++i];
-  } else if (!themeDir) {
-    themeDir = args[i];
-  }
-}
-if (!themeDir) {
+
+function printUsage() {
   console.error(`Usage: node ${path.basename(process.argv[1])} <theme-directory> [--assets <assets-dir>]`);
   console.error(`Example: node scripts/validate-theme.js themes/template`);
   console.error(`         node scripts/validate-theme.js themes/clawd --assets assets/svg`);
+}
+
+// Anything the parser does not understand is refused rather than ignored. The old
+// loop silently dropped unknown flags, surplus positional arguments, and a
+// value-less `--assets`, so a mistyped command ran against the DEFAULT paths and
+// then reported the theme as valid -- the tool answering a question nobody asked.
+//
+// Refusing unknown options costs the one thing the permissive loop got for free: a
+// path that genuinely begins with "-". `--` ends option parsing for that case, so
+// no invocation that worked before is unreachable now.
+let optionsEnded = false;
+for (let i = 0; i < args.length; i++) {
+  const arg = args[i];
+  if (!optionsEnded && arg === "--") {
+    optionsEnded = true;
+  } else if (!optionsEnded && arg === "--assets") {
+    if (assetsOverride !== null) {
+      console.error(`${FAIL} --assets given more than once`);
+      printUsage();
+      process.exit(1);
+    }
+    const value = args[i + 1];
+    // Only absent or empty counts as "no path given". A value that merely looks
+    // like a flag is still passed to the filesystem, which is the thing that
+    // actually knows whether it is a usable directory.
+    if (value === undefined || value === "") {
+      console.error(`${FAIL} --assets requires a directory argument`);
+      printUsage();
+      process.exit(1);
+    }
+    assetsOverride = value;
+    i++;
+  } else if (!optionsEnded && arg.startsWith("-") && arg !== "-") {
+    console.error(`${FAIL} Unknown option: ${arg}`);
+    console.error(`  (if that is a real path, pass it after --)`);
+    printUsage();
+    process.exit(1);
+  } else if (themeDir === null) {
+    themeDir = arg;
+  } else {
+    console.error(`${FAIL} Unexpected extra argument: ${arg}`);
+    printUsage();
+    process.exit(1);
+  }
+}
+// Falsy, not just null: `validate-theme.js "$SOME_UNSET_VAR"` passes an empty
+// string, which path.resolve() turns into the current directory -- so without
+// this the script would go and validate whatever the caller happened to be in.
+if (!themeDir) {
+  console.error(`${FAIL} No theme directory given`);
+  printUsage();
   process.exit(1);
 }
 
--- a/scripts/validate-theme.js
+++ b/scripts/validate-theme.js
@@ -153,6 +153,17 @@ try {
   process.exit(1);
 }
 
+// theme.json has to be a JSON object for the checks below to walk it. `null` crashes
+// property access with an uncaught TypeError (a stack trace instead of a message),
+// and a number/string/boolean produces meaningless FAIL lines that read like real
+// findings about a theme that was never there. An array is deliberately let through:
+// nothing below it crashes, and its total absence of fields is a genuine "theme has
+// errors" outcome rather than a broken command.
+if (!isPlainObject(raw) && !Array.isArray(raw)) {
+  console.error(`${FAIL} theme.json must contain a JSON object (got: ${raw === null ? "null" : typeof raw})`);
+  process.exit(1);
+}
+
 console.log(`\n${C}Validating theme:${D} ${resolvedDir}\n`);
 
 let errors = 0;
--- a/scripts/validate-theme.js
+++ b/scripts/validate-theme.js
@@ -149,7 +149,9 @@ try {
   const content = fs.readFileSync(jsonPath, "utf8");
   raw = JSON.parse(content);
 } catch (e) {
-  console.error(`${FAIL} Failed to parse theme.json: ${e.message}`);
+  // The try block covers readFileSync as well as JSON.parse, so this must not
+  // claim "parse" for what may have been a read failure (e.g. EISDIR).
+  console.error(`${FAIL} Failed to read or parse theme.json: ${e.message}`);
   process.exit(1);
 }
 
--- a/scripts/validate-theme.js
+++ b/scripts/validate-theme.js
@@ -115,6 +115,30 @@ if (!themeDir) {
 const resolvedDir = path.resolve(themeDir);
 const jsonPath = path.join(resolvedDir, "theme.json");
 
+// An explicit --assets override that does not resolve to a usable directory is a
+// mistake in the command, not a property of the theme -- so the message has to say
+// which one it is. Pointing --assets at a file used to produce a 27-error report
+// about the theme, sending the author hunting for theme bugs that do not exist.
+// The default `<theme>/assets` being absent is deliberately NOT handled here: that
+// one really is a property of the theme, and stays a normal validation error below.
+if (assetsOverride !== null) {
+  const overrideDir = path.resolve(assetsOverride);
+  let overrideStat;
+  try {
+    overrideStat = fs.statSync(overrideDir);
+  } catch (e) {
+    // Only ENOENT means "not there". ENOTDIR, ELOOP, EACCES and friends are all
+    // different problems and must not be flattened into one claim.
+    const why = e.code === "ENOENT" ? "not found" : `cannot be accessed (${e.code})`;
+    console.error(`${FAIL} --assets directory ${why}: ${overrideDir}`);
+    process.exit(1);
+  }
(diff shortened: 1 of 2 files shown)


rullerzhou-afk:
Thanks for the thoughtful PR. Before we decide how to proceed, could you clarify the concrete use case for distinguishing exit code 1 from 2? Is there an existing CI workflow, wrapper script, or theme publishing tool that needs this distinction today? We couldn’t find a current consumer in the repository, so we’d like to understand who would rely on this contract before adding and maintaining it.

chrono-meta:
Fair question, and the honest answer is: **there is no current consumer.** My own first commit message says so — I checked before writing the contract, and nothing in the repo calls this script programmatically. It is a manual command in the READMEs and in `docs/guides/guide-theme-creation.md`, plus the line `scripts/create-theme.js` prints at the end.

So let me separate the two halves of this PR, because only one of them depends on that question.

**1. The exit-code split (1 vs 2) — speculative, and I'll defer to you.** The consumer today is a human reading the message, not a script reading `$?`. My argument is only that `2` is the conventional slot for a usage error, so that the first wrapper someone writes doesn't have to reverse-engineer it. That's a preference, not a need. If you'd rather not carry the contract, I'll drop it.

**2. Three defects that are reachable on `main` today, with or without a consumer.** Measured just now on `upstream/main` (`fb686b82`), running the script from a real checkout:

| Input | `main` behavior |
|---|---|
| `theme.json` containing `null` (or a number/string/bool) | uncaught `TypeError: Cannot read properties of null (reading 'schemaVersion')` — a stack trace, not a validation message |
| `--assets` passed with no value | **exit 0** — the override is silently dropped and the tool reports the theme as valid |
| `--assets <dir>` that doesn't exist | fails, but as "theme has errors" rather than "your path is wrong" |

The middle row is the one I'd ask you to look at regardless of what happens to the exit codes: a theme author who mistypes that flag is told their theme passed. That's a false green from a validator, and the person it misleads is a human, not a CI job.

**Happy to reshape this however you prefer** — the simplest version is that I close this PR and open a small one containing only the three fixes above, leaving every exit code exactly as it is on `main`. Say the word and I'll do that. If you'd rather keep it as one PR with the contract in it, it's already here and tested.

rullerzhou-afk:
Thank you for the clear and candid explanation — separating the two concerns helps a lot. I agree that the false green and the uncaught stack trace are real issues worth fixing.

For the exit-code contract, I’d prefer to keep the validator simple for now: `0` for success and `1` for any failure, with clear, actionable messages explaining whether the problem is the arguments, `theme.json`, or the assets path. Since there is no current machine consumer, I don’t think we should add and document a `1`-vs-`2` distinction yet; we can introduce a machine-readable contract when a concrete consumer actually requires it.

Please feel free to reshape this PR rather than close it and open another one. Could you drop the exit-code-`2` split and its documentation, while keeping the focused fixes for the false green, malformed top-level JSON, and invalid `--assets` inputs? A missing or empty value, or a path that is not a usable directory, should fail clearly with exit `1`. It would also be good to reject unknown or extra arguments rather than silently succeeding.

Once updated, we’ll review the new head. Thank you again for being flexible.

chrono-meta:
Done — reshaped rather than closed. Exit code `2` is gone, every failure exits `1`, and the message is now what carries the distinction between a bad command, a bad `theme.json`, and a bad assets path. There are no documentation changes: I rebuilt the branch on `upstream/main` (`fb686b82`), so the exit-code tables from the previous revision are not there any more. The diff is `scripts/validate-theme.js` plus one test file.

All three fixes you asked me to keep are still in:

- **The false green.** `--assets` with a missing or empty value now fails instead of silently dropping the override.
- **Malformed top-level `theme.json`.** `null` used to crash with an uncaught `TypeError` before any message was printed; a number, string, or boolean produced FAIL lines about a theme that was never parsed. Both now exit 1 with `theme.json must contain a JSON object (got: null)`.
- **Invalid `--assets` input.** A path that is not a usable directory now says which problem it is: `--assets directory not found: <path>` when it does not exist, `--assets path is not a directory: <path>` when it is a file, and `cannot be accessed (ENOTDIR)` and friends when it is neither. Previously all of these came out as a verdict about the theme.

Your note about unknown and extra arguments covered more ground than I expected. Four different commands exit 0 on `main` today; the first is the `--assets` case above, the other three are ones I had not looked at:

| Command, on `upstream/main` (`fb686b82`) | Today | After |
|---|---|---|
| `validate-theme.js themes/calico --assets ""` | exit 0, the override is dropped | exit 1, `--assets requires a directory argument` |
| `validate-theme.js themes/calico --verbose` | exit 0, the flag is discarded | exit 1, `Unknown option: --verbose` |
| `validate-theme.js themes/calico themes/clawd` | exit 0, the second path is discarded | exit 1, `Unexpected extra argument` |
| `validate-theme.js themes/calico --assets A --assets B` | exit 0, `A` is discarded | exit 1, `--assets given more than once` |

Each of those runs checks the caller did not ask for and then calls the theme valid.

The `--assets`-at-a-file case is worth one more line, because the old output actively misdirects: pointed at `themes/calico/theme.json`, the script printed **`27 error(s)`** for `themes/calico`, a theme that validates clean. Asset lookups miss under a path that was never a directory, so the author goes hunting for theme bugs that do not exist.

Being strict about arguments has a cost, and here is where it lands:

- Refusing unknown options makes a directory whose name begins with `-` unreachable, and that used to work. `--` now ends option parsing, and the unknown-option message points at it.
- A flag-shaped value after `--assets` is not treated as missing — it is handed to the filesystem, so `--assets --verbose` fails as `directory not found: .../--verbose` rather than guessing at intent.
- The theme-directory check is falsy rather than `=== null`, because `path.resolve("")` is the current directory: `validate-theme.js "$SOME_UNSET_VAR"` would otherwise have validated whatever directory you were standing in.

I got the first two of those wrong initially and caught them by running each case against `main` instead of reading my own diff. All three have tests.

I deliberately did not change two things. A missing default `<theme>/assets` stays an ordinary theme finding — only an explicit override that does not resolve is treated as a bad command. And a `theme.json` containing `[]` still flows into the normal checks, since nothing crashes on it and listing the fields it lacks is more useful than a blanket rejection. Both have tests pinning the boundary so neither guard widens into them later.

The test file is renamed to `test/validate-theme-cli.test.js`, since it is not about exit codes any more: 24 cases driving the real CLI through `spawnSync`, asserting on messages rather than status, because status is now always `1`.

I checked the tests by removing each guard in turn and confirming that the only tests that go red are the ones asserting that guard. The controls (array `theme.json`, a missing default assets dir, a well-formed `--assets`, an assets path beginning with `-`) stayed green through every one of those reverts, which is the half that catches a test lane that fails for the wrong reason.

On regressions: I ran `node test/run-tests.js` to completion on `upstream/main` and on this head, extracted the failing test names from both (ANSI stripped, `sort -u`), and diffed the sets. Nothing appears only on this branch, and nothing only on `upstream/main` — the DSH integration failures are identical before and after, and I have left them alone.

rullerzhou-afk:
Merged. Thank you — and three things I want to name specifically, because they are why this went in without another round.

**You reshaped instead of defending.** When I asked you to drop the exit-code split, you rebuilt the branch and came back with the contract gone and the diff smaller — and you had already offered to do exactly that before I asked. That made the decision easy.

**Your commit messages measure `main` instead of describing the diff.** "Measured on upstream/main (fb686b82)" with the real observed output — including the `27 error(s)` report for a theme that validates clean — is worth more than any amount of prose about intent. I reproduced every row of your table on current `main`: all six exit-0 cases and the `null` stack trace were exactly as you described.

**And this line:** *"I got the first two of those wrong initially and caught them by running each case against main instead of reading my own diff."* That is the habit that separates a fix that works from one that looks like it works. The controls in the test file come from the same instinct — a guard with no test pinning its boundary is a guard that widens later.

For the record on my side, since this repo does not run the suite on PRs:

- The branch merges clean against current `main`, which had moved past your base to include #866's accessory-schema block in this same file. I built the merged tree and confirmed the `[]` case you deliberately let through still survives the new unconditional `themeSchema.validateTheme(raw)` call.
- Full suite on the merged tree: 8333 tests, +24 passing, and the failing-test-name set is byte-identical to `main` in both directions — your regression claim holds against current `main`, not just your base.
- I mutation-tested your suite as well: 18 of 20 mutations caught, including all 13 obvious ones. Messages are pinned, not just exit codes.

Two small things I am following up on myself — no action needed from you:

- `control: a missing DEFAULT <theme>/assets stays a theme finding` cannot actually detect what it names. `check()` prints the same text for pass and fail (only the glyph differs), and the `states: {}` fixture guarantees exit 1 on its own — so the assertions still hold with `check(assetsDirExists, …)` replaced by `check(true, …)`. Pinning the FAIL glyph closes it.
- `validate-theme-cli.test.js:154` hard-codes `cannot be accessed (ENOTDIR)`. Windows most likely maps that path to `ENOENT`, and our Windows `npm test` only runs at release-tag time — so it would surface as a blocked release rather than a red PR.

Both are a couple of lines, and neither touches the fix itself.

Thanks again — that is four of yours that have landed now.

PR #902, opened 2026-08-19
fix(install): stop shipping a broken hooks/ subset in the WSL setup docs, and fail loudly when one is installed

### LINKED ISSUE #901: Docs: WSL manual setup file list is 4 months stale — every installed hook dies with MODULE_NOT_FOUND, while install.js reports success
## Summary

The manual WSL setup instructions in `docs/guides/setup-guide.md` tell you to copy a hand-picked list of 13 hook files. That list has not been updated since 2026-04-18, while `clawd-hook.js` gained three new local dependencies between June and July. Following the documented steps produces an install that **reports complete success and then fails on every single hook invocation**.

Nothing reaches the desktop app, and there is no visible error anywhere — the hook writes its crash to stderr, which Claude Code discards for `async: true` hooks. From the user's side, Clawd simply never moves.

## Environment

- clawd-on-desk `v0.15.0` @ `3542baa`
- Windows 11 host + WSL2 (`networkingMode=mirrored`, verified working)
- Claude Code 2.1.235, Node.js v24.12.0

## Reproduction

Follow `docs/guides/setup-guide.md` → "WSL (Windows Subsystem for Linux)" → Configuration steps verbatim:

```bash
mkdir -p ~/.claude/hooks
cp /path/to/clawd-on-desk/hooks/{server-config,json-utils,shared-process,clawd-hook,install,codex-hook,codex-install,codex-install-utils,codex-remote-monitor,codex-session-index,codex-subagent-fields,copilot-hook,copilot-install}.js ~/.claude/hooks/
node ~/.claude/hooks/install.js --remote
```

The installer reports success:

```
Clawd hooks installed to /home/<user>/.claude/settings.json
  Claude Code version: 2.1.235
  Added: 15 hooks
Clawd Claude Code statusline -> /home/<user>/.claude/settings.json (updated)
```

Now trigger any hook:

```console
$ echo '{"session_id":"t","hook_event_name":"UserPromptSubmit"}' \
    | CLAWD_REMOTE=1 node ~/.claude/hooks/clawd-hook.js UserPromptSubmit
Error: Cannot find module './state-payload-size'
Require stack:
- /home/<user>/.claude/hooks/clawd-hook.js
    at Object.<anonymous> (/home/<user>/.claude/hooks/clawd-hook.js:9:38)
  code: 'MODULE_NOT_FOUND'
$ echo $?
1
```

All 15 registered hooks are dead on arrival. Note this is **not** the usual WSL2 networking problem — connectivity was independently confirmed:

```console
$ node -e "require('http').get({hostname:'127.0.0.1',port:23333,path:'/state'},r=>console.log(r.statusCode, r.headers['x-clawd-server']))"
200 clawd-on-desk
```

## Root cause 1 — the documented file list is a stale subset of the authoritative manifest

`src/remote-ssh-deploy.js:57` already declares itself the single source of truth:

```js
// ── Hook files manifest ──
//
// Single source of truth.
const HOOK_FILES = [ ... ];   // 27 entries
```

The docs list only 13. The 14 missing files are:

```
pid-cache.js                    context-usage.js
antigravity-context-usage.js    claude-rate-limits.js
claude-statusline.js            codex-rate-limits.js
quota-bucket.js                 state-payload-size.js
claude-stop-disposition.js      session-recovery-lease.js
uninstall.js                    codex-originator.js
codex-assistant-output.js       codex-user-input.js
```

Three of them are hard `require`s at the top of `clawd-hook.js:9-12`:

```js
const { fitStateBodyToByteBudget } = require("./state-payload-size");
const { extractClaudeContextUsageFromEntries } = require("./context-usage");
const { updateRecoveryLeaseFromStateBody } = require("./session-recovery-lease");
```

A fourth, `claude-statusline.js`, is what `install.js` writes into the `statusLine` setting — so the documented flow also configures a status line pointing at a file it never copied.

Timeline:

| Date | Event |
|---|---|
| 2026-04-18 | `cp` list last touched (`08cc5011`, a docs reorg) |
| 2026-06-07 | `context-usage.js` added (`fa3c259a`) |
| 2026-06-28 | `state-payload-size.js` added (`eaffaeb9`) |
| 2026-07-23 | `session-recovery-lease.js` added (`887c8cfe`) |

`docs/guides/setup-guide.zh-CN.md:162` carries the identical stale list, so both languages are affected. That line also still hardcodes `/mnt/d/animation/hooks/`, which looks like a leftover authoring path.

The in-app **Settings → Agents → WSL Scan → Pair** flow uses the real `HOOK_FILES` manifest and is unaffected — only people following the written steps hit this.

## Root cause 2 — `install.js` validates nothing

`hooks/install.js` writes `settings.json` and prints `Added: 15 hooks` without ever checking that the scripts it just registered can be resolved. A single `require.resolve()` per registered entry point, or one dry-run invocation, would have turned a silent four-month-old trap into an immediate actionable error.

This is the part that makes the bug expensive to diagnose: the installer's success message actively points you away from the real problem, so you go audit WSL2 mirrored networking, firewall rules, and ports 23333-23337 instead.

## Suggested fixes

1. **Replace the hand-maintained list with a directory copy** in both `setup-guide.md` and `setup-guide.zh-CN.md`:
   ```bash
   cp /path/to/clawd-on-desk/hooks/*.js ~/.claude/hooks/
   ```
   This cannot drift. Also worth reordering the section so `git clone` + `node hooks/install.js` (which runs inside the repo and is therefore always complete) is the primary path, and manual `cp` the fallback — currently it is the other way round.

2. **Make `install.js` fail loudly** — verify every registered hook script resolves before declaring success, and exit non-zero listing what is missing.

3. Optionally, have the docs generate their list from `HOOK_FILES` in CI, so the stated single source of truth actually is one.

Happy to send a PR for (1) and (2) if that is useful.

---

sunnyswag:
Fixes #901.

Following the manual WSL setup steps in `docs/guides/setup-guide.md` produces an install that reports complete success and then fails on every single hook invocation. Two independent causes, one commit each.

## 1. `docs(wsl)` — the documented `cp` list was a stale subset

The guide pinned an explicit 13-file `cp`, last touched 2026-04-18. `clawd-hook.js` has since gained three local requires:

| Added | File |
|---|---|
| 2026-06-07 (`fa3c259a`) | `context-usage.js` |
| 2026-06-28 (`eaffaeb9`) | `state-payload-size.js` |
| 2026-07-23 (`887c8cfe`) | `session-recovery-lease.js` |

None are in the documented list, so `clawd-hook.js:9` throws `MODULE_NOT_FOUND` on every event. `claude-statusline.js` — which `install.js` writes into `statusLine` — was missing too, and so was `pid-cache.js`, a transitive dependency of `shared-process.js`.

Both language guides now copy `hooks/*.js`. This is not a new convention: `collectHookFiles()` in `src/wsl-deploy.js:35` already ships every `.js` in the directory, so the in-app **WSL Pair** flow was always doing the right thing and only the written instructions drifted. The leftover authoring path `/mnt/d/animation/` is replaced with a placeholder.

A test in `test/remote-deploy.test.js` now rejects brace-enumerated hook lists in either guide. The existing `HOOK_FILES` closure test cannot police prose, which is why this went unnoticed for four months.

## 2. `fix(install)` — the installer validated nothing

`hooks/install.js` wrote `settings.json` and printed `Added: 15 hooks` without checking that the scripts it just registered could be resolved. Since Claude Code discards the stderr of `async: true` hooks, the failure was invisible from both ends — and the success message actively misdirects diagnosis toward WSL2 mirrored networking and ports 23333-23337, which is where I spent my time before finding the real cause.

The CLI entry point now walks the relative-require closure of the entry points it is about to register:

```console
$ node ~/.claude/hooks/install.js --remote
Clawd: refusing to install — the hooks directory is incomplete.

  missing: claude-statusline.js
  missing: state-payload-size.js  (required by clawd-hook.js)
  missing: context-usage.js  (required by clawd-hook.js)
  missing: session-recovery-lease.js  (required by clawd-hook.js)
  missing: pid-cache.js  (required by shared-process.js)

Hook scripts require each other, so copying a subset of hooks/ produces
an install that registers cleanly and then fails on every invocation.
Copy the whole directory instead:

  cp /path/to/clawd-on-desk/hooks/*.js ~/.claude/hooks/
```

Scope notes, in line with `AGENTS.md`:

- **Only what the invocation registers.** `auto-start.js` is opt-in and deliberately absent from `HOOK_FILES`, so demanding it unconditionally would reject complete remote/WSL installs. `claude-statusline.js` is checked only when the statusline is being installed.
- **CLI-only.** In-process callers keep going through the server-owned queue in `src/claude-hook-operations.js` and are deployed from the `HOOK_FILES` manifest, which `test/remote-deploy.test.js` already covers. No behavior change there.
- Requires that climb out of `hooks/` resolve only inside the packaged app tree and are skipped; a present-but-unreadable file is left alone rather than guessed at.

## Verification

Reproduced the original failure, then confirmed both paths:

- 13-file subset + `install.js --remote` → exits 1 with the message above, `settings.json` untouched
- Full `hooks/*.js` + `install.js --remote` → `Added: 15 hooks`, statusline configured, exit 0 (run against an isolated `HOME`)

Tests — every suite that requires `hooks/install`, plus the deploy-manifest suites:

| Suite | Result |
|---|---|
| `install` | 112 pass / 0 fail (7 new) |
| `server-hook-management` | 54 pass / 0 fail |
| `claude-hook-health` | 39 pass / 0 fail |
| `remote-ssh-deploy` | 57 pass / 0 fail |
| `wsl-deploy` | 33 pass / 0 fail |
| `claude-settings-watcher` | 29 pass / 0 fail |
| `doctor-agent-descriptors` | 14 pass / 0 fail |
| `remote-deploy` | 5 pass / 0 fail (2 new) |
| `doctor-installer-defaults`, `uninstall-entrypoint` | pass / 0 fail |

**Not run:** the full `npm test` suite — this machine has no `node_modules` and the environment is Linux/WSL, so per `AGENTS.md` I would not be able to present a full-matrix result honestly. The changed code is plain Node with no new dependencies, but a CI run on the complete suite is worth having before merge.

---


rullerzhou-afk (commented):
Thanks for the detailed report and for fixing both the stale WSL instructions and the installer's false-success path. I independently reviewed exact head `e2dbdd95f9f7d16cc1c3191c1103490bf0635826` against current `main` (`54480d10e56347dcd628c1a7f74c8888f095e14a`).

What I verified:

- On the old 13-file subset, the new CLI exits 1, reports the missing entry points/dependencies, and does not create `settings.json`.
- With the complete `hooks/*.js` payload, the CLI exits 0 and the Claude hook/statusline entry points run.
- `test/install.test.js` plus `test/remote-deploy.test.js` pass 117/117 both at the PR head and after applying the PR to current `main`.
- The PR merges cleanly with current `main`. The unrelated DSH failures from the full suite reproduce unchanged on `main`, so I do not attribute them to this PR.

There is one change I recommend before merge: please add an end-to-end test for the actual CLI gate in `if (require.main === module)`.

All seven new dependency tests call `__test.findMissingHookDependencies()` directly; none spawns `node hooks/install.js`. As a result, they do not protect the user-visible contract if the CLI wiring is later removed or misordered, if `claude-statusline.js` is accidentally dropped from `entryPoints`, or if the error path exits 0. This matters especially because this PR currently has no applicable PR check running the full test suite.

A focused child-process regression test should be enough:

1. Create an isolated `HOME` plus the exact old 13-file subset.
2. Run `node <partial-hooks>/install.js --remote`.
3. Assert exit code 1 and that `~/.claude/settings.json` was not created.
4. Add a positive complete-directory case asserting exit code 0.

Two non-blocking notes:

- The `readFileSync()` catch currently treats a present-but-unreadable entry point or transitive dependency as healthy. I reproduced an install that exits 0 and writes 15 hooks, followed by an `EACCES` failure when the hook runs. This behavior also exists on the base commit and is explicitly documented/tested in this PR, so I do not consider it a regression or a reason to hold the main fix. A follow-up should report unreadable files as an error or at least a warning rather than silently continuing.
- Minor chronology correction for the PR description: the enumerated WSL copy line was last changed on 2026-05-13 in `3dc6be9c` to add the Copilot files, not last touched on 2026-04-18. It still became stale when the first missing dependency landed on 2026-06-07, so this does not change the root cause.

The main fix is sound. My independent runtime checks were on macOS/Node 24 rather than a real Windows 11 + WSL2 device, so I am not treating them as WSL platform validation; the reporter's WSL evidence remains the platform-specific proof.

rullerzhou-afk:
Thanks again @sunnyswag for tracking down #901 and for the original fix. I've pushed maintainer follow-ups to this PR at `8c4fda8f0a96aca8962e276d06f9078ac5babf2a`, preserving both of your original commits and authorship.

The follow-ups:

- Add real CLI child-process regressions for the historical partial payload and the complete payload, including exit status and byte-for-byte preservation of existing configuration. Unreadable files, directories in place of scripts, and dependencies outside `hooks/` now fail preflight rather than being silently accepted.
- Cover default installation, `--statusline`, and legacy `--remote`, and add a focused installer/documentation step to the existing PR workflow.
- Refresh the English/Chinese setup guides and English/Chinese/Korean Codex/WSL notes around the existing Scan/Pair flow, agent enablement, complete-file copying, and the different Linux Node / Windows interop paths. Claude's default manual WSL command is now `node hooks/install.js`; `--statusline` is an explicit option.

Local validation on macOS / Node 24.18.0: the focused and related suites passed **454 tests, with 1 Linux-only skip**, and all four deliberate CLI-gate mutations were caught. The full suite was **9,343 passed / 19 failed / 41 skipped**; all 19 failures reproduced on the same main baseline (`7383e8b8`) in the same environment, so I am not treating the full suite as green.

I approved the new fork workflow run, and [Repository Asset Audit run 33936311089](https://github.com/rullerzhou-afk/clawd-on-desk/actions/runs/33936311089) completed successfully for this exact head. The new installer/WSL documentation step actually ran and passed all 134 tests (0 failures or skips); this workflow is focused coverage, not the full test suite or WSL device validation.

**Real Windows 11 + WSL2 validation is still outstanding.** We currently cannot reach our Windows test machine, so the local tests and Linux CI do not establish that a real WSL Claude session reaches Windows Clawd.

If you have time to try this updated head in your WSL environment, we'd appreciate checking the documented default installation and a new Claude session showing up in Windows Clawd. An incomplete-payload rejection can be checked in a disposable config directory. Please also note any Pair/connectivity warnings. There is no rush.

One remaining limitation is now documented: ordinary WSL state events discover ports 23333–23337, while Claude's HTTP PermissionRequest URL is fixed at installation time (normally 23333 in a separate WSL home). State updates can therefore work while permission bubbles are unreachable if Windows Clawd uses another port. That transport behavior is unchanged by this PR.

This PR remains open for review and device verification; I have not merged it.

rullerzhou-afk:
**Maintainer device validation — PASS for this PR's scope**

Validated exact head `8c4fda8f` on a Windows host with Ubuntu WSL2, with Clawd launched directly from the PR checkout.

- Started with no existing WSL Claude configuration, copied all 73 top-level `hooks/*.js` files from the exact checkout, and ran `node ~/.claude/hooks/install.js`.
- The installer completed successfully and registered the 11 core command hooks plus the HTTP `PermissionRequest` hook. The statusline remained uninstalled as expected. Version-gated hooks were skipped because no Claude CLI version was available.
- Exercised the installed WSL hook with controlled `SessionStart`, `UserPromptSubmit`, `PreToolUse`, `PostToolUse`, and `Stop` events. The exact-head Windows app received every event and recorded the expected state transitions.
- Confirmed that a controlled WSL permission request reached the app and that aborting the request cleaned up the pending state without leaving a stale permission bubble.
- Ran `node --test test/install-dependency-cli.test.js` with Linux Node inside WSL: **14 passed, 0 failed**. This includes rejection of the old incomplete 13-file payload and successful installation from the complete payload.
- As an additional transport check, native Linux Codex CLI 0.153.4 completed a real ephemeral WSL turn, and Clawd received its official `SessionStart`, `UserPromptSubmit`, and `Stop` hook events. Codex's expected hook-trust gate was handled for this automated smoke with a single-invocation trust bypass against the inspected exact-head hook; no persistent trust setting was changed.

I did not run an authenticated Claude model session because this distro did not already have an authenticated Claude CLI. The Claude installer and hook lifecycle were exercised directly, while the real Codex turn independently confirmed the live WSL-to-Windows transport. Therefore, this validates the installer/dependency and transport scope of this PR without claiming an authenticated Claude model end-to-end run.

I consider the outstanding real-device validation for this PR complete.

rullerzhou-afk:
Thank you @sunnyswag for finding and fixing this. The key contribution was not only spotting the stale WSL file list, but tracing the silent failure mode: the installer reported success while the asynchronous hooks later failed on missing dependencies.

Your report and original commits gave us a clear root cause and the right direction. The maintainer follow-ups added the fail-fast preflight and broader regression coverage, and today's WSL2 device validation passed on the exact head.

Merging this now. Thanks again for the careful investigation and contribution!

PR #910, opened 2026-08-20
feat(Flick): aim-and-flick the pet across the desk

danonep2:
## Flick — aim and flick the pet across the desk

peteleco (1)

Hold **Ctrl** (**Option** on macOS) and drag **away** from where the pet should go. The pet holds still — it is the cue ball, not the cue — a white dashed projection shows exactly where the shot lands, and releasing launches it there.

Off by default. While off, the modifier drag keeps its old behavior, so nothing existing changes for anyone who does not turn it on.

**Settings → General → Behavior & position → Flick**, with an intensity slider (1–100) that only appears while the feature is on.

### The gesture

| Platform | Modifier |
| --- | --- |
| Windows / Linux | `Ctrl` + drag |
| macOS | `Option` + drag |

macOS uses Option because `Ctrl`-click is the system right-click gesture there — Chromium synthesizes a `contextmenu` from it, which would pop a menu over the projection. `Cmd`-click is already the Dashboard shortcut.

A modifier **click** that never moves stays a modifier click: on Windows/Linux `Ctrl`-click keeps opening the Dashboard. Only a real pull (> 4 px) consumes the gesture.

### Design decisions worth reviewing

**One geometry module, shared.** `peteleco-geometry.js` is pure (no Electron) and is used by both the main-process runtime and the overlay renderer, so the line the user sees and the position the pet flies to are computed once and cannot disagree. `to` is resolved **after** the clamp runs, so aiming into an edge visibly shortens the line instead of promising travel the pet cannot make.

**The projection starts on the avatar, not on the window.** The pet window is a rectangle sized for the widest pose a theme has, and the sprite is not centred in it. The mapping already existed and is the one the accessory system uses: `layout.centerX` (identical to each accessory `staticFrame.cx` in every built-in theme) plus the visible `contentBox` for Y, resolved through a new `getViewBoxPointScreen` — the exact inverse of the existing `getAssetPointerPayload`. Measured on clawd: the window centre is 32 px **above** the sprite's middle at 200×200, i.e. 16% of the window height, which put the line's origin in the upper quarter of the character.

**A flick can cross monitors.** The clamp resolves the work area from the target's own position, exactly as a drag-end does, so a hard shot near a seam lands on the neighbour. The overlay unions both displays to keep the line drawable — **except** when their scale factors differ, where one window cannot render both halves at the right size; there the line is clipped at the seam while the pet still crosses.

**Intensity caps reach, not the pull.** 1–100 maps to ~60–520 px. Dragging the cursor across the desk cannot throw the pet further than the cap; it just means "full power". That cap is also what keeps the projection short.

**Motion is deliberately unhurried.** 0.7 px/ms with an ease-out quartic, so the shot spends its back half bleeding off the last few pixels and the pet settles rather than stops. The projection dissolves over the launch (320 ms) instead of popping out at mouse-up; a cancelled aim clears it immediately.

**Runtime integration mirrors `roam.js`,** which solves the same problems:

- The flick joins the reconcile protection period (`isPetelecoAnimating`), so `pet-window-runtime` never reconciles against its own in-flight writes. Every exit path — including the error exits — releases that protection.
- Roam stands down for the **whole** gesture, aim *and* flight. The flight does not hold the drag lock, so it needed its own gate (`isPetelecoActive`).
- Frames carry the size captured at launch (the #569 ratchet: re-reading live bounds per frame lets mixed-DPI Windows round the pet larger mid-flight).
- The gesture's drag lock is handed back in the `peteleco:aim-end` handler, **before** the shot fires, so the launch's per-frame "did someone grab the pet?" gate never has to reason about IPC ordering.

**Landing does the drag-end finalization but deliberately not the mini-mode snap.** Shots are clamped to the work area and routinely land exactly on an edge, so snapping there would turn most hard flicks into an accidental mini mode.

**The overlay window is kept warm** — building one mid-gesture costs the first frame of the projection — and reclaimed under `lowPowerIdleMode` on the same 30 s clock the HUD/Orbit windows use.

### Naming

`peteleco` is the internal name (module files, pref keys, docs). The UI is localized: **Flick** (en), **Peteleco** (pt-BR), **Papirotazo** (es), **弹指** / **彈指** (zh / zh-TW), **튕기기** (ko), **デコピン** (ja).

Each localized README uses its own locale's UI name rather than the internal one — a README saying "Peteleco" while the panel says "Flick" would be worse than not documenting it.

### Files

| File | Role |
| --- | --- |
| `src/peteleco-geometry.js` | All the math. Pure, no Electron. |
| `src/peteleco.js` | Aim lifecycle and the launch animation. |
| `src/peteleco-overlay-window.js` | The transparent, click-through projection window. |
| `src/peteleco-overlay.html` / `-renderer.js` | The projection — a pure view, it decides nothing. |
| `src/hit-renderer.js` | Gesture detection (which modifier; shot vs. click). |
| `src/pet-geometry-main.js` | `getPetVisualCenter`. |
| `docs/guides/peteleco.md` | User-facing contract. |
| `README.md` + 5 localized | One bullet in the feature list, beside click reactions / drag / mini mode, plus the multi-monitor paragraph — which said only that Clawd can be *dragged* across displays. |

New prefs: `petelecoEnabled` (bool, default `false`) and `petelecoIntensity` (int 1–100, default 50).

### Testing

**8483 passing, 2 failing** across the suite; both failures are pre-existing on `main` and unrelated (Codex hook installer, Windows dispatcher on non-ASCII paths).

New coverage: 5 dedicated files (geometry, runtime, overlay window, overlay renderer, wiring) plus extensions to `hit-renderer`, `roam`, `prefs`, `settings-actions`, `pet-geometry-main`, `pet-interaction-ipc`, `pet-window-runtime` and the settings-renderer harness. Notable pins:

- `getPetVisualCenter` lands on `(100, 132)` against the real `themes/clawd/theme.json` — breaks if a theme layout change moves the anchor off the sprite.
- The clamp must receive exactly **four** arguments; a fifth was what pinned the shot to one display.
- The overlay's CSS fade duration and the window's hide timer must be the same constant.
- The overlay reclaim uses the same constant as `session-hud.js`.

The two workflows that fire on this PR were reproduced locally: `repository-asset-audit`'s exact 14-file test set (225 passing) plus `npm run audit:assets` (0 errors; its one warning, the tracked-tree size budget, is pre-existing — the tree is 2.03 MiB over and this diff is ~145 KB of text). The five README-reading tests, including the `LOCALIZED_READMES` enumeration in `test/icon-canonicalization.test.js`, also pass (194).

Three notes for reviewers:

- Touching `.gitignore` (to allowlist the new guide, since `docs/**` is ignored) and `src/main.js` puts this PR inside the `telegram-retirement-package-audit` path filter, so it runs the five-target electron-builder matrix. That is expected, not a stray trigger — but it is the one job that could not be reproduced locally.
- `test/mobile-preview-server.test.js` hangs while Clawd itself is running — the app holds the preview port. Pre-existing, unrelated to this branch, but it will stall a local full-suite run.
- The mixed-DPI overlay path is guarded by scale-factor comparison rather than measured; it could not be exercised on a uniform-scaling desk.

### Deliberately out of scope

Recorded so they are not mistaken for oversights: no edge **bounce** (the shot clamps and stops — the biggest remaining gap versus the billiards metaphor, and worth its own change since it needs a segmented path); no sound on landing; no keyboard equivalent; the gesture is not surfaced in the tutorial, so it is currently undiscoverable by design of being off by default.


rullerzhou-afk:
Hi @danonep2, thank you for this, and I'm sorry it took us this long to give you a decision.

Flick is a fun idea. Little interactions like this are part of what makes a desktop pet feel alive, and I appreciate that you took it beyond a demo with tests, translations, documentation, and a careful explanation of the tradeoffs.

After thinking it through, we've decided to close this PR without merging it. Clawd has accumulated a lot of interacting features, and we need to be more selective about what we commit to maintaining. Supporting aiming and flight alongside dragging, roaming, mini mode, and different display setups is more ongoing responsibility than we're ready to take on for this interaction, even with it disabled by default.

This is a product scope decision. I don't want to ask you to spend more time fixing issues or adding tests when that wouldn't change our decision to keep this feature out of the main app. We should have made that call sooner.

I know this is disappointing after the work you've put in. Thank you for spending your time making Clawd more playful, and for sharing both the idea and the implementation with us.

PR #923, opened 2026-08-23
fix: open tutorial language menu downward

YOIMIYA66:
## Summary
- Make the first-run tutorial language menu open downward in the default window layout.
- Widen the tutorial-only picker so `Português (Brasil)` stays on one line.
- Preserve the shared picker's adaptive upward fallback for constrained windows and enlarged text.

## Problem context
The welcome screen centers its content inside the scrollable tutorial body and lets the shared language picker choose the side with more usable space. After the seventh locale was added, the 128px-wide tutorial menu wrapped `Português (Brasil)`, increasing the menu height. Even after the existing welcome-content offset was increased, the default Windows tutorial window still had slightly less room below the trigger than the menu needed, so the picker opened upward.

## What changed
- Increase the welcome step's bottom padding so the centered content sits 24px higher in the default layout.
- Set a tutorial-only 160px picker width and prevent option labels from wrapping.
- Update the default seven-language geometry regression to require downward placement.
- Keep minimum-window/enlarged-text coverage that requires upward placement and scrolling when space is genuinely constrained.

## Behavior after this PR
- In the default first-run tutorial window, all seven language options open below the trigger without overlapping the footer.
- The Portuguese label remains on one line.
- Small windows and 150%/160% text-scale layouts continue to use the shared adaptive placement and bounded scrolling behavior.
- Settings language-picker sizing and placement are unchanged.

## Validation
- `node --test test/tutorial.test.js` — 17 passed.
- `node --test --test-name-pattern="language picker|tutorial picker" test/settings-renderer-browser-env.test.js` — 9 passed.
- Windows hands-on QA — the user confirmed with a screenshot that the default tutorial window opens the complete menu downward, keeps `Português (Brasil)` on one line, and leaves clearance above the footer.
- `npm test` — did not start the suite because the current 422-file runner exceeds the Windows process command-length limit: `spawnSync D:\Program Files\nodejs\node.exe ENAMETOOLONG`.
- Batched full-suite follow-up — unrelated baseline/environment failures remain: Feishu SDK `1.73.0` versus a test expecting `1.71.1`; `minimatch is not a function` in asset/icon tests; and 13 Slack-related tests cancelled after the event loop resolved. The tutorial-specific tests above pass.

## Platform note
Automated checks and hands-on QA ran on Windows with Electron 41.10.6. macOS and Linux manual QA remain unperformed; the shared constrained-layout fallback retains automated coverage.

## Scope control
- This changes only the first-run tutorial layout and its regression coverage.
- It does not change the shared language-picker placement algorithm or the Settings UI.

### DIFF
--- a/src/tutorial.html
+++ b/src/tutorial.html
@@ -112,7 +112,8 @@ body { user-select: none; }
   display: flex;
   align-items: safe center;
   justify-content: center;
-  padding: 18px 26px 202px;
+  /* Keep the full seven-row menu below three-line pt-BR/es copy at the default 720x700 size. */
+  padding: 18px 26px 234px;
   scrollbar-width: none;
 }
 .body.step-welcome::-webkit-scrollbar { width: 0; height: 0; }
--- a/src/tutorial.html
+++ b/src/tutorial.html
@@ -112,7 +112,7 @@ body { user-select: none; }
   display: flex;
   align-items: safe center;
   justify-content: center;
-  padding: 18px 26px 154px;
+  padding: 18px 26px 202px;
   scrollbar-width: none;
 }
 .body.step-welcome::-webkit-scrollbar { width: 0; height: 0; }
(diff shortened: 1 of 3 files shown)


rullerzhou-afk:
Thank you for spotting this and for the focused fix! We reproduced a remaining edge case where the pt-BR and es welcome copy spans three lines in the default 720x700 window, then added a small follow-up on this branch so all seven current languages open downward there while the existing upward fallback still handles smaller windows and enlarged text.

We took care of the follow-up ourselves to avoid sending you through another round after our delayed review. Your original commit remains intact and credited. We validated the updated layout on macOS Electron and will merge once the new CI run is green. Thanks again for the contribution!

PR #924, opened 2026-08-23
fix: stabilize theme customization menus

YOIMIYA66:
## Summary
- Open theme customization at the top, then restore the theme-list scroll position and originating Customize-button focus on return.
- Keep theme customization menus out of the outer Settings scroll geometry and inside a stable viewport boundary.
- Build on `main`'s existing stable scrollbar gutter and closed-menu unmount lifecycle; this PR does not reimplement or claim those fixes.

## Problem context
During Windows QA, the theme customization view inherited the theme-list scroll position. The accessory picker could also extend the outer content geometry and collide with the window edge; repeated open/close cycles then made the outer scrollbar and menu direction appear unstable.

Current `main` already reserves a stable scrollbar gutter and removes closed picker menus from layout. The remaining problem was specific to theme customization navigation and to open theme menus participating in the outer scrolling layout.

## What changed
- Added one opt-in shared-picker option, `viewportPlacement: "up" | "down"`; callers that omit it retain the existing anchored/automatic behavior.
- Constrained viewport menus to the Settings content boundary with a 12 px inset and a 240 px maximum height; long menus scroll internally.
- Close viewport menus when the outer Settings content scrolls and recalculate their position after window resize.
- When scroll closes a menu, restore the trigger only if focus is still inside that menu; focus already moved elsewhere is left untouched.
- Prefer downward placement for pet color and upward placement for accessories, with safe fallback when the preferred side lacks usable space.
- Save and restore theme-list scroll position across customization navigation, including the asynchronous activate-then-customize path.
- Move focus to the Back button on entry and restore it to the originating Customize button without scrolling.
- Reuse the Settings core focus-target helper instead of maintaining a second DOM traversal and `preventScroll` fallback in the Theme tab.
- Added regression coverage for repeated open/close cycles, boundary clearance, resize reflow, 100%–160% text scaling, direction fallback, scrolling, and focus restoration.

## Behavior after this PR
- Every theme customization entry starts at the top.
- Returning to the theme list restores the previous location and keyboard focus.
- Theme menus do not add to the outer content height; accessories prefer upward placement and pet color prefers downward placement.
- Small windows fall back to the safer side, and oversized menus scroll internally without covering the Settings edge.
- Escape, outside click, arrow-key navigation, ARIA state, close animation, reduced-motion behavior, picker persistence, and existing non-theme picker defaults remain unchanged.
- Content-width stability and closed-menu unmounting continue to come from the current `main` baseline.

## Validation
- Rebased onto current `main@e5855877` before the fresh checks below.
- `node --test --test-name-pattern='viewport-fixed|preferred theme-menu|outer overflow|inactive capable pet|active pet detail|returns to the theme list|opens customization at the top' test/settings-renderer-browser-env.test.js` — 5 passed, 0 failed.
- `node --test test/settings-renderer-browser-env.test.js` — 334 passed, 0 failed, including scroll-close focus ownership and `preventScroll` coverage.
- `git diff --check` — passed, with only the repository's existing LF/CRLF warnings.
- Remote CI for head `39c97ddb`: `unit`, `audit`, and all five platform packaging jobs passed.
- `npm test` — blocked before test execution by Windows `spawnSync ... ENAMETOOLONG`; no test failure was produced.
- Earlier Windows manual QA passed the customization and menu-placement flows. The new conditional scroll-close focus restoration is covered by the 334-test Settings suite but still needs a focused click-level smoke before Ready.

## Platform note
Automated validation ran on Windows. The prior click-level QA ran at 125% device scale. Other platforms remain covered by the shared browser-environment tests but were not manually exercised.

## Scope control
- No preference schema, theme data, Settings store/controller, IPC contract, or business behavior changed.
- Existing anchored picker callers retain automatic placement.
- Stable scrollbar gutter and closed-menu unmounting are baseline dependencies, not new production changes in this PR.
- This PR does not modify the Settings disclosure-animation work.

Related to #857


rullerzhou-afk:
Thanks @YOIMIYA66 for improving the theme customization experience! The menu placement and scroll/focus restoration are welcome fixes, and thank you for including regression coverage.

We've completed review and Windows smoke testing, and the PR is now merged. Appreciate your contribution! 🙌

PR #930, opened 2026-08-24
fix: restore idle visual picker options

YOIMIYA66:
## Summary

- Restore the missing choices in Settings → Animation / Sound → Animation → Default idle animation.
- Replace the special-purpose picker state machine with the existing shared Settings select control.
- Preserve the existing `setIdleVisual` command, broadcast-driven synchronization, error toast, and failed-save rollback behavior.

## Problem context

The idle visual picker still toggled only `.open` after the shared picker CSS adopted a separate `.menu-mounted` lifecycle. The trigger and ARIA state changed, and all option nodes existed, but the menu remained `display: none` because it was never mounted.

The duplicated picker implementation also risked drifting further from shared keyboard, dismissal, animation, and cleanup behavior.

## Design provenance and maintainer decision

Related to #509. The default-idle feature was originally implemented in #679.

The maintainer direction recorded in #509 was to offer the theme-default idle visual plus supported idle variants while keeping task, permission, completion, and sleep states unchanged. The implementation in #679 therefore derives this picker from `states.idle + idleAnimations`; for the bundled Clawd theme that currently produces Theme default, Idle Look, Idle Bubble, and Idle Reading. `Free roam walk` is a separate `roam` state and is intentionally not a persistent idle-visual choice.

This PR does not redefine that product scope—it restores access to the choices established by #679. Please confirm during review whether the original #509 behavior should remain, or whether the product should be narrowed in a separate follow-up decision.

## What changed

- Build the idle visual control through the shared `buildSettingsSelect` helper.
- Delegate menu mounting, Escape and arrow-key handling, outside-click dismissal, pending state, failed-change rollback, reduced-motion behavior, and listener cleanup to the shared picker.
- Keep the mounted control adapter so idle-visual broadcasts continue to update the current selection in place.
- Add a browser-environment regression test proving the menu mounts when opened and leaves layout when closed.

## Behavior after this PR

- Opening “Theme default” displays every idle visual option provided by the active theme.
- Closing the picker removes the menu from layout instead of leaving hidden overflow behind.
- Successful selections continue to update through the existing Settings command and broadcast path.
- Failed selections retain the previous value and show the existing save-failure toast.

## Validation

- `node --test --test-name-pattern='idle visual picker|idleVisual-only|idle visual menu' test/settings-renderer-browser-env.test.js` — 4/4 passed.
- `node --test test/settings-renderer-browser-env.test.js` — 303/303 passed.
- `node --check src/settings-tab-anim-overrides.js` — passed.
- Windows click-level check — the menu opens and displays Theme default, Idle Look, Idle Bubble, and Idle Reading.
- `npm test` — could not start on this Windows checkout because `node test/run-tests.js` hit `spawnSync ... ENAMETOOLONG`; targeted Settings coverage above passed.

## Scope control

- This PR only changes the default idle visual picker and its regression coverage.
- It does not change theme data, preference schema, IPC, animation assets, or other Settings controls.

### DIFF
--- a/src/settings-tab-anim-overrides.js
+++ b/src/settings-tab-anim-overrides.js
@@ -953,152 +953,51 @@
     wrap.className = "anim-idle-visual-row";
     const row = document.createElement("div");
     row.className = "row";
-    row.innerHTML =
-      `<div class="row-text">` +
-        `<span class="row-label"></span>` +
-        `<span class="row-desc"></span>` +
-      `</div>` +
-      `<div class="row-control">` +
-        `<div class="language-picker">` +
-          `<button type="button" class="language-picker-trigger" aria-haspopup="listbox" aria-expanded="false">` +
-            `<span class="language-picker-value"></span>` +
-            `<span class="language-picker-chevron" aria-hidden="true"></span>` +
-          `</button>` +
-          `<div class="language-picker-menu" role="listbox" aria-hidden="true"></div>` +
-        `</div>` +
-      `</div>`;
-    row.querySelector(".row-label").textContent = t("animIdleVisualLabel");
-    row.querySelector(".row-desc").textContent = t("animIdleVisualDesc");
-    const picker = row.querySelector(".language-picker");
-    const trigger = row.querySelector(".language-picker-trigger");
-    const valueEl = row.querySelector(".language-picker-value");
-    const menu = row.querySelector(".language-picker-menu");
-    trigger.setAttribute("aria-label", t("animIdleVisualLabel"));
-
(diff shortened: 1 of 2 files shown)


rullerzhou-afk:
Thank you, @YOIMIYA66, for fixing this and for replacing the duplicated picker lifecycle with the shared Settings select.

I reviewed exact head `69119dca09fb8ecab4089c8b72da30feb7390102` and validated it in a real macOS Electron run with isolated user data. Mouse open/select/outside-dismissal, keyboard Home/End/arrows/Enter/Escape and focus return, runtime switching to the selected idle asset, and restart persistence all passed. CI is also green.

Much appreciated — this is ready to merge.

PR #931, opened 2026-08-24
refactor: unify settings language picker

YOIMIYA66:
## Summary

- Route the General-page language picker through the shared Settings select builder.
- Remove its dedicated mounted-control lifecycle in favor of the common Settings picker collection.
- Preserve the language control's existing compact width and save/rollback behavior.

## Problem context

Settings dropdowns share the same low-level picker engine, but the language row was the remaining Settings caller that invoked `createLanguagePicker()` directly. That bypassed Settings-specific defaults and required a separate `mountedControls.languagePicker` cleanup path.

This is a small follow-up to the picker consistency discussion around #930; it is independently based on the latest `main` and does not depend on #930.

## What changed

- Build the language dropdown with `helpers.buildSettingsSelect()`.
- Apply the shared pending lock so another language change cannot start while persistence is unresolved.
- Register and dispose the language picker through `mountedControls.settingsSelects`.
- Remove the obsolete language-only mounted-control field and cleanup branch.
- Add a `settings-language-select` layout class so the shared builder does not widen the existing 128px control.
- Update browser-environment coverage for shared registration, pending locking, cleanup, and width preservation.

## Behavior after this PR

- The language picker uses the same Settings construction and cleanup path as timeout, theme customization, Remote SSH, and other Settings selects.
- Mouse, keyboard, ARIA, menu placement, failed-save rollback, and outside-click behavior remain provided by the existing shared picker engine.
- While a language save is pending, further choices are ignored consistently with other Settings selects.
- The visible control retains its existing compact width.

## Validation

- `node --test --test-name-pattern='Settings language picker|language picker|language saves|shared picker' test/settings-renderer-browser-env.test.js` — 8/8 passed.
- `node --test test/settings-renderer-browser-env.test.js` — 302/302 passed.
- `node --check src/settings-tab-general.js` — passed.
- `node --check src/settings-ui-core.js` — passed.
- `npm test` — could not start on this Windows checkout because `node test/run-tests.js` hit `spawnSync ... ENAMETOOLONG`; targeted Settings coverage above passed.

Windows click-level QA passed for mouse selection, keyboard navigation, Escape/outside-click dismissal, page-switch cleanup, repeated open/close, language persistence, and compact-width preservation.

## Scope control

- This PR only consolidates the Settings language dropdown.
- The onboarding tutorial continues to use the shared low-level picker directly because it is outside Settings.
- Dashboard native selects and non-select controls are unchanged.

### DIFF
--- a/src/settings-tab-general.js
+++ b/src/settings-tab-general.js
@@ -89,7 +89,6 @@
   let helpers = null;
   let ops = null;
   let i18n = null;
-  const languagePickerApi = root.ClawdLanguagePicker || {};
 
   const LANGUAGE_OPTIONS = ["en", "zh", "zh-TW", "ko", "ja", "pt-BR", "es"];
   const ROAM_MOVEMENT_NATURAL = "natural";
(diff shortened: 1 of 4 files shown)


rullerzhou-afk:
Thanks @YOIMIYA66 for this cleanup and for including the Windows click QA! Moving the language picker onto the shared Settings select lifecycle is a nice simplification, while preserving the compact layout. I also verified the exact head against current main with the full test suite and an isolated Electron UI pass covering mouse and keyboard language switching. Merged in 970ad509.

PR #936, opened 2026-08-24
feat: show permission requests in one scrollable bubble

rullerzhou-afk:
## Summary

- replace the one-window-per-request permission stack with one shared permission surface containing an active request and a waiting queue
- keep long content inside a bounded, internally scrollable bubble, with a two-item preview and an expandable full queue
- preserve input drafts, IME ownership, close semantics, remote notification delivery, pet-hidden behavior, and native focus modes while requests arrive or switch
- update the runtime architecture docs and add shared-surface lifecycle, renderer, formatting, and regression coverage

This is a follow-up to #918 and the reporter's request to make long permission content easier to inspect.

## Validation

- `npm test`: 8,669 passed, 0 failed, 30 skipped
- focused final run: 40 passed, 0 failed (`permission-surface-queue` + Slack delivery acknowledgement)
- `git diff --check`
- `node --check src/permission.js`

### macOS real-device smoke

- one permission window displayed seven simultaneous requests
- adding a request updated the queue count without clearing the active Chinese draft
- the full waiting drawer expanded and scrolled independently while the input and decision controls remained fixed
- ordinary requests switched in place without resolving either request
- submitting an input request handed the shared surface over to the normal non-input window mode without a crash or duplicate window
- the submitted Chinese answer and all test decisions completed their HTTP round trips; temporary test requests were cleared and the original permission mode was restored

## Residual validation limits

- Windows real-device behavior for this follow-up has not yet been re-run
- Deskflow intercepted synthetic Chinese typing on the test Mac, so the native IME candidate popup itself was not independently verified; Chinese draft persistence and answer round-trip behavior were verified through the live input surface

### DIFF
--- a/docs/project/agent-runtime-architecture.md
+++ b/docs/project/agent-runtime-architecture.md
@@ -323,8 +323,12 @@ CodeBuddy 的 PermissionRequest HTTP 所有权只认严格的本机 managed URL�
 - DeepSeek Harness 普通 approval 进入独立 blocking adapter；人工 Allow/Deny 可用，但 auto-tools、unattended 与 per-session grant 全部 DEFER。`ask_user_question` 返回 204 交给 DSH 原生 provider
 - Codex 的 PermissionRequest 是 official command hook；hook 脚本挂起等待 `/permission`，再把 sanitized allow/deny JSON 写到 stdout
 - `POST /permission` 接收 `{ tool_name, tool_input, session_id, permission_suggestions }`；Codex 额外带 `turn_id`、`tool_input_description`、`tool_input_fingerprint`
-- 每个权限请求都会创建独立 `BrowserWindow`，多个 bubble 从右下向上堆叠
-- bubble 会通过 IPC `bubble-height` 回报真实高度，主进程据此重排
+- 本地权限 UI 是一个进程级共享 surface：只完整展示一个 active 请求，其他请求显示为可切换的等待摘要；当前 active 默认保持稳定，新的交互型请求可抢占被动通知，但不会自动打断另一个交互型请求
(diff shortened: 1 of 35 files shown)


YOIMIYA66 (changes requested):
The shared-surface direction looks solid overall, especially the active-entry identity/revision checks and the atomic native-mode handover. I found one lifecycle gap that I think should be fixed before merge: the serving surface and handover candidate both depend on a renderer `bubble-height` acknowledgement before the window is actually shown/committed, but there does not appear to be a watchdog for the case where `did-finish-load` succeeds and that ACK never arrives.

That leaves a failure mode where the blocking permission remains pending but no usable surface becomes visible. The handover path is particularly risky: the current renderer disables queue controls before sending `permission-select`; if the hidden candidate loads but never reports height, the old surface remains in place while the handover never commits or fails, so interaction recovery can be stranded indefinitely.

Please add an ACK timeout/watchdog for both cases:

- initial/serving surface: arm after publishing the first `permission-show`; clear it only after a matching `surfaceRevision`/`activeEntryId` height report. On timeout, route through the existing renderer/surface failure fallback so blocking requests cannot remain invisibly parked.
- native-mode handover candidate: arm after publishing to the candidate; clear on the matching ACK. On timeout, destroy the candidate, cancel `permissionSurface.handover`, republish the old active surface with `restoreInteractionControls: true`, and keep the pending requests intact.
- make sure timers are cleared on destroy, renderer failure, successful handover, surface replacement, and cleanup.

I would also add regression coverage for (1) `did-finish-load` with no subsequent height report on the initial surface, and (2) a handover candidate that loads but never reports height. Those tests should prove that no permission can remain indefinitely pending behind an invisible/uncommitted surface.

rullerzhou-afk:
Closing this draft because its shared-surface product direction was superseded by #937 and #944. #937 kept each permission bubble compact with expandable, scrollable details, while #944 added a conservative overflow queue without replacing the per-request permission surfaces. The commit and remaining local experiments from this draft were not merged. Thanks @YOIMIYA66 for identifying the presentation-ACK lifecycle gap during review.

Issue #525, opened 2026-06-19
Windows pet visibility recovery and free-roam target flake


Git-creat7:
## Summary

This issue tracks two regressions found while rebasing the local Windows pet visibility fix onto the current upstream main.

### Problem

On Windows, the pet render window can become logically visible while not actually being shown on screen. The app keeps running, but the desktop pet disappears or fails to display again through the normal show/toggle path.

### Suspected cause

The transparent always-on-top render window can be cloaked by Windows DWM. The existing visibility path only checks Electron visibility state, so a cloaked window can be treated as visible even though it is not actually displayed.

### Expected behavior

If the pet is toggled visible or restored, it should recover even when the underlying Windows window has been cloaked.

### Local fix that worked

A local branch fixed this by:

- keeping the render window focusable on Windows instead of calling `setFocusable(false)` there;
- detecting DWM cloaking via `DwmGetWindowAttribute(..., DWMWA_CLOAKED, ...)`;
- treating cloaked windows as not actually visible;
- forcing `hide()` before `showInactive()` when recovering a cloaked render or hit window;
- re-syncing the hit window and reasserting topmost/HWND recovery when showing a hidden or cloaked pet.

### Verification

`node --test test/pet-window-runtime.test.js` passes with the local fix.

### Problem

`test/roam.test.js` can fail intermittently:

```text
✖ switches to roam visual state when animation starts
AssertionError [ERR_ASSERTION]: should call applyState('roam') when animation starts
```

### Cause

`pickRandomTarget()` currently picks one random target. If that target is closer than `ROAM_MIN_DIST`, it returns `null`. In that case no animation starts, so `applyState(roam)` is never called and the test can fail depending on random output.

### Expected behavior

Target picking should not randomly skip a roam when the work area has valid targets available.

### Local fix that worked

A local fix makes target selection deterministic enough by:

- retrying random target selection a bounded number of times;
- falling back to the farthest work-area corner if random attempts still pick too-close positions;
- returning `null` only when the work area truly cannot provide a target at least `ROAM_MIN_DIST` away.

### Verification

- `node --test test/roam.test.js` passes.
- `test/roam.test.js` passed 20 consecutive local runs.
- Full `npm.cmd test` passes outside the restricted sandbox on Windows.

## Environment

- Windows
- Clawd on Desk 0.10.0
- Upstream main around `191f6bd`

PR #942, opened 2026-08-25
feat(win): opt-in auto-hide of the pet while a fullscreen app is foreground

### LINKED ISSUE #935: [Feature Request] Auto-hide the pet when a fullscreen app is in the foreground
## Summary

Please add an option to automatically hide (or minimize to the edge / DND) the desktop pet when a real fullscreen application is in the foreground, and restore it when the app leaves fullscreen.

## Motivation

The pet is great for glancing at agent activity, but when watching a video in fullscreen or playing a game, the floating pet can block part of the screen. Right now there is no way to make the pet hide itself automatically in that situation — the only options are manual (drag to the edge for mini mode, DND, or context-menu hide).

## Current behavior (Windows)

Based on my reading of the code and release notes:

- Borderless-fullscreen apps (the common case, e.g. browser video, most games): the pet floats **on top** of the fullscreen app (overlay mode from #538 / #562, non-activating so clicks/drags do not kick the app out of fullscreen). This is the intended design — "stay visible over fullscreen apps".
- Exclusive-fullscreen apps: the pet stands down below the app (the #538 stand-down logic, kept as an escape hatch).
- v0.16.0 (#889) correctly distinguishes maximized normal windows from real fullscreen apps, so ordinary maximization no longer triggers the overlay policy.
- The old `fullscreenOverlay` toggle was removed from Settings because it could not reliably push the pet behind borderless-fullscreen games (a Windows limitation); the hidden pref remains as an escape hatch.

So the current design direction is "keep the pet visible over fullscreen apps", which is the opposite of what an auto-hide option would do.

## Desired behavior

An opt-in setting, e.g. `Settings → General → Auto-hide pet in fullscreen`:

- When a real fullscreen app is detected in the foreground (reusing the existing `src/win-fullscreen-detect.js` probe), hide the pet window (or automatically switch to mini/silent mode).
- Restore the pet automatically when the foreground app leaves fullscreen.
- Default off, so existing behavior is unchanged.

## Suggested implementation notes

- The detection infrastructure already exists (`src/win-fullscreen-detect.js`, `src/topmost-runtime.js` stand-down logic, #889 maximized-vs-fullscreen fix), so this would mainly be a new branch that hides the pet instead of floating it.
- Windows-first is fine (same scope as #562). macOS behavior may differ because the pet lives at screen-saver level there (see #27).

## Related

- #538 / #562 — Windows fullscreen overlay mode (current behavior)
- #889 — Windows fullscreen detection improvement
- #27 — macOS "stay visible over fullscreen apps" (opposite direction, already implemented)
- #460 — context-menu hide (manual only)

Thanks for considering!

KaiC5504:
Closes #935.

Adds an opt-in setting (default **off**, Windows-only row) that hides the pet — and the floating surfaces that follow it — while a real fullscreen app owns the foreground, and brings it back when the fullscreen session ends.

## How it works

- **A second, runtime-only visibility layer.** `pet-window-runtime` gains `fullscreenAutoHidden` next to the user-owned `petHidden`; effective hidden = manual OR auto. It is never persisted, so a crash while auto-hidden can't strand the pet invisible on relaunch, and the user's manual choice is never guessed at restore time: a pet you hid yourself stays hidden through any number of fullscreen episodes.
- **No new timers.** The sync rides the existing 1s focusable poll that already tracks fullscreen state for the topmost policy (#562), edge-triggered — enter fullscreen → hide within ~1s, leave → restore within ~1s. Off Windows the probe is constant-false, so everything is inert.
- **"Show Pet" mid-game is respected.** Choosing Show Pet (tray, context menu, or hotkey) while auto-hidden brings the pet back and binds an override to that fullscreen app's window identity — the probe now reports an opaque per-window id on positive verdicts (still truthy for all existing boolean consumers; degrades to plain `true` if `koffi.address` is unavailable). Alt-tabbing out and back never re-hides; the override ends when a *different* fullscreen app takes the foreground.
- **Menu items apply the intent their label carried.** The background restore can flip visibility while a tray/context menu is open (right-clicking the tray itself takes the game out of fullscreen), so Show/Hide Pet now captures its target at build time instead of toggling whatever the state happens to be at click time.

Interactions audited: the #525 cloak-recovery watchdog gates on effective hidden (no 5s hide/show strobe), hit-window input tracks effective hidden, `fullscreenOverlay` is moot while auto-hidden, and the mac hide-on-app bridge (#416) deliberately stays on manual state.

## Testing

- 28 new tests (`node --test`): the visibility layering, the poll sync incl. deferred-apply retries, the identity-bound override (alt-tab excursions, different-app episodes, decay), build-time menu intent, and the probe's id contract with degradation paths.
- Full suite on top of current main: **8791 tests, 0 fail**.
- Live-verified on Windows 11 under a real game (Valorant, exclusive fullscreen) and fullscreen video: hide ≤1s, restore ≤3s at the same position, Show Pet override, manual-hide preserved, pref-off mid-fullscreen restores.


rullerzhou-afk:
Hi @KaiC5504 — thank you for the substantial work behind #942. The core feature, visibility-layer design, HWND-bound override idea, menu-intent fix, Windows setting, and the original Windows validation all came from your PR.

Because the change touches several shared runtime surfaces (pet/hit visibility, topmost/focusability, permission and update surfaces, menus, settings, and the packaged Koffi path), we took over the final integration in a maintainer-owned branch instead of force-pushing your fork:

https://github.com/rullerzhou-afk/clawd-on-desk/tree/codex/pr-942-takeover

The maintainer pass kept the feature and added the remaining lifecycle hardening we found during adversarial review:

- kept permission/update/floating surfaces consistent with the fullscreen visibility layer;
- preserved suspended update prompts and visible-time autoclose accounting;
- made a manual Show survive arbitrarily long excursions from the same fullscreen HWND;
- restored topmost immediately on fullscreen exit without adding a second foreground probe in the poll tick;
- bounded stale remembered HWND state with optional `IsWindow(BigInt(id))` liveness, while failing open if that probe is unavailable;
- closed the wrapper-arity and cleanup seams;
- added packaged Windows x64 positive-path smoke coverage for `koffi.address`, distinct fullscreen identities, `IsWindow`, override lifecycle, re-hide on a different HWND, immediate restore/topmost, and dead-HWND cleanup.

Validation on the takeover head (`5252decb`): affected-module suite 928 pass / 3 platform skips / 0 fail; additional adversarial and bounded-model probes passed; Windows x64 package build and native inventory audit passed; packaged Electron runtime smoke passed; and we completed the single-monitor manual F11/maximize/long-excursion/new-episode/manual-hide matrix successfully. Your original Valorant exclusive-fullscreen result is valuable complementary coverage that we could not reproduce on our current machine. Physical multi-monitor behavior remains recorded as a non-blocking residual check.

No further action is required from you. We will carry the integration forward from the maintainer branch, while keeping #942 and your authorship explicitly credited in the final merge/release record. If you notice that the takeover changed an intended behavior, please call it out here and we will address it.

Thank you again — this was a large and thoughtful contribution.

rullerzhou-afk:
Merged via maintainer integration PR #973 as e585587741154849d1df973f29f1b9e79764e294 after the full unit/audit and five-target package matrix passed, including Windows x64 and Windows ARM64 packaged Koffi/HWND checks.

Closing this PR as superseded, not rejected. #942 remains the credited source of the feature, and the merge commit includes:

Co-authored-by: KaiC5504 <kaichuan2004@gmail.com>

Thank you again for the original implementation and the Valorant exclusive-fullscreen validation.

PR #943, opened 2026-08-25
feat: add Codex auto-start control

YOIMIYA66:
## Summary
- Add an independent **Start with Codex** switch to the Codex Agent settings card.
- Gate the existing Codex cold-start hook with the new preference plus the Codex installed/enabled state.
- Preserve the currently shipped behavior for upgrades by defaulting the new preference to enabled.

## Problem context
Codex already ships an auto-start hook and gate file, but Settings exposes no way to control that behavior independently. The existing gate followed only broad lifecycle synchronization and could not represent the user's intent separately from enabling the Codex integration.

Claude Code already has its own `autoStartWithClaude` preference. This PR gives Codex the same user-level control while retaining Codex's existing gate-file architecture, which is required because the hook runs before Clawd is available.

## What changed
- Add the top-level `autoStartWithCodex` boolean preference with a compatibility default of `true`.
- Add a shared effective-gate helper requiring all three conditions: Codex integration installed, Codex Agent enabled, and auto-start preference enabled.
- Add the localized switch to the Codex card in all seven Settings locales.
- Route preference updates through the Settings controller and the existing agent-integration mutation lock.
- Publish a disabled gate before fail-closed preference/agent mutations; publish an enabled gate only after the preference snapshot commits successfully.
- Treat explicitly malformed Codex auto-start or integration flags as non-authoritative during preference recovery.
- Subscribe the main process to both Agent state and the new preference so the disk gate stays synchronized.
- Add unit and browser-environment coverage for defaults, effective gating, failed persistence, controller locking, localization-backed rendering, and switch interaction.

## Behavior after this PR
- Existing users retain Codex auto-start behavior after upgrading unless they turn the new switch off.
- Turning the switch off prevents a Codex CLI session from launching Clawd, without disabling or uninstalling the Codex integration.
- Turning the switch on only authorizes cold launch when Codex remains installed and enabled.
- Disabling or uninstalling Codex always forces the effective gate off.
- Re-enabling Codex preserves the independently stored auto-start choice.
- A failed preference write cannot leave a newly enabled disk gate behind; disabling remains fail-closed.

## Validation
- `node --test test/agent-gate.test.js test/prefs.test.js test/settings-actions-system.test.js test/settings-actions.test.js test/settings-controller.test.js test/main-codex-auto-start-gate.test.js test/settings-renderer-browser-env.test.js` — 864 passed, 2 skipped, 0 failed.
- `node --test test/reasonix-hook-delivery.test.js` — 11 passed, 0 failed.
- `npm run verify:electron` — verified Electron 41.10.4.
- `git diff --check` and committed-diff checks passed.
- `npm test` reached 8,790 tests with 8,739 passed and 37 skipped, but the existing Windows Reasonix wrapper timing assertion exceeded its parallel-suite budget (5.48–6.04s); 13 later tests were cancelled by the parent failure. The same Reasonix file passed standalone as recorded above. No Codex auto-start test failed.

## Platform note
Automated verification ran on Windows. Cold-launch behavior and the Settings switch still need hands-on Electron/CLI testing before this Draft is ready for review.

## Scope control
- Does not change Claude Code auto-start behavior or its existing preference.
- Does not install or uninstall Codex hooks when the new switch changes.
- Does not change Codex permission handling, JSONL fallback monitoring, or Agent enable semantics.


rullerzhou-afk:
Thanks for the contribution! This has now been superseded by #959, which implements the Codex cold-start setting with migration and additional fail-closed hardening. Closing this draft as superseded.

PR #944, opened 2026-08-26
feat: add overflow queue for permission bubbles

rullerzhou-afk:
## Summary

- keep the existing independent request windows while the stack safely fits
- group overflow by agent and session, keep protected representatives visible, and move the rest into a navigation-only queue drawer
- preserve Ask and Plan DOM state, IME composition, petHidden cutoff semantics, Slack once delivery, and request-level lifecycle behavior
- pause free roaming while any local permission surface is visible and restart with the full idle delay

## Safety properties

- queue preload exposes navigation and presentation ACK only; it has no decision IPC
- request windows are hidden only after the current queue revision is acknowledged
- a missing ACK, load failure, renderer crash, or window failure falls back to the original request stack with zero decisions
- overflow mode disables global Allow and Deny shortcuts to prevent blind approval
- hidden request auto-close timers continue with the existing hook-hang timeout semantics

## Tests

- focused permission and floating-window suite: 446 pass, 0 fail
- full suite: 8720 tests, 8690 pass, 0 fail, 30 skip
- syntax checks: 17 JavaScript files
- git diff --check: pass

## Manual validation

- [ ] macOS: overflow grouping, drawer scroll/select, Ask and Plan input/IME, roaming hold, petHidden, multi-display
- [ ] Windows: 125% DPI, focus/topmost/taskbar, drawer scrolling, request switching

Related to #918.


rullerzhou-afk:
## Windows validation

Ran the Windows half of the manual checklist on `00a9534d`.

**Environment**: Windows 11 Pro for Workstations 26200, 3440x1440 physical / 2752x1152 logical at 125% DPI (`AppliedDPI=120`), work area height 1104. Development build launched via `node launch.js`.

| Item | Result | Evidence |
|---|---|---|
| 125% DPI | pass | 2 representatives + launcher = 814px ≤ 1104px; shrink calculation correct |
| topmost | pass | all five windows carry `WS_EX_TOPMOST` (verified via `GetWindowLong`) |
| drawer scrolling | pass | manual |
| request switching | pass | manual |
| no decisions inside the drawer | pass | manual - navigation only, blind-approval guard holds |
| reconcile stability | pass | zero reconcile failures across the whole session |
| taskbar | not objectively verified | Electron uses `ITaskbarList::DeleteTab`, which does not surface in window styles |

Grouping behaves as designed and the representative count is computed, not fixed:

- 6 requests in 1 session → 1 representative + launcher
- 9 requests across 3 sessions → 2 representatives + launcher

The third representative is correctly dropped because `canFit` rejects it: `335+374+335 + gaps + launcher 77 + margins = 1155 > 1104`. Two fit at 814.

## P0 found and fixed during this pass

**Symptom**: with 6 concurrent requests the queue never engaged at all - all 6 bubbles stacked past the bottom of the screen, exactly the behaviour this PR exists to fix.

**Root cause**: `createQueueWindow` creates the queue window with `show: false`, and the main process waits for the renderer's presentation ACK before showing it. The renderer sent that ACK from inside `requestAnimationFrame`. Chromium suspends rAF for a window that has never been shown, so the callback never ran, the ACK was never sent, and after ~1.5s the deadline fired:

```
permission queue unavailable; falling back to request stack: presentation ACK timeout
```

A closed loop: not shown → no rAF → no ACK → never shown.

The fallback itself worked exactly as the PR body promises - it reverted to the original request stack with zero decisions lost - so the failure was silent, with no error surfaced to the user.

**Why the unit suite did not catch it**: `test/permission-overflow-queue.test.js` drives `FakeBrowserWindow`/`FakeWebContents` and injects the ACK by calling the handler directly. The real "window visibility → rAF → IPC" sequence is bypassed entirely, so all 8720 tests passed against a deadlocked feature.

**Fix**: `00a9534d` sends the ACK synchronously when `document.visibilityState === "hidden"`, keeping the frame boundary once visible so drawer measurement still observes rendered layout. Re-verified on the real machine: no ACK timeout, queue folds correctly.

Contrast worth recording: the permission bubble windows are also created with `show: false`, but they call `showInactive()` immediately afterwards - `show: false` there only avoids stealing focus. The queue window was the only one that waited for an ACK before showing, which is why it was the only one that deadlocked.

## Related finding

While validating on this branch I also found a defect in the merged #937 baseline: the Ask card's `N / M` progress indicator overlaps form content on Windows. Filed as #948. It is not caused by this PR.

## Not covered

- The entire macOS checklist in the PR body remains unverified.
- Taskbar behaviour needs a visual check.

PR #945, opened 2026-08-26
fix(qoder): surface native titles without blocking lifecycle events

brantshin:
Qoder sessions now display native custom or AI-generated titles in the Dashboard, HUD and mobile session output instead of falling back to the working directory. Native custom titles take precedence over AI titles, and Clawd user aliases remain authoritative.

## Behavior

- Read matching Qoder transcript title records incrementally and asynchronously after lifecycle acceptance, preserving title records in the middle of large transcripts without blocking Electron's main thread.
- Keep tool and permission events free of transcript scans. Explicit payload titles take a zero-I/O path and cannot be overwritten by older cached or unread records; later native title records can update them.
- Recover from interrupted reads and reject stale results after session end, disable, replacement or cleanup. Native transcript enrichment is limited to eligible local sessions.
- Preserve explicit titles across same-session restarts, sanitize display titles, and retain custom-title precedence and session isolation.

The original contribution and authorship are preserved in `d724891` and `cd12975`. Main `59322b7` is included, with maintainer follow-up fixes in `51ee35a` and `000d0eca`.

## Validation

Accepted head: `000d0eca3343d3584044477c15603b816a0e0def`.

- **656 relevant Node tests passed**, including asynchronous progress, explicit/native title ordering, interrupted reads, lifecycle cancellation, same-session restart, route/state/snapshot/alias/mobile propagation and adjacent integration contracts.
- **macOS and Windows native validation passed** for Qoder title records, Chinese renaming, resume-triggered refresh and session-end cleanup. The Qoder CLI 1.1.48 follow-up with GPT-5.1 through OpenRouter completed ten real replies and ten normal Stop events across both hosts, with correct rendered titles and final empty snapshots.
- **All seven PR checks passed** on this head: [repository asset audit](https://github.com/rullerzhou-afk/clawd-on-desk/actions/runs/34446228624), plus [unit and five-target package audits](https://github.com/rullerzhou-afk/clawd-on-desk/actions/runs/34446228752) for Windows x64/ARM64, macOS x64/ARM64 and Linux x64.

## Boundaries

AI titles written after the first Stop appear on the next accepted scan event; native `/rename` changes also refresh on the next such event. This does not promise immediate file-watcher updates. Native Qoder end-to-end sessions in installed packages and WSL were not tested; CI package checks are a separate validation layer.

A Windows Dashboard first-show blank window was observed in an earlier run and did not recur in the follow-up. Its cause is unresolved and remains tracked in #1005; this PR does not claim to fix it.


rullerzhou-afk (changes requested):
Thank you for identifying this gap and for putting together focused tests. We did a deeper maintainer review across the Qoder hook runtime, the /state -> state.js -> session snapshot pipeline, Dashboard/HUD/mobile consumers, the current Qoder installer, the official Qoder hook/transcript documentation, and the locally installed @qoder-ai/qodercli 1.1.9 package source.

The underlying premise is valid: Qoder 1.1.9 does contain custom-title and ai-title transcript records with sessionId/customTitle or sessionId/aiTitle fields, and surfacing the native Qoder title would be a useful improvement over the current cwd fallback. This is therefore a real feature gap, not a duplicate of the existing Qoder integration.

However, we found two blocking correctness/performance issues that need to be addressed before this can merge.

1. Oversized transcripts can roll a correct title back to an older title.

Once the file exceeds 16 MiB, readQoderSessionTitle() samples only the first and last 8 MiB and then treats those samples as one complete chronological sequence. A realistic sequence is:

- an old ai-title exists in the head;
- the user runs /rename while that record is in the omitted middle;
- more than 8 MiB of tool output is appended afterward;
- the next hook sees the old head title and no title in the tail.

The hook then posts the old AI title. Qoder is not a first-wins title agent in state.js, so this non-empty value overwrites the previously captured custom title and propagates the rollback to Dashboard, HUD, remote approval labels, and mobile consumers. The current oversized-file test places the newest title in the tail, so it does not cover this blind spot.

A sampled head value cannot be treated as the latest title when an unscanned gap exists. Please use a correctness-preserving strategy such as incremental offsets/metadata state, a proper chunked scan, or another monotonic source/version mechanism. Qoder 1.1.9 itself uses chunked transcript scanning and metadata maps rather than treating head/tail samples as complete truth. Please add a regression with old title in the head, newer rename or clear in the omitted middle, and no title in the tail.

2. The implementation synchronously scans up to 16 MiB on every Qoder hook event.

buildStateBody() calls readQoderSessionTitle() for every mapped event when no explicit title is present. The installer registers synchronous command hooks for SessionStart, UserPromptSubmit, Pre/PostToolUse, Stop, Notification, PermissionRequest/Denied, and SessionEnd. A normal successful tool call therefore performs up to 32 MiB of synchronous read/decode/split work across PreToolUse and PostToolUse alone; additional events repeat it.

This work also happens before the state POST checks whether Clawd is running. Because disabling an agent intentionally does not uninstall its hooks, Qoder can continue paying this cost while Clawd is closed or the integration is disabled. This is especially risky on WSL, network-backed workspaces, slower disks, or systems with aggressive filesystem scanning.

Please keep transcript enrichment out of the high-frequency synchronous path: limit it to justified low-frequency lifecycle events, use an incremental/cache-aware design, and ensure the explicit-payload fast path performs no transcript I/O. We also need a realistic large-transcript wall-time/heap test and coverage showing that Pre/Post/Permission/Notification do not repeatedly perform a large scan.

There are several additional contract gaps to resolve:

- The claimed clearing behavior is not end-to-end. An empty custom title with no AI fallback produces no session_title field; server-route-state.js drops empty titles and state.js deliberately preserves the existing sticky title. Clawd would therefore retain the old title. Qoder 1.1.9 also rejects a blank /rename, so the PR needs real versioned evidence for any clear record it intends to support. If clearing is a real upstream operation, it needs an explicit Qoder-scoped three-state protocol through the route and state layers rather than changing the global sticky-title contract.
- Session isolation is permissive rather than strict: records without sessionId are accepted for an expected session. Qoder 1.1.9 validates these private metadata shapes as requiring string sessionId and title fields. Please validate structural shape and add matching, mismatching, missing, and qoder:-prefixed session-id cases.
- The public Qoder transcript documentation does not currently list custom-title/ai-title as a stable contract. Please include sanitized real fixtures and the verified Qoder version for AI title and /rename behavior, including record ordering relative to UserPromptSubmit, Stop, and SessionEnd. Official references: https://docs.qoder.com/cli/hooks and https://docs.qoder.com/extensions/hooks
- The bounded-read test verifies only a favorable output, not the actual bytes read, the omitted-middle rollback, a line-boundary tail start, or malformed-but-valid JSON shapes.
- The state-body test stops at buildStateBody(). Please add route -> state -> snapshot coverage for set, rename, and any supported clear behavior, including alias precedence and mobile title output.
- This head is now substantially behind main. Current main added the Qoder recap permission boundary and related title/state changes. Please rebase, preserve those changes, and run the Qoder hook/installer/offline suites plus server-route-state, state, state-session-snapshot, mobile, and recap coverage. The PR currently has no CI check result, and the reported repository-wide test run did not complete.

In short: we agree with the product goal and confirmed that the native title records are real in Qoder 1.1.9, but the current head/tail algorithm is not monotonic and the all-event synchronous scan is too expensive for a command-hook hot path. Once those two blockers and the contract/test gaps above are addressed, we will be happy to review the revised approach. Thanks again for the careful initial investigation and contribution.

rullerzhou-afk (approved):
The earlier requested changes are addressed in accepted head `000d0eca3343d3584044477c15603b816a0e0def`.

The final implementation replaces incomplete head/tail sampling and synchronous scanning with asynchronous incremental enrichment after lifecycle acceptance. It preserves native custom-title precedence, explicit-title ordering, session isolation and aliases, recovers from partial I/O failures, and rejects stale reads across lifecycle changes. Same-session restart behavior was independently reproduced and corrected in the final follow-up.

Validation includes 656 relevant passing tests, real Qoder CLI sessions on macOS and Windows, and ten successful OpenRouter replies with ten native Stop events in the 1.1.48 follow-up. All seven checks for this exact head passed, including the five package targets. Current main `59322b7` is already included; the prospective merge tree matches the accepted local tree.

The PR description now reflects the final implementation and its validation boundaries. The previously observed Windows Dashboard blank window remains an unresolved, separately tracked issue (#1005); its absence in the follow-up is not presented as a fix. WSL and installed-package Qoder end-to-end sessions remain untested.

Approved for this Qoder title change. The original contributor commits and authorship are preserved.

rullerzhou-afk:
Thanks, @brantshin, for identifying this gap, investigating Qoder's native title records, and following up with the incremental parser. Your contribution brings useful session names into Clawd instead of the working-directory fallback.

This is now merged! We completed the maintainer follow-ups and native validation on macOS and Windows, and all seven PR checks passed. Your original commits and authorship are preserved.

Thank you for the contribution and for your patience while we worked through the review and cross-platform checks.

PR #946, opened 2026-08-26
fix(shortcuts): preserve the Control modifier on macOS

Yueh-H:
## Summary

- preserve the physical Control modifier when recording shortcuts on macOS
- keep CommandOrControl mapped to Command while allowing explicit Control accelerators
- render explicit Control as the macOS Control symbol and protect common single-Control shortcuts

## Problem

On macOS, shortcut recording ignored ctrlKey, and the parser treated Control as CommandOrControl. As a result, a physical Control shortcut could not be recorded or displayed distinctly from Command.

## Behavior after this change

- metaKey becomes CommandOrControl and displays as Command on macOS
- ctrlKey becomes explicit Control and displays as Control on macOS
- Windows and Linux event handling remains unchanged
- the existing Ctrl shorthand continues to canonicalize to CommandOrControl for compatibility
- no new shortcut actions are introduced

## Test plan

- node --test test/shortcut-actions.test.js (22 passed)
- node --test test/settings-actions-shortcuts.test.js (3 passed)
- npm run verify:electron (Electron 41.10.4 verified)
- npm test (8,672 passed, 0 failed, 30 skipped)
- git diff --check

## Verification boundary

The parser, keyboard-event conversion, formatter, normalization, reserved-shortcut behavior, and full automated suite were verified on macOS. A manual Electron Settings UI smoke test was not performed.

### DIFF
--- a/docs/project/theme-state-ui.md
+++ b/docs/project/theme-state-ui.md
@@ -66,7 +66,7 @@ Settings 是独立 `BrowserWindow`，采用 5 层结构：
 - `applyUpdate` 和 `applyBulk` 对同步/异步 pre-commit gate 同构
 - `hydrate()` 是唯一跳过 pre-commit gate 的入口；post-commit effects 由 router 订阅 store changes
 - 设置写入路径只有 `controller → store → subscribers`
-- `shortcuts` 使用 Electron accelerator token；prefs v15 → v16 会先把旧配置中的字面 `Control` 迁移为 `CommandOrControl`，再允许新录制的 macOS 原生 `⌃ Control` 保持为独立 token。Windows/Linux 冲突判断把两者视为同一个实体 Control 键
+- `shortcuts` 使用 Electron accelerator token；prefs v15 → v16 会先把旧配置中的字面 `Control` 迁移为 `CommandOrControl`，再允许新录制的 macOS 原生 `⌃ Control` 保持为独立 token。Windows/Linux 会在危险组合检查、加载去重、设置冲突和显示时把两者视为同一个实体 Control 键，macOS 则始终保持 ⌘ / ⌃ 独立
 - `prefs.load()` 返回 `locked && recovered` 表示文件字节从未成功读取：controller 会在 validator / command / 外部 effect 之前拒绝用户 mutation，agent runtime 的启动同步、monitor、state/permission ingress 与 session recovery 全部 fail closed；修复文件访问并重启后才恢复。可读的 future-version `locked && !recovered` 继续保持既有的当前进程内存可改、磁盘不覆盖语义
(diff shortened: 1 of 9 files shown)


rullerzhou-afk:
Hi @Yueh-H, thanks for the contribution!

Before we evaluate this change further, could you share the specific use case that led you to need macOS Control (`⌃`) as a distinct shortcut modifier? It would be helpful to know what keyboard and macOS modifier-key configuration you use, which shortcut you tried to set, and what happened with the current behavior.

Clawd already allows shortcuts to be customized under Settings → Shortcuts, so we’d like to understand what the existing customization flow does not cover for your setup.

We’ll keep the PR open for now while waiting for more context. Thanks!

Yueh-H:
Hello @rullerzhou-afk 

Thanks for asking. In my fork, I added a new action called “Focus session 1”. I wanted to bind the physical macOS Control+Shift+1 (⌃⇧1) to this action so that it would not conflict with my existing Command-based shortcuts.

I use a MacBook Pro with the Apple Internal Keyboard and the default macOS modifier-key configuration.

To reproduce the issue, I opened Settings → Shortcuts, clicked Change for an existing action, and pressed ⌃⇧1. Clawd displayed and saved it as ⇧1, dropping the physical Control modifier.

This is the gap in the existing customization flow: although shortcuts can be changed, I cannot successfully configure a shortcut that uses the physical macOS Control key.

With this PR, the same input is correctly stored as `Control+Shift+1` and displayed as ⌃⇧1, while Command-based shortcuts remain separate.

Because “Focus session 1” exists only in my fork, the same issue can be reproduced using any existing shortcut action, such as “Toggle pet visibility”.

rullerzhou-afk:
Thanks for the clarification — the use case makes sense.

Since maintainer edits are enabled, we added a follow-up commit on top of your change: `a01ff399`.

The follow-up closes two compatibility gaps:

- adds a prefs v15 → v16 migration so an existing literal `Control` binding keeps its previous `CommandOrControl` meaning, while newly recorded macOS native Control bindings remain explicit `Control` accelerators;
- treats `Control` and `CommandOrControl` as conflicting on Windows/Linux, where they resolve to the same physical key, while keeping them distinct on macOS.

It also adds coverage for migration, save/load persistence, Settings registration, and startup restoration. Verification completed with `npm run verify:electron`, the full test suite (`8,680 passed, 0 failed, 30 skipped`), and a macOS Electron runtime smoke confirming that native Control and Command shortcuts can be registered simultaneously. The runtime smoke did not include a physical Settings UI keypress.

We are sending the updated head through an additional review now. Thank you again for the contribution and the clear reproduction details.

PR #949, opened 2026-08-26
fix: keep Ask progress below overflowing form content

### LINKED ISSUE #948: Ask card progress indicator overlaps form content on Windows at 125% DPI
## Summary

On Windows at 125% DPI, the `N / M` progress indicator in an expanded Ask (`AskUserQuestion`) card overlaps the form content instead of sitting below it as a separator.

Found while validating #944 on top of `7ab6a240`, the merge commit of #937. #937 was merged with its `Windows/Linux GUI validation` item unchecked and recorded as residual risk; this is one of the things that risk was covering.

## Environment

- Windows 11 Pro for Workstations 26200
- 3440x1440 physical / 2752x1152 logical, **125% DPI** (`AppliedDPI=120`), work area height 1104
- Development build, `node launch.js`

## Observed

Two pages of the same Ask card, both wrong:

- **Single-select page**: `1 / 2` sits on the bottom border of the "Other" text input. Typing makes it worse as the textarea grows.
- **Multi-select page**: `2 / 2` sits on the option container's bottom border line, and the two options below it are drawn *outside* that border.

The initial position is already wrong before any typing. Typing only amplifies it.

## Expected

`.elicitation-progress` carries `border-top` and is the last child of the scroll container, so it is meant to sit below the content as a separator line — not to be painted over.

## Where it lives

- `src/bubble.html:23-27` — `#elicitationProgress` is the last child of `.detail-scroll`, after `.elicitation-form`
- `src/bubble.css:337-351` — `.card.elicitation-scrollable .elicitation-progress.visible` has `padding-top: 8px; border-top: 1px solid var(--card-border)`; the element itself has no absolute positioning

## Analysis

This is not a spacing problem. The container height is computed too small and the overflow is not clipped, so form content paints over the row the progress label occupies. Both symptoms share one root cause: the height budget is wrong from the start, and a growing textarea amplifies it.

## Why the macOS pass did not catch it (hypothesis, not verified)

#937 states it keeps expanded cards inside **CSS-zoomed** BrowserWindows. Windows 125% DPI goes through the CSS zoom path, which is a different height-computation path from the 135% text scale used in the macOS checks. The macOS item "Ask progress has 4px vertical CSS padding" verifies spacing, not the height budget under zoom. I have a strong suspicion here but no direct evidence — the zoom link needs to be confirmed before fixing.

## Not included

Linux GUI validation remains completely unverified. This report only covers Windows.

rullerzhou-afk:
## Summary

- keep the expanded Ask form at its intrinsic flex height so its descendants cannot paint through the progress row
- preserve the single outer `detail-scroll` overflow owner and the fixed action/footer region
- add a source contract plus a real hidden-Electron layout regression for long single-select and multi-select Ask cards at 125% scaling

## Root cause

#937 placed the existing elicitation form inside a fixed-height column flex scroller. The form still had `min-height: 0` and default flex shrinking, so Chromium shortened the form box while its visible descendants kept painting at their natural height. The progress row followed the shortened box and was painted underneath the overflowing options/text area.

## Validation

- `node --test test/bubble-elicitation-layout-electron.test.js` — 1 pass, 0 fail
- Ask/expanded-bubble focused group — 25 pass, 0 fail
- `git diff --check` — pass
- `npm test` — 8709 pass, 8 fail, 25 cancelled, 39 skip; the failures are in baseline-identical Windows environment paths: Mobile Preview Server startup hangs before opening a listener, and two Codex installer nested PowerShell launches return `status: null` (isolated rerun: 39/41 pass). No changed file participates in either failing path.

Closes #948

### DIFF
--- a/src/bubble.css
+++ b/src/bubble.css
@@ -316,9 +316,11 @@ body { padding: 6px; }
 /* ── Elicitation form ── */
 .elicitation-form {
   display: none;
+  /* The outer detail scroller owns overflow. Keep the form at its intrinsic
+     height so its children cannot paint through the following progress row. */
+  flex: 0 0 auto;
   flex-direction: column;
   gap: 10px;
-  min-height: 0;
 }
 .elicitation-form.visible { display: flex; }
 .card.elicitation-scrollable .elicitation-form {
--- a/test/bubble-elicitation-layout-electron.test.js
+++ b/test/bubble-elicitation-layout-electron.test.js
@@ -1,0 +1,42 @@
+"use strict";
+
+const test = require("node:test");
+const assert = require("node:assert/strict");
+const path = require("node:path");
+const { spawnSync } = require("node:child_process");
+
+function electronExecutable() {
+  try {
+    const value = require("electron");
+    return typeof value === "string" && value ? value : null;
+  } catch {
+    return null;
+  }
+}
+
+test("expanded Ask progress stays after overflowing form content", { timeout: 30_000 }, (t) => {
+  const executable = electronExecutable();
+  if (!executable) return t.skip("Electron executable is not installed");
+  if (process.platform === "linux" && !process.env.DISPLAY && !process.env.WAYLAND_DISPLAY) {
+    return t.skip("Electron bubble layout audit needs an X11/Wayland display (CI can use xvfb-run)");
+  }
+
+  const fixture = path.join(__dirname, "fixtures", "bubble-elicitation-layout-electron.js");
+  const env = { ...process.env };
+  delete env.ELECTRON_RUN_AS_NODE;
+  const args = ["--disable-gpu"];
+  if (process.platform === "win32") args.push("--force-device-scale-factor=1.25");
+  if (process.platform === "linux") args.push("--no-sandbox");
+  args.push(fixture);
+  const result = spawnSync(executable, args, {
+    env,
+    encoding: "utf8",
+    timeout: 25_000,
+  });
+
+  assert.strictEqual(
+    result.status,
+    0,
+    [result.stdout, result.stderr].filter(Boolean).join("\n") || `Electron exited ${result.status}`
+  );
+});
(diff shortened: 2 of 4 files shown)


rullerzhou-afk:
## Windows real-machine validation — passes

Verified on the machine that originally reported #948.

**Environment**: Windows 11 Pro for Workstations 26200, 3440x1440 physical / 2752x1152 logical at 125% DPI (`AppliedDPI=120`). Live desktop pet build, real `AskUserQuestion` card issued by Claude Code.

Both pages of a two-question Ask card, with option descriptions long enough to force overflow:

- **Single-select page**: `1 / 2` now sits fully below the "Other" text input. Previously it was painted on the input's bottom border.
- **Multi-select page**: `2 / 2` now sits below the entire option list, and no option paints outside the question card's border. Previously it landed between two options with the ones below drawn outside the border.

## Test-effectiveness check

I did not take the new test's green run at face value — a passing test proves nothing unless it can fail. Reverted `src/bubble.css` to the pre-fix state (restore `min-height: 0`, drop `flex: 0 0 auto`) and re-ran the Electron layout test:

```
single: progress overlaps form descendants
  progressTop:       439.10
  formContentBottom: 695.95
```

It fails immediately, and the numbers match the reported symptom exactly — the progress row sits 257px above the content it should follow. The test genuinely covers the defect.

Worth noting the guard in the fixture:

```js
assert.ok(layout.detailScrollHeight > layout.detailClientHeight + 1,
  `${kind}: fixture must exercise the outer overflow path`)
```

This asserts overflow actually occurred before checking placement, which prevents the fixture silently degrading into a no-op if the content ever stops being long enough. Good call.

## Interaction with #944

#944 is still open and touches the same bubble subsystem, so I checked the combination rather than assuming. No file overlap (#944 does not touch `bubble.css` or any elicitation file). Local trial merge of both branches: **458 bubble/permission tests, 455 pass, 0 fail, 3 skip**. The trial merge was local only and has been discarded.

Order does not matter for correctness, but whichever lands second should pick up the other — validating #944 on a branch without this fix will still show the overlap.

## Side finding

The fixture creates its window with `show: false` plus `backgroundThrottling: false` and reads layout from a double `requestAnimationFrame`. It completes in ~927ms with no timeout, which confirms `backgroundThrottling: false` keeps rAF running for a window that has never been shown.

That is directly relevant to the P0 fixed in #944 (`00a9534d`), where a never-shown window's suspended rAF deadlocked the queue's presentation ACK. So that class of deadlock has two viable fixes: acknowledge synchronously while hidden (what #944 does), or disable throttling on the window (what this fixture does). Recording it here so the next person hitting it has both options.

PR #954, opened 2026-08-27
fix: restore Codex Desktop focus for scoped sessions

draintovmasyan783-creator:
## Summary

- resolve Codex Desktop thread IDs from rawSessionId before falling back to id
- cover the profile-scoped runtime session shape in focus helper tests
- update the snapshot focus test to use a canonical scoped session key

## Why

Remote profile isolation introduced canonical session keys in #736 while retaining the original provider ID in rawSessionId. Codex Desktop focus still parsed entry.id as codex:<uuid>, so local scoped entries were marked unfocusable and the HUD/dashboard could not open their threads.

The fallback to id preserves compatibility with legacy unscoped entries. Existing local-only, originator, remote-host, platform, and malformed-input guards remain unchanged.

## Tests

- node --test test/session-focus.test.js test/session-focus-handoff.test.js test/state-session-snapshot.test.js test/session-ipc.test.js
- 86 passed, 0 failed

### DIFF
--- a/src/session-focus.js
+++ b/src/session-focus.js
@@ -16,7 +16,9 @@ function normalizeOsPlatform(options) {
 function getCodexThreadId(entry) {
   if (!entry || entry.agentId !== "codex") return null;
   if (!isCodexDesktopOriginator(entry.codexOriginator || entry.originator)) return null;
-  const match = normalizeString(entry.id).match(CODEX_THREAD_SESSION_ID_RE);
+  const sessionId =
+    normalizeString(entry.rawSessionId) || normalizeString(entry.id);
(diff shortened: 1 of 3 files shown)


rullerzhou-afk:
Thank you for the contribution and for restoring Codex Desktop focus for profile-scoped sessions.

I verified the exact PR head 20a024e on macOS: opening the session from the Clawd Dashboard switched Codex Desktop from a different task to the intended thread. I also tested the synthetic merge against the latest main; the full suite passed with 8,771 passed, 31 skipped, and 0 failed.

Merged as 5a332b0c. Thanks again!

PR #956, opened 2026-08-28
feat(shortcuts): focus visible sessions by index

Yueh-H:
## Summary

- add configurable, initially unassigned `Focus session 1` through `Focus session 9` global shortcut actions
- resolve each shortcut against the current ordered, visible, focusable HUD sessions at invocation time
- show matching `1` through `9` indices in the Session HUD while leaving unfocusable rows unnumbered
- keep the new actions out of the first-run tutorial shortcut summary

## Behavior

The mapping is dynamic rather than tied to a session ID. If HUD ordering changes, the next key press follows the new visible order. The actions are unassigned by default so the app does not reserve nine new global shortcuts without user intent.

This PR intentionally does not add or change an Open session dashboard shortcut.

## Testing

- `npm test` — 8,813 passed, 0 failed, 31 skipped (8,844 total)
- targeted shortcut, settings, focus, HUD, prefs, tutorial, and i18n tests — 542 passed
- `npm run verify:electron` — Electron 41.10.4 verified
- isolated macOS Electron QA — HUD rendered `1`, `2`, `3` for three focusable rows and left an unfocusable row blank; native `Control+Shift+9` registration succeeded


rullerzhou-afk:
Thank you for putting this together. The initially unassigned shortcuts, dynamic focusability checks, and the UI/runtime test coverage made this a useful exploration of a real gap: Clawd currently has no keyboard path for navigating directly to a session.

After reviewing the interaction more closely, though, we do not plan to merge the current `Focus session 1` through `Focus session 9` design as-is. This is a product-semantics concern rather than a code-quality concern.

Clawd's HUD and Dashboard ordering is activity-driven. As `updatedAt` changes, "session 1" can become a different session seconds later. That makes the number a live ranking rather than a stable address. If the user must first inspect the HUD to learn the current number, nine persistent global shortcut actions do not provide the stable, muscle-memory behavior their names imply. The HUD's folded `+N` rows can also leave valid shortcut targets without a visible number/title mapping.

Our current direction is narrower:

- add one configurable global action to open the Session Dashboard, initially unassigned;
- while the Dashboard is visible, support local keyboard selection with visible `1` through `9`, arrow keys, Enter, and Escape;
- keep the row identity and numeric mapping stable for that Dashboard presentation, even while status text continues to update;
- rebuild the ordering the next time the Dashboard opens;
- avoid permanent session slots, permanent HUD numbering, and nine global session shortcuts.

We are not planning to include a general "Next session" cycling mode in the initial scope. We may separately evaluate a semantic shortcut for jumping to a session that needs attention, but only after the completion/acknowledgement behavior is consistent across agents.

I could not find an existing issue specifically tracking this keyboard-navigation design, so we will track the narrower follow-up separately. Because this is a substantial product pivot, please do not spend time reworking this branch yet. We will keep this PR as an important reference and preserve credit for the original idea and exploration.

Thanks again for identifying the missing keyboard workflow and for taking the time to build and validate a concrete proposal.

rullerzhou-afk:
@Yueh-H Following up on our August 31 discussion: the keyboard-navigation follow-up has landed through #972 and is present on `main`.

The final interaction uses one configurable Quick Select shortcut, unassigned by default, to enter a temporary keyboard mode in the full Session Dashboard. Digits 1–9 select the displayed targets, with the numeric mapping frozen for that presentation while status information continues to update. It is available on macOS and Windows; Linux is not supported for this mode. Arrow/Enter selection and general session cycling are outside the implemented feature.

We are closing this PR as superseded by that follow-up. The original nine-global-shortcut and permanent HUD-numbering design was not adopted, so no rebase or further rework of this branch is needed.

Thank you for identifying the missing keyboard workflow and for building and testing a concrete proposal. Your original idea and exploration helped shape this direction, and this PR remains linked from #972 as the design reference. Apologies for the delay in closing the loop here.

PR #975, opened 2026-08-31
feat(hermes): Hermes Agent over Remote SSH (Phase 1)

### LINKED ISSUE #916: feat(hermes): extend Remote SSH deploy to Hermes Agent
## Background

I run Hermes Agent on a remote Linux server (systemd service, dedicated profile,
scheduled jobs) and drive it from a Windows machine. Clawd never sees those
sessions, and I could not find a supported way to install the Hermes plugin on a
remote host.

`docs/guides/guide-remote-ssh.md:14` lists the supported remote agents:

> "At least one supported remote agent is installed on the remote: Claude Code,
> Codex CLI, or Copilot CLI"

## The asymmetry

WSL deploy already handles Hermes, using the same shape SSH deploy uses:

| | `src/wsl-deploy.js` | `src/remote-ssh-deploy.js` |
|---|---|---|
| agents | codex, codebuddy, zcode, codewhale, **hermes**, qoder, reasonix, qoderwork | claude, codex, copilot |
| transport | `wsl-utils` stdin pipe | `buildScpArgs` (scp) |
| install step | copy `hermes-install.js` + `hermes-plugin/*`, then `node hermes-install.js --json` | scp `HOOK_FILES`, then `node ~/.claude/hooks/<agent>-install.js --remote` |

So the "copy the installer to the other side and run it there" pattern already
exists for Hermes — just not over SSH. `hermes-install.js` also already has a
machine-readable mode (`--json`) that WSL deploy drives.

Without it, the Hermes plugin only ever posts to `http://127.0.0.1:{23333..23337}/state`
on the machine it is installed on, which on a remote host is the server itself.

## What seems to be missing

1. `hooks/hermes-install.js` — a `--remote` mode (it already has a `--json` machine mode,
   but no remote handling at all, while `hooks/install.js` deals with remote mode throughout)
2. `src/remote-ssh-deploy.js` — ship `hermes-install.js` + `hermes-plugin/*` to the remote
   (via `HOOK_FILES` or a Hermes-specific staging dir, whichever fits) and add a Hermes
   install step alongside the Claude/Codex/Copilot ones
3. `docs/guides/guide-remote-ssh.md` — supported-agent list (lines 14 and 35)

I don't think `hooks/hermes-plugin/__init__.py` needs to change: remote-ssh-deploy
already sets up the forwarded port, so the plugin's hardcoded `127.0.0.1` should
resolve back to the local Clawd the same way the Claude hook does.

To be clear about what I am *not* proposing: `capabilities.subagent: false` in
`agents/hermes.js` is out of scope — `known-limitations.md` explains it is blocked
on upstream Hermes emitting paired subagent start/stop events.

## Questions before I open a PR

1. Was Hermes left out of Remote SSH deliberately? I checked `known-limitations.md`
   and did not see it listed there, unlike the Copilot/Telegram and Orca remote
   boundaries, so I may be missing context.
2. If it is welcome, would you prefer Hermes reuse the existing SSH deploy path, or
   is there a reason the WSL and SSH deploy paths are kept as separate
   implementations that I should respect?

Happy to do the work — I have the exact setup to test it end to end.

Environment: Windows 11, Clawd 0.15.0 from source, Hermes on a remote Linux server
(systemd, non-default profile).

xfurqan0:
Closes #916.

## Problem

Clawd already tracks Claude Code, Codex CLI and Copilot CLI over Remote SSH, but Hermes Agent
only works locally and in WSL (#916). A remote Hermes session has no state or permission surface
in Clawd, and the Hermes plugin's transport is not safe to expose over a tunnel as-is: it scans
local ports, trusts local runtime discovery, and sends process-tree metadata that has no meaning
(and leaks information) across hosts.

## Fix

Implements the plan agreed in #916, within the constraints you set there (including your answers
on stage cleanup, exit semantics, the local gate, and ordering/Repair).

**Python plugin secure transport** (`hooks/hermes-plugin/__init__.py`, v0.3.0)
- Re-reads `$HOME/.claude/hooks/clawd-remote.json` on every state/probe/permission request — no
  `(mtime,size)` cache; schema v2 validated field-for-field in parity with `hooks/server-config.js`
  (cross-language fixture test).
- Secure mode latches for the process lifetime via the `clawd-ssh-secure-v1` marker, the identity
  file, or `CLAWD_SSH_REMOTE=1`. Once latched, a missing or invalid identity fails closed: no
  five-port scan, no local runtime discovery, zero network I/O.
- Uses only the identity's `remotePort`; sends `x-clawd-routing-nonce` on every request (header
  only — never path/query, so tracebacks cannot leak it); the nonce is never logged and nonce keys
  are redacted. Identity is re-read between the permission probe and POST; a port drift between the
  two fails closed instead of re-probing silently.
- In secure mode no `agent_pid`, `source_pid`, `pid_chain`, `editor` or tmux fields are collected
  or sent; the process-meta resolver is never started; a valid `orca_pane_key` is kept after the
  latch. Prefers `hermes_constants.get_hermes_home()` when importable.

**Installer remote mode** (`hooks/hermes-install.js`)
- New `--remote --json` mode: absolute `--source-dir`, repeatable `--target-home` (root first,
  then profiles), `--cli-timeout-ms`; invalid arguments emit a single error sentinel and exit 1.
- Each target directory is classified immediately before mutation (`absent` / `managed` / `legacy`
  / `foreign` / `symlink`); foreign and symlinked directories fail closed with no bytes changed.
- Atomic copy → sha256 readback → CLI enable → config.yaml enable verification → only then the
  `clawd-ssh-secure-v1` marker (0600), read back after writing.
- Any per-target failure makes the aggregate result an error and the process exit non-zero — a
  default-home success plus a named-profile failure is a hard failure, per your Q2 answer.
- Result stays `schemaVersion: 1`; per-profile copy/hash, enable, marker and activation
  diagnostics are added additively and spread conditionally, so local/WSL results are unchanged
  (all 20 pre-existing installer tests pass untouched).
- Remote uninstall deletes exactly the managed leaves (`plugin.yaml`, `__init__.py`, marker,
  `__pycache__/*.pyc`) and removes directories with `rmdir`; ownership conflicts are warnings
  without mutation; no recursive deletion anywhere in remote mode.
- Remote-only CLI resolver candidate `<home>/../.local/bin/hermes` for git-based installs (your
  Raspberry Pi baseline); local/WSL candidate order untouched.

**Deploy phase** (`src/remote-ssh-deploy.js` + `src/hermes-installer-result.js`)
- Preflight discovers and freezes `hermesHome`/`hermesTargets` (additive JSON fields) with the
  same classification algorithm; ownership conflicts fail the deploy before any Hermes mutation.
- New steps `hermes-files` → `install-hermes` run inside `secureDeploy` under the existing
  serialized transport coordinator, deploy lease, fencing, ownership checks and unknown-result
  recovery — no Hermes-specific SSH deployer. Per your Q4 answer the phase is the final remote
  mutation of the deploy (after the identity-tracked components, immediately before the success
  return), and a test pins that ordering.
- Assets go to a separate sibling exact stage `<staging>/<leaseId>-hermes`; cleanup is
  `rm -f <two files> && rmdir <stage>` under the fence, on known results only. An unknown
  SSH/mutation result preserves the stage, deploy lock and recovery evidence — no follow-up
  cleanup mutation (your Q1 answer).
- `installHermes` is NOT added to `REMOTE_IDENTITY_STEP_NAMES` and the persisted identity
  transaction schema is unchanged (test-pinned). On failure the transaction stays uncommitted;
  Repair resumes the same `toNonce` and re-runs ownership, hash, marker, enable-entry and
  activation readback; a matching managed installation is a no-op (`unchanged`).
- Per-profile CLI timeout and the outer fenced-command timeout are budgeted by frozen target
  count (capped). `hermes plugins list --json` and `systemctl --user list-units 'hermes*'` are
  optional diagnostics only (warnings, never errors).
- After the deployment is fully verified, the local Hermes Agent switch is enabled through the
  same Settings controller command the Settings → Agents toggle uses (your Q3 answer);
  `integrationInstalled` is preserved exactly — a remote-only deploy never marks the local
  integration as installed.
- Uninstall runs `hermes-install.js --uninstall --remote --json` under the same lease/fence
  before the shipped installer is removed.
- No gateway restarts, ever. Progress reports "enabled — effective on next session" vs
  "module replaced — gateway restart required"; `systemctl` is a best-effort hint.

**Server-side defense in depth** (`src/server-route-state.js`, `src/server-route-permission.js`)
- Profile-bound remote requests have PID/process-tree/tmux fields stripped before they reach
  session/liveness state; local behaviour is unchanged.

**Shared module** (`src/hermes-installer-result.js`)
- `parseHermesInstallerResult` + constants moved verbatim from `src/wsl-deploy.js`, which
  re-exports them; WSL and SSH share only the result parser and asset list, per your constraint.

## Tests & verification

- Branch is merged with current `main` (no conflicts; the i18n key-order and
  server-route tests pass on the merged tree).
- Full `npm test` on the merged tree — Windows 11, Node v22.23.2 (`engines >=22.12.0`;
  `.nvmrc` 24.18 not available locally — upstream CI covers it), Python 3.13:
  **9437 tests, 9374 pass**; the 6 failures are all in files this PR does not touch and all
  trace to Windows symlink-creation permissions (agent-installation-detector fixtures and one
  upstream dsh-install-bridge fixture); the 13 cancellations are the pre-existing
  `slack-notify-client` event-loop issue.
- Files touched by this PR, on the merged tree — Windows: **509 tests, 500 pass, 0 failures**
  (9 POSIX-only skips).
- Same 11 test files on Linux (Ubuntu, Node v22.23.2, python3 3.12; run at `a04375a1`, before
  the main merge — every touched file re-verified on Windows after the merge): **403/404 pass,
  0 skips** —
  every POSIX/symlink-gated test (C6 symlink classification, preflight target freezing, lease
  fencing of multiline mutation blocks, secure managed command readback) executed and passed.
  The one non-running file (`server-route-permission.test.js`) needs the `electron`
  devDependency, which was not installed there; it passes on Windows with full dependencies.
- i18n: both new step keys present in all 7 locales at the same relative position
  (`test/i18n.test.js` order check green).
- Real host — deployed from the Electron app (Clawd 0.16.0 + this branch, Windows 11) to an
  Ubuntu server (kernel 6.8.0-138), Hermes Agent v0.20.5 git `b2c4f1f3` (git install, CLI in
  `~/.local/bin`), remote Node v22.23.2, two homes (`~/.hermes` + `~/.hermes/profiles/qarpus`),
  two live user-unit gateways:
  - **Deploy:** `~/.claude/hooks/` created 0700 with the full hook payload; both plugin dirs
    contained exactly `plugin.yaml`, `__init__.py`, `clawd-ssh-secure-v1` — sha256 of both
    assets byte-identical to the repo files, marker 19 bytes, no trailing newline, mode 0600;
    both `config.yaml` files gained the `plugins.enabled: [clawd-on-desk]` block (neither had a
    `plugins:` key before); the sibling Hermes stage was removed; **neither gateway was
    restarted** by the deploy (ActiveEnterTimestamps 10 days old, `NRestarts=0`); locally the
    Hermes switch was enabled automatically through the Settings controller while
    `integrationInstalled` stayed `false`.
  - **Activation:** after a manual `systemctl --user restart` of each gateway, plugin discovery
    went 55→56 found / 49→50 enabled; the capability sandbox denied `tools.override` (expected —
    the plugin only registers hooks); `__pycache__` appeared only after each gateway's own
    restart.
  - **Live session:** a WhatsApp turn of ~7 minutes with 30+ tool calls streamed
    SessionStart / UserPromptSubmit / PreToolUse / PostToolUse / Stop — plus PostToolUseFailure
    and StopFailure from a genuinely failing Hermes tool — through the profile-bound ingress.
    Every remote event carried no `agent_pid` / `source_pid` / process-tree fields while a
    concurrent local Claude Code session kept its PIDs; no port probing appeared in any server
    log (the secure latch pins requests to the single identity port).
  - **Repair:** rotated the identity in ~16 s (nonce, marker, host-prefix, hook settings, both
    `config.yaml` rewritten; plugin code bit-identical at its original mtimes — the managed
    no-op path); both gateways stayed up (`NRestarts=0`) and events kept flowing within 90
    seconds of the rotation, with no Hermes restart at any point.
  - **Revoke + redeploy:** revoking the remote identity made the loaded plugin fail closed
    immediately (event stream stopped; Clawd marked the session interrupted, then
    stale-deleted it); a fresh Deploy issued a third identity and events resumed on the next
    session. The uninstall path (exact-leaf deletion, `rmdir` chain, foreign-dir preservation)
    is covered by the unit suites; a live uninstall was not run on this host.

## Docs

- `docs/guides/guide-remote-ssh.md` + zh-CN: Hermes prerequisites (CLI locations including
  `~/.local/bin/hermes`), deploy flow step, Phase 1 scope ("account-default" layout, standard
  `~/.hermes` + `~/.hermes/profiles/*`; custom `HERMES_HOME`, multiplexed gateways,
  `profile-isolated` unsupported/unverified), activation semantics and the no-gateway-restart
  rule, automatic local enable after a verified deploy, troubleshooting checklist.
- `docs/guides/known-limitations.md` + zh-CN: new "Hermes Agent over Remote SSH (Phase 1)" row.
- `docs/guides/setup-guide.md` + zh-CN: pointer from the Hermes section to the Remote SSH guide.
- `README.md`, `AGENTS.md` constraint line, `docs/project/agent-runtime-architecture.md`.

---


rullerzhou-afk:
Thank you @xfurqan0 for bringing Hermes Agent to Remote SSH, and for the detailed implementation and testing notes!

We’ve now validated the integration with real Hermes sessions on a Raspberry Pi 5 and Windows Clawd. State updates reach the desktop, and with the follow-up fix in #1003, the opt-in terminal permission flow passes manual Approve/Deny tests—including another successful Deny test after restarting Clawd.

Your contribution is merged and working in our live setup. Thanks for all the work you put into this!

xfurqan0:
Thank you for the thorough validation, and for catching the callback bug in #1003 — that one was on me: I tested the permission flow through the state events but never through the registered pre-tool callback itself, so the duplicate `tool_name` never surfaced on my side. The regression test you added is exactly what was missing.

Glad to hear it holds up on a real Pi 5 with Approve/Deny across a restart. Thanks for merging!

PR #978, opened 2026-09-02
fix(codex): restore Windows thread deep links with terminal fallback

Free-LZJ:
## Summary

Selecting a Codex Desktop session on Windows previously brought the app window forward without selecting the requested conversation. Restore the existing `codex://threads/<thread-id>` navigation path on Windows so Dashboard and HUD clicks can open that conversation, including sessions without a source PID.

If opening the link throws or rejects, retain the best-effort terminal/window focus fallback and contain failures from both stages. Window focus alone cannot verify which Desktop conversation or composer is selected, so Desktop Direct Send explicitly uses the manual clipboard fallback. This identity guard also covers malformed thread IDs and Desktop sessions carrying Orca metadata; non-Desktop delivery retains its existing confirmation gate. English and Chinese known limitations document the behavior.

## Validation

- 457 targeted tests passed, 0 failed, 0 skipped: session focus, handoff, Telegram Direct Send, session snapshots and state.
- `git diff --check` passed. The published combined diff matches the locally reviewed and GUI-tested candidate.
- Windows 25H2 / Codex Desktop MSIX 26.901.5280.0 x64 / Electron 41.10.4: production Dashboard/HUD windows, renderers, preloads, IPC and focus functions exercised through trusted mouse events against the installed app. Six normal navigations reached the expected existing conversation, including missing-PID and minimized-window cases; screenshots were visually checked.
- An injected protocol-open rejection exercised the real Windows focus fallback without crashing. The existing conversation remained selected, confirming that window focus is not thread-selection confirmation. Deep-link navigation worked again after removing the injected fault.
- The earlier full-suite attempt on this Windows environment encountered long-command-line, sandbox GPU and symlink-permission limitations. GPU failures passed outside the sandbox; remaining symlink failures reproduced on unmodified main. The 457-test targeted run above passed independently.

## Verification boundary

The GUI harness used isolated metadata for two existing Desktop conversations. Full Clawd startup, live hook ingestion, Telegram network delivery, Codex cold startup and non-Windows runtime behavior were outside that GUI check. Protocol navigation still depends on the installed Codex Desktop version registering and handling `codex://`.

## References

- [OpenAI Commands: Deep links](https://learn.chatgpt.com/docs/reference/commands#deep-links)
- [Windows URI-handler report: openai/codex#14686](https://github.com/openai/codex/issues/14686)
- [Windows-tested `codex://` launching: openai/codex#26500](https://github.com/openai/codex/pull/26500)
- [Previous Clawd Windows fallback: #381](https://github.com/rullerzhou-afk/clawd-on-desk/pull/381)


rullerzhou-afk:
Thanks @Free-LZJ for identifying and fixing the Windows-only downgrade! We reproduced the issue on Windows: the old path could bring the correct Codex Desktop window forward while leaving the wrong conversation selected. Your deep-link change successfully opens the requested conversation.

I updated this branch with current main and a follow-up in e248b429: Desktop Direct Send now explicitly uses the manual clipboard fallback. A window-focus confirmation cannot verify the selected conversation or composer, so it must not authorize automatic pasting. This guard also covers malformed thread IDs and Desktop entries carrying Orca metadata; CLI delivery keeps its existing confirmation gate. The behavior is covered by regression tests and English/Chinese documentation.

Validation of the combined change:

- 457 targeted tests passed, with no failures or skips.
- Windows GUI checks exercised the production Dashboard/HUD renderers, preloads, IPC and focus functions against the installed Codex Desktop app. Six normal navigations reached the expected existing conversation, including missing-PID and minimized-window cases.
- An injected deep-link rejection exercised the real Windows window-focus fallback without crashing; it did not claim to select a conversation. Normal navigation worked again after removing the injected fault.

The GUI harness used isolated session metadata; full application startup and live hook ingestion were outside this check. We will merge once the remote checks for the updated head are green. Thank you for the contribution!

PR #980, opened 2026-09-02
fix(linux): keep the pet visible on all workspaces

### LINKED ISSUE #979: Linux: pet does not stay visible across workspaces (setVisibleOnAllWorkspaces is macOS-only)
## Summary

On Linux the pet does not follow the user across workspaces — it disappears as soon as you switch to another workspace and only reappears when you come back. On macOS it correctly stays visible across all Spaces.

The cause is that `setVisibleOnAllWorkspaces()` is only ever called on the macOS path.

## Environment

- Clawd on Desk 0.16.0 (.deb)
- Ubuntu 26.04, GNOME Shell 50.1, Wayland session
- Pet runs under XWayland (`--ozone-platform=x11`, as shipped)

## Steps to reproduce

1. Run Clawd on Desk on GNOME with more than one workspace.
2. Switch to another workspace.
3. The pet is gone. Switch back — it reappears.

Expected: the pet stays visible on every workspace, as it does across macOS Spaces.

## Root cause

`src/topmost-runtime.js` — the only pet-window call to `setVisibleOnAllWorkspaces` is inside `reapplyMacVisibility()`, which returns immediately off macOS:

```js
function reapplyMacVisibility() {
  if (!isMac) return;                       // src/topmost-runtime.js:244-245
  const applyElectronCrossSpace = (win) => {
    const options = { visibleOnFullScreen: true };
    if (!getShowDock()) options.skipTransformProcessType = true;
    win.setVisibleOnAllWorkspaces(true, options);   // :249
  };
  ...
}
```

That module only defines `isWin` and `isMac` (`src/topmost-runtime.js:71-72`); there is no Linux branch. Grepping the tree, no `setVisibleOnAllWorkspaces` / `_NET_WM_STATE_STICKY` handling exists for Linux anywhere outside `src/roam-fence-picker.js` (a different, short-lived window).

Electron supports `setVisibleOnAllWorkspaces()` on Linux, where it maps to `_NET_WM_STATE_STICKY`.

## Evidence that the mechanism is the right one

Setting the sticky flag externally on the pet window makes it behave exactly as expected (follows every workspace switch, no other side effects):

```
$ wmctrl -lx | grep clawd
0x00600004  0 clawd-on-desk.clawd-on-desk  host  Clawd     # desktop 0 -> vanishes on switch

$ wmctrl -i -r 0x00600004 -b add,sticky
$ wmctrl -lx | grep clawd
0x00600004 -1 clawd-on-desk.clawd-on-desk  host  Clawd     # -1 = all workspaces, follows correctly
```

I have been running this as a polling workaround for the pet window and it has been stable.

## Proposed fix

Call `setVisibleOnAllWorkspaces(true, { visibleOnFullScreen: true })` for the Linux pet windows, alongside the existing macOS path.

Note the pet is drawn by **two** stacked windows and both need it — the render window and the transparent hit window (`isPetWindow()` in `src/topmost-runtime.js` covers exactly this pair). `src/pet-window-runtime.js` already has `isLinux` branches at both creation sites (~:2207 for `renderWin`, and the hit window at ~:2291), so there is a natural place to hook in, and it should be re-applied wherever `reapplyMacVisibility()` is currently invoked so it survives window re-creation.

Bubbles / HUD windows may deserve the same treatment, but I only verified the pet windows.

I've opened a PR with this change plus tests. Happy to reshape it if you'd rather have the call live in the `reapplyMacVisibility()` seam (renamed for both platforms) so it is re-applied at every lifecycle point instead of only at window creation.

yunho09:
Fixes #979.

## Problem

On Linux the pet disappears the moment you switch workspace and only reappears when you switch back. On macOS it correctly follows the user across Spaces.

## Cause

The only pet-window call to `setVisibleOnAllWorkspaces()` is inside `reapplyMacVisibility()` (`src/topmost-runtime.js`), which returns immediately off macOS:

```js
function reapplyMacVisibility() {
  if (!isMac) return;
  ...
  win.setVisibleOnAllWorkspaces(true, options);
}
```

That module defines only `isWin` / `isMac`, and nothing else pins the pet windows on Linux. So the pet is never marked sticky there.

Electron's `setVisibleOnAllWorkspaces()` maps to `_NET_WM_STATE_STICKY` on Linux. Setting that flag externally on the shipped build makes it behave exactly as expected, which confirmed the mechanism:

```
$ wmctrl -lx | grep clawd
0x00600004  0 clawd-on-desk.clawd-on-desk  host  Clawd     # desktop 0 -> vanishes on switch
$ wmctrl -i -r 0x00600004 -b add,sticky
0x00600004 -1 clawd-on-desk.clawd-on-desk  host  Clawd     # -1 = all workspaces, follows correctly
```

## Change

Adds a small Linux-only helper, `applyLinuxAllWorkspaces()`, next to where `reapplyMacVisibility` is wired up. It calls `setVisibleOnAllWorkspaces(true)` on both pet windows (the render window and the input window) before each window's first `showInactive()`.

- **No options.** `visibleOnFullScreen` and `skipTransformProcessType` are macOS-only.
- **Other platforms untouched.** macOS keeps going through `reapplyMacVisibility()`. In `createHitWindow()`, the call sits after main's Windows `prepareActivation()` setup and its fallback.
- **Guarded.** The call is wrapped in `try`, so a missing or throwing native call cannot break window setup.
- **Hit window contract kept.** `focusable: !isLinux` on `hitWin` (the drag contract in AGENTS.md) is not touched.

On XWayland the render window is the one that matters. The input window is created non-focusable, which Chromium turns into an override-redirect X11 window that mutter already keeps on every workspace.

## Tests

`test/pet-window-runtime.test.js` adds `describe("cross-workspace visibility (#979)")`. Every case uses mutually exclusive platform flags:

- **Linux:** both windows get `setVisibleOnAllWorkspaces(true)`, each before its first `showInactive()`.
- **Windows:** neither window gets the call.
- **macOS:** neither window gets a direct call, and both window constructors still delegate to `reapplyMacVisibility()`.
- **Linux, method missing or throwing:** window creation still completes.

Results:

- `node --test test/pet-window-runtime.test.js`: 151 pass, 0 fail.
- Full `npm test` on Linux: the same 17 failures as `main`, all in unrelated suites. The only difference is 5 more passing tests.

## Manual verification

Smoke test of `aa7fbfc0` on Ubuntu 26.04 / GNOME Shell 50.1 (Wayland session, pet under XWayland), with the external `wmctrl` workaround disabled. It covers workspace switching with visibility and click/drag, hide/show, mini-mode return, and app restart, with an unpatched `main` control run. `41dfe837` on top changes only comments and blank lines under `src/`, so the runtime code at the head is the code that was tested. Details are in the review follow-up comment on this PR.

## Known side effect

Once sticky, the WM-managed render window can take keyboard focus on every workspace, not just the one it started on. A/B test on GNOME: after switching workspace 1 → 2 → 1, focus went back to the previously active app on `main`, but stayed on the pet on this branch. The follow-up comment has details. This PR does not address it.

https://claude.ai/code/session_015qaSrRmBK1WvvewxyuSdVb

### DIFF
--- a/src/pet-window-runtime.js
+++ b/src/pet-window-runtime.js
@@ -153,6 +153,20 @@ function createPetWindowRuntime(options = {}) {
   const buildTrayMenu = options.buildTrayMenu || noop;
   const buildContextMenu = options.buildContextMenu || noop;
   const reapplyMacVisibility = options.reapplyMacVisibility || noop;
+
+  // The pet has to follow the user across workspaces the way it follows macOS
+  // Spaces. macOS goes through reapplyMacVisibility(); Linux has no equivalent
+  // call anywhere in the tree, so the pet simply disappeared when the user
+  // switched workspace. Electron's setVisibleOnAllWorkspaces() maps to
+  // _NET_WM_STATE_STICKY on Linux, which is exactly the missing piece. Guarded
+  // because window managers that ignore the hint must not break window setup.
+  function applyLinuxAllWorkspaces(win) {
+    if (!isLinux || !win) return;
+    if (typeof win.setVisibleOnAllWorkspaces !== "function") return;
+    try {
+      win.setVisibleOnAllWorkspaces(true, { visibleOnFullScreen: true });
+    } catch {}
+  }
   // #640: re-run the editing-overlap dodge whenever the hit geometry syncs —
   // the hitbox can change without the window moving (state switches between
   // hitboxes, theme reload), which changes the overlap answer.
@@ -2205,6 +2219,7 @@ function createPetWindowRuntime(options = {}) {
     renderWin.setFocusable(false);
 
     if (isLinux) {
+      applyLinuxAllWorkspaces(renderWin);
       renderWin.on("close", (event) => {
         if (!isQuitting()) {
           event.preventDefault();
@@ -2311,6 +2326,7 @@ function createPetWindowRuntime(options = {}) {
     // window until createHitWindow() returns and the caller assigns it.
     applyHitInputState(hitWin);
     if (isMac) hitWin.setFocusable(false);
+    applyLinuxAllWorkspaces(hitWin);
     hitWin.showInactive();
     keepOutOfTaskbar(hitWin);
     if (isWin) hitWin.setAlwaysOnTop(true, topmostLevel);
--- a/test/pet-window-runtime.test.js
+++ b/test/pet-window-runtime.test.js
@@ -33,6 +33,7 @@ function makeWindow(bounds = { x: 10, y: 20, width: 100, height: 100 }) {
     setShape: (shape) => calls.push(["setShape", shape]),
     setIgnoreMouseEvents: (value) => calls.push(["setIgnoreMouseEvents", value]),
     setAlwaysOnTop: (...args) => calls.push(["setAlwaysOnTop", ...args]),
+    setVisibleOnAllWorkspaces: (...args) => calls.push(["setVisibleOnAllWorkspaces", ...args]),
     setFocusable: (value) => calls.push(["setFocusable", value]),
     showInactive: () => calls.push(["showInactive"]),
     hide: () => calls.push(["hide"]),
@@ -2825,6 +2826,67 @@ describe("pet-window-runtime", () => {
       true,
       "pop-up-menu",
     ]);
+  });
+
+  it("pins both pet windows to all workspaces on Linux so the pet follows workspace switches", () => {
+    const renderInstances = [];
+    const hitInstances = [];
+    const harness = createRuntime({ isLinux: true });
+
+    harness.runtime.createRenderWindow({
+      BrowserWindow: makeBrowserWindow(renderInstances),
+      size: { width: 120, height: 120 },
+      initialWindowBounds: { x: 40, y: 0, width: 120, height: 120 },
+      initialVirtualBounds: { x: 40, y: 0, width: 120, height: 120 },
+      preloadPath: "preload.js",
+      loadFilePath: "index.html",
+      themeConfig: { ok: true },
+      setRenderWindow: harness.setRenderWin,
+      isQuitting: () => false,
+    });
+    harness.runtime.createHitWindow({
+      BrowserWindow: makeBrowserWindow(hitInstances),
+      preloadPath: "preload-hit.js",
+      loadFilePath: "hit.html",
+      hitThemeConfig: { ok: true },
+    });
+
+    const expected = ["setVisibleOnAllWorkspaces", true, { visibleOnFullScreen: true }];
+    assert.deepStrictEqual(
+      renderInstances[0].calls.find((call) => call[0] === "setVisibleOnAllWorkspaces"),
+      expected,
+    );
+    assert.deepStrictEqual(
+      hitInstances[0].calls.find((call) => call[0] === "setVisibleOnAllWorkspaces"),
+      expected,
+    );
+  });
+
+  it("does not touch cross-workspace visibility off Linux (macOS keeps reapplyMacVisibility)", () => {
+    const renderInstances = [];
+    const hitInstances = [];
+    const harness = createRuntime({ isLinux: false });
+
+    harness.runtime.createRenderWindow({
+      BrowserWindow: makeBrowserWindow(renderInstances),
+      size: { width: 120, height: 120 },
+      initialWindowBounds: { x: 40, y: 0, width: 120, height: 120 },
+      initialVirtualBounds: { x: 40, y: 0, width: 120, height: 120 },
+      preloadPath: "preload.js",
+      loadFilePath: "index.html",
+      themeConfig: { ok: true },
+      setRenderWindow: harness.setRenderWin,
+      isQuitting: () => false,
+    });
+    harness.runtime.createHitWindow({
+      BrowserWindow: makeBrowserWindow(hitInstances),
+      preloadPath: "preload-hit.js",
+      loadFilePath: "hit.html",
+      hitThemeConfig: { ok: true },
+    });
+
+    assert.equal(renderInstances[0].calls.some((call) => call[0] === "setVisibleOnAllWorkspaces"), false);
+    assert.equal(hitInstances[0].calls.some((call) => call[0] === "setVisibleOnAllWorkspaces"), false);
   });
 
   it("reloadWindowWebContents ignores destroyed windows and webContents", () => {


rullerzhou-afk:
Thanks @yunho09 for tracking down #979 and covering both the render window and the input window. I reviewed `76f59f188d928b43b5678764476a05845c44a98d` with independent reviewers. The small Linux-only helper is a reasonable approach; I do not see a demonstrated need to move this into the macOS visibility lifecycle or expand this PR to bubbles/HUD.

Please address the following before the next review:

1. **Update against current `main` and preserve the Windows activation work.** The actual conflict is next to `createHitWindow()`'s first show: current main has `prepareActivation()` and its fallback, while this PR inserts `applyLinuxAllWorkspaces(hitWin)`. Keep both, before `showInactive()`, and retain main's Windows constructor behavior and activation/fallback tests. Taking the PR side wholesale would undo that Windows fix.

2. **Use real, mutually exclusive platform flags in the new tests.** `createRuntime()` defaults `isWin` to `true`, so the new `createRuntime({ isLinux: true })` test currently runs with both Windows and Linux enabled. Please explicitly use `{ isLinux: true, isWin: false, isMac: false }`, and cover Windows and macOS separately for the negative case. In the Mac case, also assert that the existing `reapplyMacVisibility()` delegation still happens. I verified the gap with an in-memory mutation: a deliberately broken helper that requires both `isWin` and `isLinux` still passes both new tests, but the corrected Linux fixture catches it. This is a test coverage issue, not a claim that the current helper contains that bug.

3. **Record a smoke test of the updated PR commit in your reported GNOME/XWayland environment, with the external `wmctrl` workaround/poller disabled.** Please include workspace switching with both visibility and click/drag working, plus hide/show, mini-mode return, and app restart. The existing external-flag experiment is useful diagnosis; the remaining question is whether the actual patched app preserves the behavior. If hide/show reproduces a loss, address the affected re-show path; otherwise keep the creation-time helper. No broader Linux/compositor matrix is being requested here.

Two small optional improvements: use `setVisibleOnAllWorkspaces(true)` without the macOS-only `visibleOnFullScreen` option, and add missing-method/throwing-method tests if keeping those defensive guards. The option's platform scope is documented in [Electron's API reference](https://www.electronjs.org/docs/latest/api/base-window#winsetvisibleonallworkspacesvisible-options).

I independently ran `node --test test/pet-window-runtime.test.js` at the reviewed head: **142 passed, 0 failed, 0 skipped**. This is focused automated evidence; I have not performed Linux GUI validation. Thanks again for keeping the fix narrowly scoped.

yunho09:
Thanks for the thorough review, @rullerzhou-afk. All three points are addressed in three commits:

- `ab3c9ec2`: merge of `main`
- `aa7fbfc0`: the review fixes
- `41dfe837`: tighter tests and a corrected helper comment from an independent review pass. It changes only comments and blank lines under `src/`, so the runtime code at the PR head is exactly the code smoke-tested below.

I merged `main` instead of rebasing, so `76f59f18` stays reviewable and there was no force-push.

### 1. Updated against `main`, Windows activation work kept

Merged `main` at `782edefc`. In `createHitWindow()`, main's `prepareActivation()` block and its `setFocusable(true)` fallback now come first, then `applyLinuxAllWorkspaces(hitWin)`. Both run before `showInactive()`. The merge result differed from `main` by exactly the original +78/−0, so nothing of main's was removed. That includes the `show: false` / `focusable` constructor options and the three activation/fallback tests, all still passing. `git diff main...HEAD` now touches only `src/pet-window-runtime.js` (+17) and `test/pet-window-runtime.test.js` (+130).

### 2. Mutually exclusive platform fixtures

The two old cases are replaced by `describe("cross-workspace visibility (#979)")`:

- **Linux** `{ isLinux: true, isWin: false, isMac: false }`: both windows get exactly `setVisibleOnAllWorkspaces(true)`, each before its first `showInactive()`.
- **Windows** `{ isWin: true, isMac: false, isLinux: false }`: neither window gets the call.
- **macOS** `{ isMac: true, isWin: false, isLinux: false }`: neither window gets a direct call, and render-window and hit-window creation each still delegate to `reapplyMacVisibility()`.
- **Linux, method missing** and **Linux, method throws**: window creation still completes, and both windows still show and load. Both cases also check that their fake took effect: no method on the window, and exactly two throws.

To check that the tests can actually fail, I ran the new block against eight hand-made mutants of `src/pet-window-runtime.js`. Every mutant failed at least one case:

- your helper that requires both `isWin` and `isLinux`
- a gate on `!isMac` or on `!isWin` instead of `isLinux`
- no `try/catch`
- `visibleOnFullScreen` put back
- the hit window pinned after `showInactive()`
- the render-window call dropped
- `reapplyMacVisibility()` dropped from `createHitWindow()`

`node --test test/pet-window-runtime.test.js`: 151 passed, 0 failed. Full `npm test` on Linux: the branch (at `41dfe837`) has 10,170 tests / 10,091 pass / 17 fail / 13 cancelled, and `main` has 10,165 / 10,086 / 17 / 13. The failing test names are identical on both, all in unrelated suites (e.g. Dashboard renderer, notifier, shortcuts, SVG channels, terminal focus). The only difference is the 5 new passing cases.

**Optional items:** both done. The call is now `setVisibleOnAllWorkspaces(true)` with no options. For the guards, I kept the `try/catch`, now covered by the missing-method and throwing-method tests. I dropped the separate `typeof` check: the `try/catch` already catches the `TypeError` a missing method throws, and removing the check changed no test result.

### 3. Smoke test of `aa7fbfc0`

**Environment:** Ubuntu 26.04 LTS, GNOME Shell 50.1, Wayland session. The pet runs under XWayland (`--ozone-platform=x11`) on Electron 41.10.4, launched from source with `node launch.js`. The external `wmctrl` watcher and the installed app were stopped before the first launch, and the watcher's own log has no pins during the test.

Two setup details, for transparency:
- The source build used an isolated `HOME` holding a copy of my prefs, so it couldn't rewrite my real hook configs.
- It ran with `CLAWD_DISABLE_SANDBOX=1`, because Ubuntu's AppArmor user-namespace restriction blocks the Chromium sandbox for an unpackaged Electron.

**Method:** a probe logged the pet windows' X11 state roughly every 200 ms while I did the GUI steps by hand. Mutter keeps X11 windows on other workspaces mapped but marks them `WM_STATE=Iconic`. So a render window that is still `Normal` after a switch is on screen. Workspace numbers below are 0-based `_NET_CURRENT_DESKTOP` indices.

**Control run on `main` (`782edefc`), same setup:** the render window had `_NET_WM_DESKTOP=2` and no `STICKY`. After switching to workspace 3 it went `Iconic`, which reproduces #979.

| Step (`aa7fbfc0`) | Result | Evidence |
|---|---|---|
| Launch | ✅ | render window has `_NET_WM_STATE_STICKY` and `_NET_WM_DESKTOP=0xFFFFFFFF` |
| Workspace switching | ✅ visible | render window stayed `Normal` on every workspace visited (0/1/2/3), with both manual and `wmctrl` switches |
| Click / drag after switching | ✅ | launched on workspace 2, then right-click menu → Mini Mode worked on workspace 0; a drag on workspace 1 moved the window from `+3552+572` to `+2674+734` |
| Hide / show | ✅ no loss | Hide Pet unmapped both windows; Show Pet (tray) mapped them again with `STICKY` and desktop `ALL` still set; still visible after switching |
| Mini mode | ✅ | mini pet stayed visible across switches; Exit Mini Mode went back to the previous position, still visible on another workspace |
| App restart | ✅ | Quit from the menu while in mini mode, then relaunch: the new window is `STICKY` and visible on another workspace |

Hide/show kept the state, so per your guidance I kept the creation-time helper and left the re-show paths alone.

Two findings from the smoke test that you should know about:

1. **The input window is override-redirect on X11.** The hit window is created with `focusable: false` on Linux, and Chromium turns non-activatable X11 windows into override-redirect ones. Mutter never manages it, so it stays mapped on every workspace even without this PR: in the control run it stayed mapped while the render window went `Iconic`. Calling the helper on it has no visible effect here; its `_NET_WM_STATE` never shows `STICKY`. I kept the call so both pet windows follow one contract. I'm happy to drop it if you'd rather the diff touch only the render window.

2. **Focus side effect (not addressed in this PR).** The render window is WM-managed and has no `WM_HINTS`, even after `setFocusable(false)`, so mutter can give it keyboard focus. On `main` it already takes `_NET_WM_STATE_FOCUSED` at startup. Once it is sticky, it can take focus on every workspace. A/B test: activate VS Code on workspace 1, then switch 1 → 2 → 1. On `main`, focus went back to VS Code. On this branch it stayed on the pet. The probe log also shows `FOCUSED` moving to the pet after my manual switches. I didn't try to fix it here: the Electron-level fixes I can see are constructing the render window non-focusable (which makes it override-redirect) or a different window type, and both change stacking and clamping far beyond this fix. I wanted you to be able to weigh it before merging, and I can open a separate issue.

PR #981, opened 2026-09-02
fix(dashboard): hide unavailable remote session actions

undefined-moe:
## Summary
- Hide the disabled Jump to Terminal button for remote sessions that cannot be focused locally
- Hide the session automation dropdown when a session cannot configure automation and has no existing override
- Keep local/WebUI unavailable states and existing override revocation visible

## Tests
- node --test test/session-renderer-behavior.test.js test/session-focus-unavailable.test.js
- git diff --check

### DIFF
--- a/src/dashboard-renderer.js
+++ b/src/dashboard-renderer.js
@@ -862,28 +862,32 @@ function createCard(session, now) {
 
   const actions = document.createElement("div");
   actions.className = "actions";
-  const button = document.createElement("button");
-  button.type = "button";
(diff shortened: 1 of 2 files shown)


rullerzhou-afk:
Merged — thank you, @undefined-moe, for this focused Dashboard cleanup! Hiding unusable controls makes the session cards clearer, and I appreciate the care taken to preserve the existing override revocation path.

I reviewed the change and verified the Dashboard on macOS with isolated test sessions, including override revocation and “Mark read.” All 189 focused tests and the repository asset audit passed. Thanks for the contribution!

PR #982, opened 2026-09-03
docs: document Homebrew cask install

jlimcode:
## Summary
- Document `brew install --cask clawd-on-desk` in the Quick Start READMEs (en / zh-CN / zh-TW / ja / ko / es)
- Add Homebrew as an install and upgrade path in the macOS and Linux setup guides
- Note that Homebrew cask installs can update with `brew upgrade --cask clawd-on-desk`

The cask landed in Homebrew/homebrew-cask as [clawd-on-desk 0.16.0](https://github.com/Homebrew/homebrew-cask/pull/284822). Linux is x86_64 AppImage only, matching the upstream release assets.


rullerzhou-afk:
Thanks for getting Clawd into Homebrew and following up with the installation docs! This gives Homebrew users a convenient way to install and upgrade Clawd. We appreciate the work on both sides, and we’ll review the documentation against our newly released v1.0.0.

PR #993, opened 2026-09-04
fix(state): probe working source pid only after the working window

### LINKED ISSUE #992: [BUG] v1.0.0: every working session is deleted between hook events (working-source-exit lost its silence gate) — pet flips busy ⇄ idle
**现象（TL;DR）**：升级到 v1.0.0 后，桌宠在有会话持续运行时不停地在「忙碌」和「空闲」之间来回跳。每个 hook 事件之后不到 1 秒，会话就被 `working-source-exit` 删掉，下一个事件再把它建回来。

## Symptom

With one or more Claude Code sessions continuously running, the pet flickers between the working and idle animations. `session-debug.log` shows every working session being deleted within ~1s of each hook event and re-created by the next one:

```
23:56:39.324 event sid=s1.bG9jYWw.YjNiM2Uw… state=working … agentPid=36076 sourcePid=32956 pidReachable=1 -> incoming=working/PreToolUse
23:56:43.984 stale-delete working-source-exit sid=s1.bG9jYWw.YjNiM2Uw… state=working … agentPid=36076 sourcePid=29688 pidReachable=1
23:56:51.283 event sid=s1.bG9jYWw.YjNiM2Uw… <deleted> -> incoming=working/PreToolUse
23:56:53.055 stale-delete working-source-exit sid=s1.bG9jYWw.YjNiM2Uw… state=working … agentPid=36076 sourcePid=27132 pidReachable=1
```

317 `working-source-exit` deletions in one ~9-minute window, across three live sessions. `agentPid` is stable and alive throughout; `sourcePid` is different on every single event.

## Environment

- Clawd on Desk v1.0.0 (Windows 11, x64, NSIS install)
- Claude Code 2.1.260, running inside the Claude desktop app (sessions launched by `claude.exe`, whose parent chain is `claude.exe → claude.exe → explorer.exe`)
- Not reproducible on v0.16.0 with the same setup.

## Root cause

Two things combine. Only the first one is new in v1.0.0.

**1. The `working-source-exit` delete lost its silence gate (regression, `05372be6`, #960).**

Before v1.0.0 the source-liveness probe only ran for a working session that had already been quiet for the working window:

```js
} else if (age > workingStaleMs) {
  if (session.pidReachable && session.sourcePid && !isProcessAlive(session.sourcePid)) {
    return { action: "delete", reason: "working-source-exit" };
  }
  …
}
```

`05372be6` moved the probe to the top of a new working-like branch that runs on **every** sweep, with no age condition (`src/state-stale-cleanup.js`):

```js
if (isWorkingLikeState(session.state)) {
  if (session.pidReachable && session.sourcePid && !isProcessAlive(session.sourcePid)) {
    return { action: "delete", reason: "working-source-exit" };
  }
  if (workingStaleMs > 0 && age > workingStaleMs) { … }
  return { action: null };
}
```

A session that is actively reporting events is now judged on its source PID on every sweep, which is exactly when that PID is least meaningful.

**2. On Windows, the Claude hook's `source_pid` is a per-event process (pre-existing).**

Claude Code launches each hook through a throwaway wrapper. Sampled live chains for the Clawd hook:

```
node.exe (hook, 26308) <- pwsh.exe (35836) <- claude.exe (12912) <- claude.exe (5420) <- explorer.exe
```

`pwsh.exe` is in `BASE_TERMINAL_NAMES_WIN`, so the process walk in `hooks/shared-process.js` picks that wrapper as `terminalPid`, and `stablePid = terminalPid || lastGoodPid` ships it as `source_pid`. The wrapper exits with the hook, so the stored `sourcePid` is already dead when the next sweep reads it. Every cached entry confirms it — `agentPid` alive, `stablePid` dead:

```
{"version":2,"namespace":"claude-code","cwd":"C:\FormerD\Repos\clawd-on-desk",
 "stablePid":5724,   ← dead
 "agentPid":36076}   ← alive
```

This is harmless as long as the PID is only consulted after a real silence window; with (1) it deletes every live session between events. (The launcher chain here has no real terminal above `claude.exe`, so the outermost-terminal-wins rule in the walk never corrects it — a session started from Windows Terminal does get a stable pid.)

## Fix

Restore the silence gate: probe source liveness only once the working-like session has actually gone quiet for its effective working window, keeping `05372be6`'s separation of the active-turn timeout from the idle-card cutoff. Orphaned/lingering working sessions are still retired after `workingStaleMs`, exactly as before v1.0.0.

PR follows.

## Note (separate defect, not fixed here)

The wrong `source_pid` also degrades click-to-focus for these sessions — `src/state.js` merges `sourcePid || existing.sourcePid`, so the per-event value overwrites whatever the walk resolved at `SessionStart`, and focus then targets a dead PID. The comment block above `stablePid` in `hooks/shared-process.js` already calls this hazard out for the empty-walk cases; the pwsh-wrapper case reaches it through a *successful* walk instead. Happy to open a second issue if you want that tracked separately.

PeterShanxin:
Fixes #992.

## Problem

`getStaleSessionDecision` probes `sourcePid` liveness for every working-like session on **every** sweep, with no age condition. A session that reports a hook event every second is therefore deleted about a second after each event and recreated by the next one, so the pet flips between the working and idle animations for the whole run.

The probe used to be gated behind the working window (`else if (age > workingStaleMs)`); `05372be6` (#960) moved it into the new working-like branch and dropped the gate.

That combines badly with how `source_pid` is produced on Windows. Claude Code launches each hook through a throwaway wrapper:

```
node.exe (hook) <- pwsh.exe <- claude.exe <- claude.exe <- explorer.exe
```

`pwsh.exe` is in `BASE_TERMINAL_NAMES_WIN`, so the walk in `hooks/shared-process.js` selects that per-event wrapper as `terminalPid` and ships it as `source_pid`. It is dead before the next sweep reads it, while `agentPid` stays alive the whole session. From one ~9-minute window of `session-debug.log` on v1.0.0, three live sessions, 317 deletions:

```
23:56:39.324 event sid=…YjNiM2Uw… state=working … agentPid=36076 sourcePid=32956 -> incoming=working/PreToolUse
23:56:43.984 stale-delete working-source-exit sid=…YjNiM2Uw… state=working … agentPid=36076 sourcePid=29688
23:56:51.283 event sid=…YjNiM2Uw… <deleted> -> incoming=working/PreToolUse
```

## Change

Move the source-liveness probe back inside the working-window check. A working-like session is only judged on its source PID once it has actually been silent for its effective `workingStaleMs`; before that it is left alone.

Behaviour preserved:

- `agent-exit` is unchanged and still deletes immediately when the agent process dies — that is the branch that retires a session whose terminal was really closed, since the agent dies with it.
- Orphaned/lingering working sessions are still deleted with `working-source-exit` after `workingStaleMs`, as they were before v1.0.0.
- `05372be6`'s separation of the active-turn timeout from the idle-card cutoff is kept: a working-like session still settles through the working timeout first and stamps that transition.

One deliberate consequence worth a maintainer's eye: with `workingStaleMs === 0` (the local-Codex "never idle a silent turn by age" choice) the source probe no longer runs for those sessions either. `agent-exit` still covers a dead Codex process, and `codex-hook.js` sends `source_pid = agentPid` on the `preferAgentPid` path, so I do not think anything is lost — say the word if you would rather keep a floor there.

## Verification

- New regression test in `test/state-stale-cleanup.test.js`: an actively reporting working session whose `sourcePid` is dead and whose `agentPid` is alive yields `{ action: null }`. It fails on `main` with `{ action: 'delete', reason: 'working-source-exit' }`.
- `node --test test/state-stale-cleanup.test.js` — 46/46 pass. The two existing `working-source-exit` assertions (the generic one and the OpenCode stale-floor one) both already use aged sessions and still pass unchanged.
- Full `node test/run-tests.js` on Windows 11: 9410 pass / 2 fail. Both failures reproduce on unmodified `main` in the same worktree and are environmental — `cigarette-accessory-electron` (Electron render harness) and one Reasonix 5s-timeout timing test that passes on re-run.

### DIFF
--- a/src/state-stale-cleanup.js
+++ b/src/state-stale-cleanup.js
@@ -167,11 +167,18 @@ function getStaleSessionDecision(session, options = {}) {
   // and stamp that transition. Otherwise a 20-minute-old Codex turn can be
   // changed to idle with its old timestamp, then deleted immediately by the
   // ordinary 10-minute idle cutoff on the next sweep.
+  //
+  // Source liveness only carries information once that window has elapsed. A
+  // source_pid frequently belongs to a per-event launcher rather than to the
+  // session host — Windows Claude Code runs every hook through a throwaway
+  // pwsh wrapper, so the pid shipped with an event is already gone when the
+  // next sweep reads it. Probing it while the turn is still reporting deletes
+  // the live session between events, and the following event recreates it.
   if (isWorkingLikeState(session.state)) {
-    if (session.pidReachable && session.sourcePid && !isProcessAlive(session.sourcePid)) {
-      return { action: "delete", reason: "working-source-exit" };
-    }
     if (workingStaleMs > 0 && age > workingStaleMs) {
+      if (session.pidReachable && session.sourcePid && !isProcessAlive(session.sourcePid)) {
+        return { action: "delete", reason: "working-source-exit" };
+      }
       return { action: "idle", reason: "working-timeout", updateTimestamp: true };
     }
     return { action: null };
--- a/src/state-stale-cleanup.js
+++ b/src/state-stale-cleanup.js
@@ -168,17 +168,24 @@ function getStaleSessionDecision(session, options = {}) {
   // changed to idle with its old timestamp, then deleted immediately by the
   // ordinary 10-minute idle cutoff on the next sweep.
   //
-  // Source liveness only carries information once that window has elapsed. A
+  // With an age window enabled, only check source liveness after it elapses. A
   // source_pid frequently belongs to a per-event launcher rather than to the
   // session host — Windows Claude Code runs every hook through a throwaway
   // pwsh wrapper, so the pid shipped with an event is already gone when the
   // next sweep reads it. Probing it while the turn is still reporting deletes
   // the live session between events, and the following event recreates it.
   if (isWorkingLikeState(session.state)) {
-    if (workingStaleMs > 0 && age > workingStaleMs) {
-      if (session.pidReachable && session.sourcePid && !isProcessAlive(session.sourcePid)) {
-        return { action: "delete", reason: "working-source-exit" };
-      }
+    const workingWindowElapsed = workingStaleMs > 0 && age > workingStaleMs;
+    // Zero disables age-based expiry, not process-death cleanup. In particular,
+    // a local Codex session may have only a source PID when agent PID discovery
+    // failed, so the earlier agent-exit check cannot retire it on its own.
+    if (
+      (workingStaleMs === 0 || workingWindowElapsed)
+      && session.pidReachable && session.sourcePid && !isProcessAlive(session.sourcePid)
+    ) {
+      return { action: "delete", reason: "working-source-exit" };
+    }
+    if (workingWindowElapsed) {
       return { action: "idle", reason: "working-timeout", updateTimestamp: true };
     }
     return { action: null };
--- a/test/state-stale-cleanup.test.js
+++ b/test/state-stale-cleanup.test.js
@@ -176,6 +176,22 @@ describe("state stale cleanup decisions", () => {
     }).result, { action: null });
   });
 
+  it("does not probe the source before or at a positive working-window boundary", () => {
+    const now = 2_000_000;
+    for (const state of ["working", "thinking", "juggling"]) {
+      for (const age of [1_000, WORKING_STALE_MS]) {
+        for (const agentPid of [null, 10]) {
+          const { result, calls } = decision(session({
+            state, agentId: "claude-code", agentPid, sourcePid: 20,
+            updatedAt: now - age,
+          }), { now, alivePids: new Set([10]) });
+          assert.deepStrictEqual(result, { action: null });
+          assert.deepStrictEqual(calls, agentPid ? [10] : []);
+        }
+      }
+    }
+  });
+
   it("keeps working-like state set explicit", () => {
     assert.strictEqual(isWorkingLikeState("working"), true);
     assert.strictEqual(isWorkingLikeState("thinking"), true);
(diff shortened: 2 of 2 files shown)


rullerzhou-afk:
Thanks @PeterShanxin for the clear diagnosis, logs, and focused regression test! I reproduced the premature source-exit deletions on the pre-PR baseline and confirmed your working-window guard fixes that real Windows hook chain.

I pushed a small maintainer follow-up, `ab155208`, to this branch. It keeps your positive-window behavior unchanged and preserves source-exit cleanup when `codexWorkingStaleMs=0`. This covers the fallback where agent PID discovery failed but a previously reachable source PID is available. Zero still means no age-based expiry. This supersedes the zero-timeout consequence discussed in the original description; no additional timeout floor is introduced.

Validation of the updated head:

- Independent second review found no blockers; focused tests 50/50 and adjacent state/hook/settings tests 650/650 passed. Additional differential review checked both decisions and PID probe order against the relevant base/head behavior without mismatches.
- Actual Windows Claude Code 2.1.260 API-backed session with isolated Clawd Electron runtime, real PowerShell hooks and four sequential `sleep 12` tool calls: 61 seconds, zero premature source-exit deletions, zero idle/empty-session samples during the turn, normal completion afterward. OS process snapshots confirm the temporary PowerShell sources exited while Claude stayed alive. The same earlier base scenario produced six premature deletions. This is Claude Code CLI/ConPTY testing, not a Claude Desktop GUI claim.
- Zero-timeout boundary in the real Electron runtime: controlled missing-agent-PID metadata plus an actual live Node source process. The session remained while that process was alive, then was deleted by the normal production sweep after it voluntarily exited. This is a real-process boundary test, not a claimed natural Codex PID-discovery failure reproduction.
- Local full suite: 9,411 pass / 5 fail / 42 skip. All five failures are Windows symlink `EPERM` setup failures, reproduced identically on untouched main. The long worktree exceeded the runner's Windows absolute-argv limit, so the same full inventory and timeout were run with relative test paths. Fresh CI remains the merge gate.
- Test CLI and Electron exited normally; checked user settings/runtime/prefs and the installed Claude executable were unchanged in this retest.

Thanks again! I'll merge once the updated-head CI is green.

PR #999, opened 2026-09-07
feat(dashboard): add dashboard session model row

easyhak:
## Summary

- render `Model: <id>` as its own row on dashboard session cards when `session.model` is set
- report model from the Claude Code hook (`SessionStart.model`) — the largest remaining producer gap
- keep that label current across `/model` switches by carrying the model on the statusline's existing `metadata_only` channel
- reject unusable model ids at the hook boundary (blank, >128 chars, NUL / line breaks)
- add `dashboardModel` copy in all seven locales

## Why

`model` has been wired through the server since #270 (`160bb00e`) for WebUI metadata, and six hooks already send it (Codex, Qwen Code, QwenWork, Qoder, QoderWork, ZCode) — but nothing in the UI ever read it. Claude Code was the biggest remaining producer gap; this PR adds the dashboard row and closes that producer in one pass.

The renderer is agent-agnostic (`session.model` only). Hooks that already populate the field work as-is; agents that send nothing get no row. Codex sets `model` in its generic state and permission builders, so it stays current across switches; Qwen Code forwards whatever its payloads carry, which upstream populates on `SessionStart`.

Claude Code reports it twice, deliberately. `SessionStart` labels the session immediately, and `hooks/claude-statusline.js` then re-reports the live model on every refresh. That second path reuses machinery that already exists: the statusline already POSTs `metadata_only: true`, which `src/server-route-state.js` routes to `updateSessionMetadata` *around* the `updateSession` lifecycle machine — no session creation, no `recentEvents` append, no `updatedAt` bump. So a `/model` switch relabels the card without disturbing the completion badge or the last-event row, on every Claude Code version and without registering either model-switch hook. `updateSessionMetadata` writes only on a real change, so the sub-second statusline heartbeat adds no broadcasts, and `metadataUpdatedAt` is deliberately not stamped — that field is context/quota telemetry freshness.

It does change how often the statusline POSTs, in one direction. `buildStateBody` used to return `null` — suppressing the request entirely — when there was neither quota nor context usage; a model alone now produces a body. For subscription users nothing changes, since `rate_limits` is present on every refresh and they already POSTed each time. What is new is one localhost POST per refresh for sessions that report neither: API-key/Bedrock/Vertex users before their first token usage, and Claude Code versions that do not emit `context_window.current_usage`. The statusline is a fresh process per render, so there is nowhere to cache a last-posted value without adding a disk read/write that costs more than the POST; the request is best-effort with a 150ms timeout, and an unchanged model stops at `updateSessionMetadata` without a broadcast.

The statusline reports `model.id`, never `model.display_name`: `SessionStart` reports the id, so falling back to a display name would flip the card label on the first refresh. It rides the same `isClaudeStatuslineMetadataAllowed` gate as context usage — with Claude quota collection off the statusline is not installed at all, and the card falls back to the `SessionStart` value.

## Known limitations

- `SessionStart.model` is optional upstream — `startup` and `compact` pass a model, `clear` does not, and the two `resume`/`fork` call sites differ. The sticky merge in `state.js` (`model || existing.model`) keeps a session labelled once anything reports one, and the next statusline refresh fills the gap
- startup-recovered sessions begin with `model: null` (the recovery lease carries no model), so after a Clawd restart the row is missing until the next statusline refresh — and stays missing for a session whose statusline telemetry gate is closed
- model ids are normalized at the Claude hook boundary only. The server trims and nothing more, and the other six producers forward their payload verbatim. Now that the value is rendered and rides `sessionSnapshotSignature` on every broadcast, moving that normalization into the state layer (beside `normalizeTitle`) is a worthwhile follow-up
- a subagent on its own model is not distinguished; no hook payload exposes it (`SubagentStart` carries only `agent_id`/`agent_type`). The card describes the session
- Session HUD is unchanged — too narrow for another row
- Kimi, Cursor, Gemini and the other hooks that send no model simply get no row

## Validation

- `npm test` (9314 pass, 0 fail) plus targeted tests for hook parsing, malformed ids, the statusline id-only rule, the route's telemetry gate, and the `.meta` regression guard
- `updateSessionMetadata` test pins the badge contract: a `done` session stays `done` across a model relabel, with `recentEvents` / `updatedAt` / `metadataUpdatedAt` untouched
- real `dashboard-renderer.js` output with hook modules for Claude Code, Codex, Qwen Code
- statusline path exercised end to end as a real child process against a stub server: stdin payload → `model: "claude-opus-5"` on `/state`, with the visible status line still rendering `display_name`. The model-only payload (no quota, no context) was confirmed to POST as well
- upstream hook contracts read off the Claude Code 2.1.260 binary rather than assumed: the `PreModelSwitch` executor's timeout→block promotion, `SessionStart`'s optional `model` and its differing call sites, the statusline payload's `model: { id, display_name }`, and `SubagentStart`'s field set
- the `SessionStart` row was checked in the running app on in my macOS.


rullerzhou-afk:
Thanks @easyhak for contributing the Dashboard model row and keeping Claude model changes in sync through the statusline metadata channel!

The review and targeted tests passed, including a real Claude Code model switch and the Dashboard layout on Windows. Merged into `main` — thanks for the clear implementation and regression tests. 🐾

Issue #883, opened 2026-08-15
OpenCode: resumed existing sessions lack context usage until the next assistant update


PeterShanxin:
## Summary

After the #830 context-usage implementation, fresh OpenCode sessions report context usage correctly, but an **existing session resumed after a Clawd/OpenCode plugin restart or late attach can initially show no context percentage in Clawd**.

The context percentage appears later once the resumed session produces a new authoritative assistant update. This looks like a cold-start / resume hydration gap rather than a failure of the #830 real-time pipeline.

## Observed behavior

Windows 11 ARM64, OpenCode Desktop 1.18.18, packaged Clawd/OpenCode plugin build from PR #859.

1. Start Clawd with the packaged OpenCode plugin configured.
2. Start/restart OpenCode Desktop.
3. Create a **new** OpenCode session and complete a turn.
   - Clawd shows the context percentage as expected.
4. Resume/continue an **existing** OpenCode session.
   - The session title/lifecycle can appear, but context usage may initially be absent.
5. Continue using that existing session.
   - After a later assistant update, the context percentage appears and remains correct.

In the latest reproduction, the continued session initially had no context chip while a newly created session did; after additional activity, the continued session also gained its context percentage.

## Likely boundary

This is consistent with PR #859's current source of truth: context usage is emitted from authoritative `message.updated` assistant data. An already-existing session may therefore have no context metadata to send to Clawd until the next assistant completion emits another `message.updated`.

That is an inference from the observed timing; it should be verified against the actual resume/startup event sequence before implementing a fix.

## Expected behavior

When Clawd/OpenCode attaches to or resumes an existing session, the latest available context usage should ideally be hydrated without requiring the user to wait for another assistant turn.

Possible approaches should be investigated before implementation (for example, querying the current/latest assistant message on resume/startup), while preserving the ordering, generation, dedup, and lifecycle-isolation guarantees from #859.

## Scope

This is **not a blocker for PR #859**. The real-time #830 pipeline works correctly once a fresh authoritative context update arrives.

Please keep any fix separate from #859 and avoid changing its token formula, HUD semantics, lifecycle state, or async ordering guarantees.

Related: #830, PR #859.

rullerzhou-afk:
Thanks Peter, good catch. Agreed this should stay separate from #859. If you'd like to take the resume-hydration follow-up, please go ahead — we'd be happy to review it. If you'd rather leave it to us, just let us know.

Tsdsj:
[...]

PR #1006, opened 2026-09-10
fix(opencode): hydrate context usage for resumed sessions

Tsdsj:
Resumed OpenCode sessions can lack context usage until another assistant update. This adds bounded startup/resume hydration and preserves a single resume signal when history capacity is occupied or bootstrap metadata arrives before the Clawd session exists. Refs #883.

## Changes

- Inspect up to 20 recent, non-archived root sessions in the initialized directory; read one page of 20 messages per session and select the latest usable assistant, including decreased usage after compaction.
- Share four history-request slots between bootstrap and event-driven hydration, with a two-second timeout. Retain one deduplicated pending intent per session and drain automatically when capacity becomes available; explicit selections precede unserved bootstrap candidates.
- Retain a compact unaccepted sample and replay it behind the normal lifecycle POST, including when that signal arrives while the original POST is pending. Update delivered state only after metadata acceptance.
- Preserve signals received during the 30-second history-query cooldown using one instance timer; a failed retry does not create a recurring poll. Pending intents and cached samples share the existing 1024-session bound.
- Cancel stale work on deletion, recreation, disposal, eviction or newer live context, including zero-token updates. Recheck queued historical metadata before sending while preserving independently merged title updates.
- Keep metadata-only delivery, token calculation, HUD semantics and MiMo capability unchanged. Bootstrap never creates missing Clawd sessions or claims lifecycle ownership.

## Validation

- Added 24 scheduling regressions using the plugin's Node runtime and real `handleStatePost` / `state.js` modules. Both reported sequences failed on `e269ce4d`, while the normal control passed. Cancellation, live precedence, cooldown, capacity, isolation and title/context merging are covered.
- `node --test test/opencode*.test.js test/server-route-state.test.js`: **369 passed, 0 failed**.
- `npm run verify:electron`: Electron 41.10.4 verified.
- `npm test`: **9760 passed, 10 failed, 41 skipped** (9811 total). The failures exactly match the previously reproduced `test/session-renderer-behavior.test.js` baseline failures; unrelated renderer files remain unchanged.
- Real OpenCode CLI **1.18.30** on macOS, imported synthetic history and real Clawd HTTP/state modules: both scheduling sequences leave context absent on `e269ce4d` and produce **320 / 1000 (32%)** after the revision. The early-bootstrap case injects one controlled `session.status`; the capacity case uses one real host `tui.session.select`, with the first four SDK reads held and then released. No new assistant event or model call; peak history concurrency is four.
- Actual Dashboard HTML/preload/renderer in hidden Electron displays **Context: 320 / 1,000 (32%)**, with no renderer errors. Only context metadata fields changed in the existing session.

## Remaining acceptance / limits

The original **Windows 11 ARM64 / OpenCode Desktop 1.18.18 packaged-app** recovery scenario and its actual event sequence remain a separate acceptance item. Node/module tests and imported-fixture process checks do not validate that Desktop host. Older sessions outside the startup window need a supported event; this does not add lifecycle discovery for hosts that emit no session activity.

### DIFF
--- a/hooks/opencode-family-plugin/core.mjs
+++ b/hooks/opencode-family-plugin/core.mjs
@@ -70,6 +70,13 @@ const CONTEXT_LIMIT_CACHE_MS = 60 * 1000;
 // tracked, so long-lived hosts (days of sessions) stay bounded even though
 // _lastStatePerSession itself is intentionally unbounded.
 const MAX_CONTEXT_USAGE_ENTRIES = 1024;
+// Resume hydration reads one recent page. Bootstrap examines at most 20
+// directory-owned root sessions; it never creates Clawd lifecycle activity.
+const CONTEXT_HISTORY_LIMIT = 20;
+const CONTEXT_BOOTSTRAP_LIMIT = 20;
+const CONTEXT_HISTORY_TIMEOUT_MS = 2000;
+const CONTEXT_HISTORY_RETRY_MS = 30 * 1000;
+const CONTEXT_HISTORY_MAX_CONCURRENT = 4;
 // Fire-and-forget: the IIFE never blocks the event hook's return value, so a
 // generous timeout is safe. 200ms was too tight when Clawd's IPC roundtrip
 // (main → renderer → main) ran under load and silently timed out.
@@ -726,7 +733,7 @@ export function createOpencodeFamilyPlugin(config) {
     return sessions || null;
   }
 
-  function getContextState(instanceToken, sessionId, create = false) {
+  function getContextState(instanceToken, sessionId, create = false, activityObserved = true) {
     const normalized = normalizeSessionId(sessionId);
     if (!normalized) return null;
     const sessions = getContextSessionMap(instanceToken, create);
@@ -739,9 +746,18 @@ export function createOpencodeFamilyPlugin(config) {
         latestSequence: 0,
         inFlight: new Map(),
         delivered: null,
+        liveRevision: 0,
+        hydration: null,
+        nextHydrationAt: 0,
+        activityObserved,
       };
       sessions.set(normalized, state);
+      if (sessions.size > MAX_CONTEXT_USAGE_ENTRIES) {
+        const oldest = sessions.keys().next().value;
+        if (oldest !== normalized) sessions.delete(oldest);
+      }
     }
+    if (state && create && activityObserved) state.activityObserved = true;
     return state || null;
   }
 
@@ -760,7 +776,9 @@ export function createOpencodeFamilyPlugin(config) {
       ? _contextStateByInstance.values()
       : [getContextSessionMap(instanceToken)].filter(Boolean);
     for (const sessions of maps) {
-      for (const sessionId of sessions.keys()) ids.add(sessionId);
+      for (const [sessionId, state] of sessions) {
+        if (state.activityObserved) ids.add(sessionId);
+      }
     }
     return ids;
   }
@@ -795,7 +813,12 @@ export function createOpencodeFamilyPlugin(config) {
       ..._lastStatePerSession.keys(),
       ..._sessionParentById.keys(),
       ..._sessionParentById.values(),
-      ..._statePostTailBySession.keys(),
+      // A bootstrap-only metadata queue cannot create a session and must not
+      // manufacture SessionEnd ownership during a global disposal.
+      ...[..._statePostQueueBySession].filter(([, queue]) => (
+        (queue.active && queue.active.body.metadata_only !== true)
+        || queue.pending.some((entry) => entry.body.metadata_only !== true)
+      )).map(([sessionId]) => sessionId),
     ]);
     for (const sessionId of collectContextSessionIds()) ids.add(sessionId);
     const root = normalizeSessionId(_rootSessionId);
@@ -925,7 +948,7 @@ export function createOpencodeFamilyPlugin(config) {
   // Snapshot mutable request data synchronously. session.deleted cleanup runs
   // immediately after enqueue, so cwd/title/process fields must already be
   // frozen before a queued delivery waits behind an older request.
-  function snapshotPost(body, logTag) {
+  function snapshotPost(body, logTag, contextDirectory) {
     const outbound = { ...(body || {}) };
     // Enrich every outbound body with process-tree fields. Cached after first
     // call so this is just a few object assignments per POST.
@@ -941,7 +964,9 @@ export function createOpencodeFamilyPlugin(config) {
     // walk finds no terminal to report.
     const orcaPaneKey = orcaPaneKeyFromEnv();
     if (orcaPaneKey) outbound.orca_pane_key = orcaPaneKey;
-    const cwd = resolveSessionDirectory(outbound.session_id);
+    const cwd = contextDirectory === undefined
+      ? resolveSessionDirectory(outbound.session_id)
+      : { directory: contextDirectory, source: "context-instance" };
     if (cwd.directory) outbound.cwd = cwd.directory;
     else delete outbound.cwd;
     outbound.agent_pid = process.pid;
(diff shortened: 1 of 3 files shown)


hooks/opencode-family-plugin/core.mjs
```diff
@@ -1641,6 +1669,132 @@ export function createOpencodeFamilyPlugin(config) {
...
+      instance.historyRequests.delete(controller);
+    }
+  }
+
+  function hydrateContextSession(rawId, instance, activityObserved = true) {
+    const sessionId = normalizeSessionId(rawId);
+    const state = getContextState(instance.instanceToken, sessionId, true, activityObserved);
+    if (state.hydration || state.inFlight.size > 0
+      || (state.delivered && Number.isFinite(state.delivered.limit) && state.delivered.limit > 0)
+      || Date.now() < state.nextHydrationAt
```
rullerzhou-afk: **[P2] Preserve a retry when bootstrap metadata reaches Clawd before the session exists**

`nextHydrationAt` is set before the history read, including when the read succeeds but Clawd cannot accept the metadata yet. A reproducible startup sequence is:

1. Bootstrap lists session S, reads a valid assistant sample, and posts its context while S is absent from Clawd.
2. The real state route returns 204 without `X-Clawd-Metadata-Accepted`; `state.delivered` correctly stays null.
3. Within 30 seconds, `session.status` arrives for S and creates its Clawd lifecycle record. This guard discards the hydration attempt.

No retry is scheduled when the cooldown expires. With no further supported event, S remains visible with no context, which retains the #883 symptom. The existing retry test only succeeds after advancing the clock **and emitting another event**, so it misses this sequence.

Please retain the undelivered sample or a bounded retry intent and service it after the lifecycle makes delivery possible. An SDK failure cooldown should not silently lose the resume signal. Any replay/retry still needs the instance/generation/live-revision fences and must not create a missing session itself.

Regression: start with a valid bootstrap history and no Clawd session, verify the initial metadata is unaccepted, then emit the first lifecycle event inside the cooldown. Context must become available without another assistant update or an extra manually emitted resume event.

hooks/opencode-family-plugin/core.mjs
```diff
@@ -1641,6 +1669,132 @@ export function createOpencodeFamilyPlugin(config) {
...
+        && normalizeDirectoryOwnershipKey(info.directory) === normalizeDirectoryOwnershipKey(instance.directory)
+      )).sort((a, b) => (b.time?.updated || 0) - (a.time?.updated || 0)).slice(0, CONTEXT_BOOTSTRAP_LIMIT);
+      let cursor = 0;
+      await Promise.all(Array.from({ length: CONTEXT_HISTORY_MAX_CONCURRENT }, async () => {
+        while (current() && cursor < candidates.length) {
+          const info = candidates[cursor++];
+          // Live observation wins, including pending SDK work and zero-token
+          // updates. A list response is never evidence of a new generation.
+          const id = normalizeSessionId(info.id);
+          if (getContextState(instance.instanceToken, id)
```
rullerzhou-afk: **[P2] Keep capacity-blocked selections pending**

A resume/selection that arrives while all four history slots are occupied creates a context state at line 1712 and then returns at the concurrency guard on line 1716. No pending work is retained. When the bootstrap worker later reaches that session, this existence check skips it as though hydration or a usable live sample had already taken ownership.

I reproduced this with five directory-owned bootstrap candidates: hold the first four message reads, make the fifth session visible in Clawd, emit one `tui.session.select` for it, then release the four reads. Only the first four IDs are ever queried; the selected fifth session receives no context. The current concurrency test supplies an additional event after capacity becomes free, which hides the lost original selection.

Please retain a bounded, deduplicated pending hydration request and drain it when capacity is released, or make the bootstrap scheduler distinguish an observed-but-unserved session from an in-flight/delivered/live-authoritative one. Preserve cancellation and live-update fences when servicing that request.

Regression: the five-session sequence above should query and hydrate the selected session after a slot is freed, without another event and without exceeding four concurrent history reads.

rullerzhou-afk (changes requested):
Reviewed `e269ce4d`. There are two reproducible hydration scheduling gaps that leave a resumed session without context until another suitable event arrives. Please address the inline comments before merging.

Validation on Windows:

- Existing focused suite: `node --test test/opencode*.test.js test/server-route-state.test.js` — **344 passed, 0 failed, 1 skipped** (the POSIX file-mode test).
- An additional local reproduction harness exercises the plugin's Node runtime branch and the real `handleStatePost` / `state.js` modules, with synthetic OpenCode SDK history and an in-process fetch-to-route adapter. The normal event-driven hydration control passes; both reported regression cases fail. The same result reproduces outside the sandbox.

Please add regressions for both sequences, including cancellation on deletion/disposal and precedence of newer live context while requests are pending. Keep metadata-only delivery, the existing token formula, and the four-request bound.

The original Windows 11 ARM64 / OpenCode Desktop 1.18.18 packaged-app scenario remains a separate acceptance item. This review's Node/module tests do not validate that Desktop host or its event sequence.

hooks/opencode-family-plugin/core.mjs
Tsdsj: Addressed in e90d5974. The plugin now retains a compact unaccepted history sample and replays it behind the normal lifecycle POST, without another history read or waiting for the query cooldown. A lifecycle signal arriving while the original metadata POST is still pending is retained as well.

Added regressions for both timings in `test/opencode-family-hydration-scheduling.test.js`, using the Node plugin and real `handleStatePost` / `state.js`: the initial bootstrap POST receives an unaccepted 204 and creates no session; one `session.status` inside the cooldown then supplies context, with no additional resume or assistant event. Signals received during an SDK failure cooldown also wake once when it expires, without creating a recurring poll.

Deletion, recreation, both disposal paths and newer live context invalidate retained work. Historical metadata already queued behind a slow POST is checked again before sending, including zero-token live updates; merged title updates are preserved.

All 24 new scheduling tests and the full 369-test focused suite pass. The original Windows ARM64 Desktop packaged-app acceptance remains separate, as documented in the updated PR description.

hooks/opencode-family-plugin/core.mjs
Tsdsj: Addressed in e90d5974. Bootstrap and event-driven requests now use the same bounded scheduler. A selection retains one deduplicated pending intent on its instance/session state when capacity is full, and released slots drain it automatically. Explicit selections take priority over unserved bootstrap candidates; `session.list` and history reads share the four-slot bound.

The exact five-session regression is in `test/opencode-family-hydration-scheduling.test.js`: hold the first four reads, seed the fifth Clawd session, emit one `tui.session.select`, release the reads, and assert that the fifth session receives context without another event and with a peak of four. The old concurrency test now asserts automatic draining instead of sending a second event. Added cancellation, live/zero-token precedence, timeout, cross-instance isolation and bounded-retention cases.

I also repeated this with real OpenCode CLI 1.18.30, imported history and the real Clawd HTTP/state modules on macOS. The host emitted exactly one `tui.session.select`: `e269ce4d` queried only four IDs and left context absent; the revision queried the fifth and displayed 320 / 1000 (32%), with peak concurrency four. The original Windows Desktop packaged-app acceptance is still separate.

rullerzhou-afk (approved):
复审 e90d5974：此前要求修改的两项 hydration 调度问题均已修复，本次批准替代早上针对 e269ce4d 的请求修改。

验证：
- 原样保留早上的独立复现脚本，连同 OpenCode / state route 相关测试在 Windows 沙箱内外复跑：371 passed、0 failed、1 项 POSIX 平台跳过。
- Windows x64、官方 OpenCode Desktop 1.18.18、从当前提交构建的 packaged Clawd，使用隔离环境中的真实历史会话进行实际模型续聊。正常生命周期恢复后，历史用量 8,712 / 200,000 在 07:18:46.017 UTC 被 Clawd 接收；新一轮首条非零 assistant 用量更新到 07:19:04.728 UTC 才到，提前约 18.7 秒。之后新的实时用量 8,777 / 200,000 也正常接收。没有合成生命周期事件或替换 SDK / HTTP / 模型路径；接收结论来自真实打包应用日志。

范围：新 assistant 初始化时存在全零事件，不宣称没有任何新 assistant 事件。仅打开历史页而不产生会话活动时，Desktop 没有发出可用恢复事件；这属于 PR 已声明不新增 lifecycle discovery 的范围。原 Windows ARM64 环境仍未验证。本次批准针对既定 hydration 范围与已验证的 x64 续聊路径。

rullerzhou-afk:
已合并，感谢 @Tsdsj 的贡献，也感谢你认真跟进这轮 review，补齐调度修复和回归测试！

我们已完成 Windows x64 / OpenCode Desktop 1.18.18 的真实续聊补测：旧会话恢复活动后，历史上下文会在新回复完成前被 Clawd 接收。原 ARM64 场景仍保留为未覆盖的验证项；仅浏览历史页而没有会话活动的情况，按本 PR 已声明的范围限制记录。

PR #1010, opened 2026-09-11
feat(recovery): durable session history so a reboot no longer loses a conversation

TalexDreamSoul:
## Problem

After an unexpected restart there is no way to find the conversations that were open. Resuming today means remembering a session UUID and typing it into `showResumeInput()` (`src/main.js:4058`) by hand, plus re-picking the right folder — and a restart is exactly when you have neither.

**The evidence already exists, and the recovery path deletes it.**

`~/.clawd/session-recovery-v1` already records session id, cwd, title and state for every live Claude Code session. But `loadActiveRecoveryLeases()` does this (`hooks/session-recovery-lease.js:619`):

```js
if ((record.pid && !processAlive(record.pid, options)) || …) {
  … fs.unlinkSync(filePath);   // ← the row is erased
  continue;
}
```

After a reboot every recorded PID is dead, so the **first launch after a restart unlinks every lease** — precisely the rows a user needs at that moment.

## Why this is a new store, not a change to the lease

That deletion is correct for what a lease *is*. A lease answers *"is this still running?"* and must fail closed — the PID and `processStartIdentity` checks are what stop a recycled PID from resurrecting a phantom session (`887c8cfe` restores after a **Clawd** restart, not an OS one).

Recall is the opposite requirement: it has to **outlive** the process it describes. Loosening the liveness checks to serve it would trade a real safety property for a feature. So this adds a sibling store instead; `session-recovery-v1` is untouched and its 25 tests still pass.

| | `session-recovery-v1` | `session-history-v1` (new) |
|---|---|---|
| Question | Is it still running? | What was I working on? |
| On dead PID | row deleted | row kept — that's the point |
| Failure mode | closed | open |
| Retention | 24h / liveness | 30d / 200 rows |

## Changes

**`hooks/session-history.js`** — one row per session, written from the same hook event that already updates the lease. Pointers only: session id, cwd, title, timestamps. No prompts, responses, or tool output. Same `0700`/`0600` as the lease dir, same strict key allow-list and filename binding.

**Reuses the existing arbiter.** `classifyBody` is exported as `classifyStateBodyForRecovery` rather than deciding "still running" a second, divergent way — AGENTS.md is explicit about not writing a parallel arbiter.

**Boot identity with no spawned process.** `os.uptime()`, compared with tolerance instead of equality (whole-second uptime plus NTP steps make exact equality wrong). The Windows lease writer carries an explicit *"never spawn a second PowerShell here"* rule and this shares its path, so `sysctl`/`wmic` probes were not an option. **A row that never recorded an ending, last seen under an earlier boot, is what `interrupted` means** — no separate crash flag needed, because a machine that dies gets no chance to write one.

**Debounced writes.** Hook traffic is tens of events per task (AGENTS.md notes 40+ for QwenWork). Writes are skipped unless identity changed or 30s elapsed — an O(1) overwrite per session, never a per-event append.

**`src/session-history-loader.js`** — resolves the resume target in main, reading `cwd` **back from the store** so a renderer can never choose the folder a session is relaunched in. It also probes Claude Code's own transcript at `~/.claude/projects/<encoded cwd>/<id>.jsonl`, and **fails open**: that layout is Claude Code's private detail, so only a present project dir with the file absent is a confident "gone" — unknown still gets offered, because hiding a resumable session is the worse error.

**Dashboard** grows a "Recent sessions" section, shown most usefully on the *empty* dashboard — the state a machine comes back up in. One click resumes through the existing `buildClaudeArgs` `--resume` path (`src/launch-claude.js:165`), which is already allow-listed by `normalizeClaudeSessionId()`. With no history the empty state keeps its exact previous node tree.

**IPC.** Both channels are restricted to the trusted Dashboard main frame, the same way the Kimi quota capability is. `dashboard:resume-session` accepts *exactly* an `agentId`/`sessionId` pair — no `cwd`, and deliberately **no mode field**, so `--dangerously-skip-permissions` has no route in from the renderer; that stays behind the pet menu flow that confirms it explicitly.

## Verification

**Full suite: 9875 tests / 9818 pass / 13 fail.** Clean `main` on the same machine: **9840 / 9783 / 13**. Same 13 pre-existing failures (`focus-windows.test.js`, `session-renderer-behavior.test.js` — the latter's fake DOM has drifted from the renderer's node tree; unrelated to this PR and not touched by it). **+35 tests, all passing, zero new failures.**

The central test asserts the actual regression:

```
✔ keeps the row that loadActiveRecoveryLeases erases after a reboot
```

— writes a lease *and* a history row for a dead PID, confirms `loadActiveRecoveryLeases()` returns nothing and deleted its file, and confirms the history row survives and reads `interrupted: true`.

Also covered: debounce and its bypasses, clean ending then resume clearing `endedAt`, out-of-order events, headless/foreign-agent/WSL refusal, age and count pruning, foreign/misfiled/oversized record rejection, transcript probe present/missing/unknown, the resume list hiding already-live sessions, the renderer sending only the id pair, and the render tick never reaching disk.

## Notes for review

- **Claude Code only** for now. `SUPPORTED_AGENT_IDS` is `{claude-code}`, which is also the only agent `buildClaudeArgs` can `--resume`. Rows are agent-tagged so others can follow.
- `getSessionHistory` / `resumeSessionFromHistory` are **required** deps on `registerSessionIpc`, matching the existing `requiredDependency` convention — I mis-wired them into `registerSettingsIpc` while building this and an optional dep made it fail silently, so a hard throw seemed right.
- Retention (30d / 200), the debounce window (30s), and the boot tolerance (5m) are judgement calls, easy to move.
- Resume is never automatic. Relaunching N sessions on login would spawn N terminals unbidden; the list always waits for a click.


rullerzhou-afk:
Thank you, @TalexDreamSoul! Separating durable session history from live recovery leases is the right direction: finding an old conversation should not weaken the checks that decide whether a process is still alive. Your original commit and authorship are preserved.

I pushed a follow-up in aebad64ba156d2399e872c9e2945821f778bfacb to harden the implementation:

- Serialize history updates and retention under per-record locks, preserve invalid/foreign records, and make terminal events win same-timestamp races.
- Share the launcher's session-ID validation; recheck the installed/enabled gate and local-live identity in main; honor `CLAUDE_CONFIG_DIR` and distinguish an unreadable transcript from a missing one.
- Coalesce duplicate Resume requests across Dashboard reloads. A terminal launch is only **submitted**, not proof of a resumed session; the existing hook/state path confirms it. Failures and confirmation timeouts keep an explicit manual retry path.
- Keep Resume visible on the empty Dashboard, close stale-history refresh races, and add regression coverage and architecture documentation.

Validation of this source revision:

- **PASS — macOS focused tests:** 153 passed, 0 failed, 4 Windows-only cases skipped (157 total).
- **PASS — native Windows focused tests:** the same 157 cases passed, with no failures or skips, on a hash-verified copy of the source.
- **PASS — real macOS source-app/Claude round trip:** a real interactive conversation was recorded, Claude exited, Clawd was restarted, and the actual Dashboard Resume action launched exactly one `claude --resume` session even after a double-click. Real hooks brought the session back into the live list. A follow-up question that did not contain the earlier test token returned that token from the resumed conversation.
- **Known baseline failures — full local suite:** 9,894 total / 9,823 passed / 27 failed / 44 skipped. The failing test-name set exactly matches the original PR and the relevant local-main baseline tests (DSH/doctor, Windows focus, and session-renderer behavior); the 19 added regression cases pass. This is not a claim that the full suite is green.
- **NOT TESTED:** an actual OS reboot/power loss, Windows or Linux GUI conversation-resume flows, or a packaged/signed-app end-to-end resume flow. Windows unit tests and the macOS app restart should not be read as those results.

The feature remains a local, manual Claude Code resume index, not a transcript backup or automatic recovery. Thanks again for the thoughtful foundation!

Merge update: all 13 checks passed for the accepted head `aebad64ba156d2399e872c9e2945821f778bfacb`: [Repository Asset Audit](https://github.com/rullerzhou-afk/clawd-on-desk/actions/runs/34673902408), [six-platform/runtime compatibility jobs](https://github.com/rullerzhou-afk/clawd-on-desk/actions/runs/34673902403), and [unit plus five-target package audit](https://github.com/rullerzhou-afk/clawd-on-desk/actions/runs/34673902420). The exact integration tree with main `1b48833` was also tested locally: 9,909 total / 9,838 passed / the same 27 baseline failures / 44 skipped. Merged as `7efe5db6340bac7ad1324d74ee2c82742253caa9`, preserving both commits; its tree matches that tested integration tree. The GUI/reboot acceptance limits above still apply.

PR #1013, opened 2026-09-13
feat(omp): add OMP (oh-my-pi) as a first-class state-only agent

TalexDreamSoul:
## Why

Clawd ships built-in integrations for two dozen CLI agents but not **OMP** (oh-my-pi). OMP users today have to register it by hand as a *custom application* and run a community bridge ([Crosery/clawd-on-desk-omp](https://github.com/Crosery/clawd-on-desk-omp)) that POSTs to the loopback `/state` endpoint. That bridge works — it is explicitly built to need no changes here — but it means OMP is the one agent that needs a manual setup dance.

This makes OMP a registered agent like any other: Settings → Agents, auto-sync, Doctor coverage, cleanup on uninstall.

## Shape

OMP is a fork of the agent Pi is built on (`@oh-my-pi/pi-coding-agent` vs `@earendil-works/pi-coding-agent`), so the integration follows `agents/pi.js` deliberately closely.

**`hooks/omp-extension.ts` differs from `hooks/pi-extension.ts` by exactly three lines** — the package import, the core import, and the exported function name:

```
9c7   < import coreModule from "./omp-extension-core.js";
      > import coreModule from "./pi-extension-core.js";
320c318 < export default function clawdOmpExtension(omp: ExtensionAPI): void {
        > export default function clawdPiExtension(pi: ExtensionAPI): void {
```

The process-tree walk, terminal detection and loopback POST are the same problem for both, and letting two copies drift apart is how one of them quietly goes stale. Everything OMP-specific is in `hooks/omp-extension-core.js`.

## Three behaviours that differ from the Pi core

Each is a thing OMP taught the community bridge the hard way, and each is covered by a test.

**1. Completion binds to `session_stop`, never `agent_end`.** OMP fires `agent_end` at *every* agent-loop boundary — including scheduling pauses with background jobs still running, queued follow-ups, and settles that left tool calls in flight. Reporting completion from it makes Clawd play the finish chime while the session is still working. `session_stop` is the settled turn: it never fires for task/subagent sessions and defers until agent-owned background jobs are idle.

> Worth flagging for a separate look: `hooks/pi-extension-core.js` binds `agent_end → Stop → attention`. If Pi's `agent_end` has the same semantics as OMP's, Pi likely has this bug too. I have not verified that against Pi and have not touched its core here.

**2. `session_switch` / `session_branch` are reported, and the session being left is retired.** OMP can move an interactive terminal to another conversation with no shutdown for the old one. Without a synthetic `SessionEnd`, Clawd keeps a live row for a session nothing will ever report on again.

**3. A `session_title` is always sent.** Several interactive OMP sessions legitimately share one working directory, and Clawd's folder-name fallback then renders every row — and every jump target — identically. OMP's own session name is preferred; the folder is the fallback.

## The community bridge

The installer **fails closed** when `~/.omp/agent/extensions/clawd-on-desk-omp.ts` is already present. Both report the same lifecycle events, so installing alongside it would make OMP load both and POST twice per event. Clawd leaves the bridge in place, skips its own install, and says why.

## Icon

`assets/source/agent-icons/omp.png` is a **neutral placeholder** — a pi glyph on a tile — recorded as `fallback: true` with `exportMode: "passthrough"`, the route `assets/source/agent-icons/README.md` sanctions when no official source exists yet. **It is not OMP brand artwork.** I did not want to appropriate the bridge project's logo or invent provenance for a manifest that records upstream name, URL and licence. Please replace it once licensed artwork is available; `omp` is added to the fallback allowlist in `test/state-agent-icons.test.js` so the swap is a reviewed change.

## Touchpoints

Registry, recap metrics policy (explicit per-agent, per the AGENTS.md rule), prefs defaults, server agent-id map, settings order + action allowlists, integration sync, installation detector, Doctor descriptors, cleanup-integrations, icon manifest, `package.json` scripts, AGENTS.md constraint, and the agent lists across all six READMEs.

`hooks/omp-*.js` are deliberately **not** added to `HOOK_FILES` — Pi's are not either; this is a local extension install, not a Remote SSH hook payload.

## Verification

**Full suite: 10028 tests / 9969 pass / 13 fail.** Clean `main` on the same machine: **9987 / 9928 / 13**. Same 13 pre-existing failures (`focus-windows.test.js`, `session-renderer-behavior.test.js`), **+41 tests, all passing, zero new failures.**

Installer verified end to end against a temp tree: clean install writes `index.ts` + core + marker; re-running reports `updated: false`; a present community bridge blocks it; an unmanaged directory is never clobbered; Pi's marker is not accepted as OMP's and vice versa; uninstall removes only Clawd-managed directories.

One note on how I found the recap requirement: adding an agent to the registry without a `src/recap-metrics.js` policy throws at module load and took **504 tests** with it. The AGENTS.md rule about explicit per-agent recap policies is doing real work — worth knowing if anyone adds agent number 26.

## Not included

- Per-language integration bullets exist only in `README.md` and `README.zh-CN.md`; the other four get the agent-list mention. I would rather leave a gap than put machine translation in your README — happy to add them if you tell me the wording you want.
- OMP stays state-only, like Pi: no permission layer, no bubbles, `permissionApproval` and `interactiveBubble` both false.


TalexDreamSoul:
@rullerzhou-afk when you have a slot — CI is green across all 17 checks (incl. Windows x64/ARM64 and macOS x64/ARM64 packaging, Wayland/XWayland smoke). Local suite matches main exactly: 13 failures before and after, +41 new tests all passing.

Two things in here are yours to decide rather than mine, so I'd rather flag them than have you find them:

**1. The icon is a placeholder.** `assets/source/agent-icons/omp.png` is a neutral pi glyph recorded as `fallback: true`, not OMP brand artwork. I didn't want to reuse the community bridge's logo or invent provenance for a manifest that records upstream name/URL/licence. `omp` is in the fallback allowlist in `test/state-agent-icons.test.js`, so swapping in licensed artwork is a reviewed one-line change.

**2. Pi may have the same `agent_end` bug this PR avoids.** OMP fires `agent_end` at every agent-loop boundary — including scheduling pauses with background jobs still running — so binding completion to it plays the finish chime mid-turn. This PR binds `session_stop` instead. `hooks/pi-extension-core.js` currently binds `agent_end → Stop → attention`. **I have not verified whether Pi's `agent_end` has the same semantics**, so I deliberately left Pi's core alone. If it does, that's a separate fix and I'm happy to send it.

No rush on any of it — and if the scope is more than you want in one PR, I can split the docs or the registration wiring out.

rullerzhou-afk (changes requested):
Thanks for the careful, well-scoped work here. I reviewed exact head `72190d1e293a21f26b9f2cd7886edce455829000` against current `main` (`782edefc7ebff70d462f326f1354451d1b86ffa3`) and checked the current upstream OMP contracts. The overall direction makes sense and all 17 GitHub checks are green, but I found several correctness and test-safety gaps that need fixing before merge.

1. **Blocker — the test suite writes to the developer's real `~/.omp`.** `test/server-hook-management.test.js` does not provide `syncOmpExtensionImpl`. Its missing-fake Proxy is spread into a plain object in `src/server.js`, so the guard never sees the absent key and startup calls the real `registerOmpExtension()`. Running `npm test` on this head actually created `~/.omp/agent/extensions/clawd-on-desk/{index.ts, omp-extension-core.js, .clawd-managed.json}` on my machine; I restored the original state immediately. Please add the OMP fake and expected calls, and add a regression guard proving this test cannot touch the real home/config tree.

2. **Blocker — Doctor does not handle OMP at all.** The descriptor uses `configMode: "omp-extension"`, but `checkAgent()` only dispatches `"pi-extension"`. A healthy managed OMP fixture deterministically returns `manual-only / info / Unsupported config mode: omp-extension` with no Fix action; broken and foreign-marker states are equally invisible. Please generalize the Pi extension checker using the descriptor's integration id (the current Pi marker validator is hard-coded to `integration === "pi"`) and add healthy/missing/incomplete/foreign OMP Doctor cases.

3. **High — the installer hard-codes only the default OMP agent directory.** `DEFAULT_PARENT_DIR` is always `~/.omp/agent`, while current OMP resolves user extensions through the active agent directory: named profiles use `~/.omp/profiles/<name>/agent/extensions`, and the default profile also honors `PI_CONFIG_DIR` / `PI_CODING_AGENT_DIR` (plus `OMP_PROFILE` / `PI_PROFILE`). As written, Settings can report a successful install that a profile session never loads. Please make install/detect/Doctor/cleanup share one OMP-aware resolver, or explicitly scope and document the integration as default-profile-only.

4. **High — a community bridge added after the managed extension still produces duplicate reporting.** If Clawd installs first and `clawd-on-desk-omp.ts` appears later, the next sync returns `standalone-bridge-present` but leaves Clawd's existing managed directory in place, so OMP loads both. Please handle that ordering explicitly: remove/disable only the verified Clawd-managed copy or surface a hard actionable conflict, while preserving foreign files. Add the built-in-first-then-bridge regression case.

5. **Lifecycle ordering — shutdown does not drain the old session's synthetic `SessionEnd`.** Deliveries are chained per session. On A→B switch, A's synthetic end and B's shutdown live on different chains; with A's post deliberately blocked, B's awaited `session_shutdown` resolves first. An immediate process exit can therefore strand A. Please serialize the transition globally or make shutdown drain all outstanding lifecycle tails, and cover slow-old-end + immediate-shutdown.

6. **Adjacent gaps:** `agents/omp.js` says POSIX package-path startup detection exists, but `src/state.js` only adds Pi needles; OMP therefore has no POSIX startup keep-awake fallback when only OMP is enabled. Also add `hooks/omp-extension-core.js` to both Orca validator / `NESTED_TERMINAL_ENV` drift-guard lists in `test/focus-orca.test.js`, and update the setup/runtime docs that currently document Pi but not OMP.

Evidence boundary: the focused OMP-related suite passed **304/304**; exact base/head full-suite runs showed no new failure delta in this environment, but the suite is not safe while item 1 remains. A real OMP interactive smoke is **NOT TESTED** here because no `omp` executable is installed. Current upstream source does confirm `session_stop`, the exported extension types, and directory `index.ts` loading, so I am not challenging those contracts.

GitHub currently reports this PR as conflicted with current `main`; please update the branch after addressing the findings above.

TalexDreamSoul:
Addressed all six findings plus the conflicts in 4 commits on `feat/omp-agent` (head `44dc41ac`). `mergeable` is now `true` (was `dirty`); CI is green across all 17 checks, including the Windows/macOS x64+ARM64 packages and the Wayland/XWayland smokes.

**1. Blocker — the suite wrote a real `~/.omp`.** Reproduced exactly as you described: `HOME=$(mktemp -d) node --test test/server-hook-management.test.js` created `$HOME/.omp/agent/extensions/clawd-on-desk/{index.ts,omp-extension-core.js,.clawd-managed.json}` with **55/55 passing**.

The root cause is one level above the missing fake. `src/server.js` hands the runtime `{...ctx}`, and a spread copies only **own enumerable** properties — so *every* seam this fixture answered virtually was dropped before integration-sync read it. `typeof ctx.syncOmpExtensionImpl === "function"` turned false, the real installer ran, and the missing-fake tripwire never fired for anything.

The guard now reads the seam list from `integration-sync.js`'s own source and exposes it as own enumerable properties, so the spread carries the tripwire through and a new agent cannot be added without a fake. That immediately named **three** un-faked seams rather than one — `syncOmpExtensionImpl`, `syncZcodeHooksImpl`, `syncDeepSeekHarnessPluginImpl` — all three had been silently running their real installers on every run of this file. All three now have fakes and appear in the expected sync order (which also changes the reference arrays: `zcode`, `deepseek-harness`, `omp` inserted at their real positions, and `omp` in the disabled-agent case since OMP is not in that `Set`).

Evidence: with only the guard in place the file fails naming all three and writes nothing; with the fakes it is 55/55. The full suite under a redirected `HOME` with a fake `omp` on `PATH` writes nothing at all.

**2. Blocker — Doctor had no OMP path.** The checker is now shared by both extension modes, and the integration id comes from the descriptor's own config mode (`"omp-extension"` → `"omp"`) instead of a hard-coded `"pi"` — OMP writes `integration: "omp"`, so a Pi-only validator reported a healthy install as needs-review forever. The incomplete-files wording is now agent-aware (`${agentName} extension files are missing or incomplete`) rather than always "Pi". Added healthy / missing-dir / incomplete / foreign-marker cases: **all four fail on the previous code** (`Unsupported config mode: omp-extension`, no Fix) and pass now. The cross-marker case is bidirectional: an OMP-mode descriptor on a Pi marker is `needs-review` with no Fix, and vice versa.

**3. High — hard-coded directory.** `hooks/omp-install.js` now owns the resolution and install, uninstall, cleanup, the installation detector and the Doctor descriptor all go through it. I mirrored OMP's own resolver (read out of the installed `@oh-my-pi/pi-coding-agent` bundle) including the two details that are easy to get wrong: a profile **beats** `PI_CODING_AGENT_DIR` (OMP returns no `agentDirOverride` while a profile is active), and `PI_CONFIG_DIR` is always joined *under* the home directory even when it looks absolute. Profile names are validated with OMP's slug rules and an invalid name means "no profile", exactly as OMP treats it.

On the second half of your ask: rather than silently claim full coverage, Clawd manages the one directory it resolves and says so — Doctor appends the profiles it does **not** manage (`unmanagedOmpProfiles`) instead of a bare "verified", and it is in `known-limitations.md` in both languages. If you would rather it installed into every existing profile, that is a bigger change (per-directory failure matrix: bridge in one profile, foreign marker in another); I can do it as a follow-up if you want that behaviour.

**4. High — bridge added after the managed extension.** Handled explicitly in both orders: when the bridge is present the installer now removes only a copy its own marker proves it wrote, and leaves a foreign directory untouched. Both orders have a regression test.

**5. Lifecycle ordering.** `session_shutdown` now drains every outstanding tail (`drainDeliveries`), read after its own enqueue so it also covers a chain that was empty a moment earlier; `session_stop` keeps awaiting only its own chain, since the session continues running. The regression test fails on the previous code with exactly your symptom ("shutdown must not resolve while another chain is still owed") and passes now.

**6. Adjacent.** OMP has the POSIX startup needle `agents/omp.js` claimed (the **scoped** `@oh-my-pi/pi-coding-agent` path — the unscoped Pi needle also matches an OMP process, and OMP keeps helper processes on that path, so the comment records that it stays a bounded launch-time keep-awake). `hooks/omp-extension-core.js` was added to both Orca drift-guard lists in `test/focus-orca.test.js`. Setup guide, state mapping, known limitations and the runtime architecture doc now cover OMP next to Pi in both languages.

**Pi's `agent_end`:** left alone, as you said — I did not want to change Pi's core from an OMP PR. Happy to send it as a separate fix if you confirm the semantics.

### Update on your evidence boundary: a real OMP interactive smoke now passes

You noted a real OMP smoke was NOT TESTED because no `omp` executable was installed. One is installed here, so I ran it: extension installed into a throwaway `HOME`, a stub Clawd ingress on 23334 (23333 is a live Clawd on this machine, and I was not willing to inject a phantom session into it), then a real interactive `omp` session.

- The extension **loads** from `~/.omp/agent/extensions/clawd-on-desk/index.ts` in a real session and posts to the loopback ingress.
- `session_start` → `SessionStart` / `idle`, with a real payload: `agent_id: "omp"`, `hook_source: "omp-extension"`, `session_id: "omp:01a0a58f-…"`, **`session_title: "OMP · tmp"`** (the always-send-title behaviour firing on a real session), plus `agent_pid`, `cwd` and `pid_chain`.
- `session_shutdown` → `SessionEnd` / `sleeping` on quit, and the awaited delivery settled before the process exited — the ordering property item 5 is about.

What that smoke does **not** cover, so it should not be read as blanket verification: `session_stop` (needs a real model — the throwaway `HOME` has no credentials, so the turn stops at "No model selected"), `session_switch` / `session_branch`, and the synthetic-`SessionEnd` retirement. Those rest on the unit suite (22 cases) only.

### Suite

Sequential runs against `782edefc`, both with `HOME` redirected:

| | tests | pass | fail |
|---|---|---|---|
| `782edefc` | 10165 | 10106 | 13 |
| `44dc41ac` | 10225 | 10166 | 13 |

+60 tests, +60 pass, and the failing sets are **identical** (the same 13 Dashboard / Direct Send / editor-parent-pid-window cases). Neither run wrote anything under its `HOME`. `npm run audit:assets` reports 0 errors (the tracked-tree warning-budget warning is pre-existing).

One caveat worth recording: an earlier pair of runs executed concurrently and produced 14 failures, one of which was `opencode-family session directory` — that test passes 15/15 in isolation on both revisions, so it was contention, not a regression. The numbers above are from the sequential runs.

rullerzhou-afk:
Thank you, @TalexDreamSoul, for the thoughtful OMP integration and for working through the review feedback so carefully. Your profile-aware installer, Doctor coverage, bridge ownership handling, lifecycle fixes, documentation, and regression tests made this a strong contribution. After a small maintainer follow-up for OMP 18.2.4 completion/profile edge cases, the final head passed all 17 checks and is now merged. Really appreciate the care and persistence here!

Issue #995, opened 2026-09-05
Proposal: optional external review command for auto-approved requests (downgrade-only)


chrono-meta:
Today, with Permission handling set to Auto-approve (or after "Allow for this session"), Clawd answers `allow` to every tool request on its own; #994 (draft: hold destructive tool requests for a human under auto-approve) carves out the destructive ones and hands them back to the card. Everything else still goes through with no second look — and when I leave an agent running, that long run of ordinary edits and commands nobody reviewed together is the part that worries me more than the `git push --force` #994 now holds.

Scoping first: sessions started with `--dangerously-skip-permissions` or `defaultMode: "auto"` never send `PermissionRequest` to Clawd at all (#163, #865), so nothing here can reach them — this is for people who run Auto-approve in Clawd.

I already run a review step of my own before an agent's work counts as done, and I would like Clawd to be able to consult a step like that at the moment it is about to auto-allow. This is the smallest shape I could find that does that without Clawd itself deciding anything.

## Proposal

An opt-in setting, empty by default, so nothing changes unless someone fills it in:

```
permissionAutomation.externalReview.command    # argv array, e.g. ["/usr/local/bin/my-gate", "--stdin"]; no shell
permissionAutomation.externalReview.timeoutMs  # default 5000
```

When set, and only when automation is about to answer `allow` by itself, Clawd runs the command with the request as JSON on stdin — `tool`, `toolInput`, `agent`, `sessionId`, `cwd` — and maps the exit code:

| exit | Clawd does |
|---|---|
| 0 | allow, as today |
| 1 | ask — the normal permission card, with the command's last stdout line shown as a note |
| 2 | ask, with the warning badge (a block from the gate is still shown to the human, never applied silently) |
| any other code, timeout, not found, not executable | ask — a broken gate must fail toward the human, never toward `allow` |

Two things keep this from changing how the app decides:

- **It can only downgrade.** The command can turn an automatic `allow` into `ask`; it can never turn `ask` or `deny` into `allow`, and it does not run on the destructive hold from #994 — that path stays human-only. The worst a bad gate can do is ask you more.
- **Clawd stays a relay.** It runs an executable the user configured and reads a number. No policy, no vocabulary, no network. It should sit behind the local-ingress authentication in #976, if I am reading that work right: the command runs locally, after the request has been accepted, on the same object.

[...]

PR #994, opened 2026-09-05
fix(permissions): hold destructive tool requests for a human under auto-approve

chrono-meta:
`bubble-format.js` already knows which commands are destructive — the comment above `IRREVERSIBLE_PATTERNS` says exactly what it is for:

```js
// Like the MCP relabel (#445), this never touches Allow/Deny semantics or the
// no-decision fallback — it only routes the human's attention to destructive
// decisions (force-push, publish, bulk delete, history rewrite).
```

That contract was right for #613, when every request still ended in a human clicking Allow. It is no longer enough: since #740 (`auto-tools` / `unattended`) and #749 (per-session grants), a request can be allowed before any human sees it.

## What goes wrong today

With **Permission handling → Auto-approve** (or a per-session grant), a Claude Code `Bash` request for `git push --force origin main` takes this path:

1. `server-route-permission.js:2017` — not a `PASSTHROUGH_TOOLS` name, so it reaches `showPermissionBubble(entry)`.
2. `permission.js:2246 maybeAutoApprovePermission` → `permission-automation-policy.js:414 evaluatePermissionAutomation`. `Bash` is in `CLAUDE_COMPATIBLE_TOOL_APPROVAL_NAMES`, the interaction is `tool-approval` with `allowDeny`, so the result is `AUTO_ALLOW`.
3. `resolvePermissionEntry(entry, "allow")` writes `{ "decision": { "behavior": "allow" } }` back to the hook. No bubble, no Telegram / Lark card, no badge — the badge lives in the renderer (`bubble-renderer.js:1455`) and the renderer is never reached.

The same is true for `rm -rf`, `npm publish`, `git reset --hard`, `gh repo delete`, `terraform destroy`, and for the sweep that runs after a user clicks *Allow for this session* (`session-automation-coordinator.js resolveIfAllowed` → `permission.js:2166 canAutoResolvePendingPermission`, which calls the same policy function).

Two existing fixtures illustrate the gap: `test/permission-auto-approve.test.js` and `test/permission-slack-announce.test.js` both used `toolInput: { command: "rm -rf /" }` as the canonical *ordinary tool request that automation should allow*.

## Why hold (defer to the human) instead of deny, and why it rides `detectIrreversible`

[...]

PR #1021, opened 2026-09-15
feat(permissions): opt-in reminder that holds destructive requests under automation

chrono-meta:
Follows #995 and your direction there. This is the narrower, default-off built-in reminder, as a
fresh PR rather than a reopen of #994. The real-machine run you asked for is recorded at the end.

The setting is `destructiveActionReminder`, boolean, `default: false`. When it is on and permission
automation is about to allow an ordinary tool request by itself, a recognized destructive command
waits for a human through the approval channels that are already there, and the card says which
pattern matched. When it is off, nothing reads it.

It can only downgrade. `evaluatePermissionAutomation()` gained one optional input:

```js
// src/permission-automation-policy.js
    && interaction.capabilities.allowDeny
  ) {
    // The one place the reminder applies: an ordinary tool request that
    // automation would have allowed by itself.
    return reminderHold === true
      ? AUTOMATION_ACTION.DEFER
      : AUTOMATION_ACTION.AUTO_ALLOW;
  }
```

That is the whole authority: `AUTO_ALLOW` becomes `DEFER` on one branch. It cannot reach `DEFER`
requests, it never produces a deny, and the question-answer and plan-review branches below it are
untouched — a caller that does not pass the flag gets today's behavior, which is why the existing 22
policy tests pass unmodified.

## The three things your scope note called out

**Pre-display-preview input.** You were right not to let me assume `entry.toolInput`. It is
`truncateDeep(rawInput)`, so every string over `PREVIEW_MAX` (500) is already cut, and a command
whose destructive part sits past that point is not in it. The reminder therefore reads a bounded view
derived from the accepted input, next to where the display view is derived:

```js
// src/server-route-permission.js — at each of the 7 accepted-request sites
        const toolInput = truncateDeep(rawInput);
        const permissionDetail = preparePermissionDetail(toolName, rawInput);
        const permissionReminder = preparePermissionReminder(toolName, rawInput);
```

Supported shapes and budget, stated rather than implied: one command from
`command` / `CommandLine` / `Command` / `cmd` / `script`, in that order (the same order and field
names the existing hint uses) — either a **string** or an **array of strings**, which is joined with
single spaces, since that is the text a shell would have received rather than a guess. Capped at
`SCAN_MAX` = 4096 characters, `SEGMENT_MAX` = 50 shell segments, and `ARGV_MAX` = 256 array entries.
Anything that is not a command line (a number, an object, a nested array) yields no scan input. Past
the budget is a documented miss with a test that says so.

**One correction to the shared matcher, and it is the reason I would not have shipped this without
the pass.** `SHELL_TOOLS` in `bubble-format.js` listed `bash`, `shell`, `run_command`, `exec`,
`run_terminal_cmd` — but `execute_bash`, `powershell` and `run_shell_command` are all in
`CLAUDE_COMPATIBLE_TOOL_APPROVAL_NAMES`, i.e. names automation will allow on its own. A force-push
sent under one of those was never scanned:

```
execute_bash         auto-tools=auto-allow  reminder=null   <-- automatic allow
powershell           auto-tools=auto-allow  reminder=null   <-- automatic allow
run_shell_command    auto-tools=auto-allow  reminder=null   <-- automatic allow
bash                 auto-tools=auto-allow  reminder={"hold":true,"tag":"force-push"}
```

The three names are added, which also means the display hint now badges them — a small improvement to
#613 rather than a regression, but it is a behavior change on your merged feature and I would rather
name it than slip it in. To stop the class from recurring, the test derives the eligible-tool list
from the policy's own set and fails if a shell-shaped name there is not scanned; the only exemptions
are `bashoutput`, `bashoutputtool` and `kill_shell`, which carry no command line.

I kept the two calls separate instead of adding a field to `preparePermissionDetail`, so a display-text
builder does not start carrying a gate concern.

**On covering every path — and one I got wrong first.** My first version asserted that the two helpers
are called an equal number of times, on the theory that whoever adds one has to add the other. That
invariant is blind in exactly the place it needed to see: `tryRemoteOnlyApproval()` builds its entry
from its own field list and its three call sites spread **neither** helper, so the counts stayed equal
while a bubbles-disabled install had no reminder at all — and that path resolves a request through
`ctx.maybeAutoResolveSessionPermission()` without a card ever existing, so a session grant would have
released a recognized `git push --force` silently. Measured, setting on, same command, before the fix:

```
A ordinary route       stamp={"hold":true,"tag":"force-push"}   sweep gate = false
B remote-only (bubbles off)  stamp=undefined                    sweep gate = true
```

So the invariant is now totality at runtime instead of a source count: for each permission adapter, and
again with bubbles disabled, the pending entry must carry a derived view. The remote-only arm also
asserts it actually took that branch (`showPermissionBubble` never called), because an arm that
silently ran the ordinary path is how the gap survived the first test. Reverting the three spreads
turns that lane red.

There is also a safety net in the one funnel every accepted request passes through: an entry that
arrives with no view is stamped `{hold: true, tag: "not-inspected"}` rather than treated as unmatched,
so a path added later fails toward the human. The runtime lane asserts no real path reaches it. And the
consult is scoped to the tool-approval/unknown intents, so the net can never put a reason on a question
card.

**Session trust — a decision, not an inheritance.** `canOfferSessionTrust(entry)` is defined as
`canResolve(entry, { sessionOnly: true, mode: MODE_AUTO_TOOLS })`, which is the same predicate the
sweep uses, so a hold there does two things at once. The choice I made: **a held card does not offer
session trust.** The reminder exists to put a human in front of this one operation, and that card is
the single worst place to offer "stop asking me for this session" — one click there would disable the
reminder for everything after it. Your follow-on condition then holds structurally rather than by a
separate check: because matched requests keep deferring through the same chokepoint, a grant that is
already in place does not release them, and a sweep containing both kinds resolves only the ordinary
siblings. If that definition ever changes, this fails rather than silently inheriting something else:

```js
    assert.match(
      coordinator,
      /function canOfferSessionTrust\(entry\) \{\s*return canResolve\(entry, \{ sessionOnly: true, mode: MODE_AUTO_TOOLS \}\);/,
      "if this shape changes, the reminder's effect on the trust offer must be re-decided, not inherited"
    );
```

> 🟥 **Corrected 2026-09-16 — this paragraph was wrong about the one case it named.** It claimed
> the reason reaches the remote card "remote-only included". It did not: in remote-only operation the
> request reaches a human regardless, so the reminder is never *why* it is pending, and the gate that
> rendered the line returned null there — every time. Measured on a real machine, 2/2. Fixed in
> `509e3ec0`; see **Field run** below for the evidence and **Two tiers** for what the card says now.

**The reason, on every channel, in one wording.** `approvalDetailReminder` /
`approvalDetailReminderValue` in `src/i18n.js` (7 locales) render as a field on the remote card, so
Telegram and Lark get it through the existing `detail` + `fields` payload; the
local card gets `reminderTag` in the bubble payload and renders the same sentence. The held card's
reason wins over the local hint on purpose — the renderer's own scan reads the truncated copy, so on a
long command the two disagree and the main process is the one that is right. Timeout, disconnect and
no-decision paths are untouched; a held request is an ordinary pending entry from there on.

## The fixture choices, for review together

Three lists, not two, because "never matched" and "matched but deliberately excused" are different
answers and collapsing them would make a policy choice indistinguishable from the matcher quietly
breaking. `evaluatePermissionReminder()` returns `null`, `{hold:true, tag}`, or
`{hold:false, tag, exception}`.

Held (25 fixtures) — force-push, remote/local branch delete, hard reset, filter-repo, `rm -rf` of a
real path, `git clean -f`, publish, `gh repo delete`, `gh repo create --public`, `terraform destroy`,
`kubectl delete`, a SQL drop through a database client, and the same commands behind a wrapper
(`sudo`) or later in a chain (`cd x && npm publish`).

Unmatched (9 fixtures) — `grep -rn "git push --force" .`, `git commit -m "docs: how to npm publish"`,
`echo npm publish`, `echo docs\; npm publish`, `git push origin main`, `rm notes.txt`,
`psql -c 'SELECT 1'`. These are non-executing text and stay on the automatic path.

Excused (14 fixtures) — the three you named, and these are the choices I would most like your opinion on:

| Excused | Why | If you disagree |
|---|---|---|
| `--dry-run` anywhere in the matched segment (`npm publish --dry-run`, `kubectl delete --dry-run=client`, `git push --force --dry-run`) | A dry run performs nothing, whatever it is a dry run of. Applied to every pattern rather than scoped to publish. | Delete the `DRY_RUN` clause; three fixtures move from excused to held. |
| `git push --force-with-lease` | The lease aborts the push when the remote moved unobserved, which is the accidental loss this reminder is for. Holding the careful form is what teaches people to click through the reminder. A segment that also carries a bare `--force` is still held. | Delete the `FORCE_WITH_LEASE` clause; two fixtures move. This is the one I expect you may want the other way, and the DCG rule you linked is the argument for my side. |
| Deleting only disposable directories (`rm -rf node_modules`, `./dist`, `packages/web/node_modules`) | Reproducible by a build or an install. This is the `rm -rf ./node_modules` case you hit while testing #994. | Delete the `deletesOnlyDisposablePaths` clause; four fixtures move. |

The disposable-path carve-out is deliberately narrow and stays held when it cannot read the operands
plainly: `rm -rf dist src` (one target is not disposable), `rm -rf /var/tmp` (absolute),
`rm -rf ../node_modules` (traversal), and anything with a glob or a variable are all in the held list.
The basename list is `node_modules dist build out target coverage temp tmp venv .venv .cache .next
.nuxt .turbo .parcel-cache .pytest_cache __pycache__`.

Three over-blocks that a review pass turned up, now excused, each with a fixture:
`npm publish --help` and `terraform destroy --help` (asking a command to describe itself performs
nothing — **long form only**, because `-h` is psql's and mysql's HOST flag and excusing it would wave
through `psql -h db -c 'DROP TABLE users'`, which stays in the held list as a fixture), and
`rm -rf dist/*` (a trailing `/*` is the more common way to write `rm -rf dist`; only a trailing `/*`
is peeled, so `rm -rf */` and `rm -rf /*` stay held).

And one over-block I deliberately kept, because I do not think it is one: `rm -f package-lock.json`
holds. The pattern is `rm -[a-zA-Z]*[rf]`, so a single-file `rm -f` matches even without `-r`. A
single-file delete is still a delete, and nothing in the string distinguishes `rm -f build.log` from
`rm -f .env.local`. It is in the held fixtures so the choice is visible; narrowing the pattern to
require `-r` is a one-line change if you would rather have it that way.

Two more choices worth naming:

- **A scan that cannot complete ends at the human.** The display hint swallows a surprise and shows no
  badge, which is right for decoration; for a reminder the quiet direction is an automatic allow, so
  any throw becomes a hold with the reason `scan-error`. One pattern list, two error policies —
  `detectIrreversibleStrict()` is the shared matcher and `detectIrreversible()` is now a wrapper that
  catches, so the hint's behavior is unchanged and the two cannot drift into separate lists.
- **An unreadable setting.** If `isDestructiveReminderEnabled()` throws, an unmatched request keeps
  today's behavior and a matched one ends at the human. A settings read that throws is already a broken
  state, and the direction that costs an extra card seemed better than one that silently disarms a
  guard someone switched on. Happy to make it plain false if you prefer.
- **The reason is only shown when the reminder is actually why the card is there.** With automation
  off — or for an agent that is not automation-eligible — every request reaches a human anyway, so
  "Held for your review" would be a false claim about what Clawd did. In that case the card shows the
  ordinary destructive-action hint, exactly as before, and no reminder reason. What it does *not*
  change: the hint itself does not know about the policy exceptions, so a `--dry-run` publish still
  gets the generic badge it got before this change. That is pre-existing behavior and I left it alone
  rather than widen the diff into your merged feature.

Two things I left out deliberately, in case you would rather have them: an **excused** match shows
nothing on the card, so an operator cannot see "matched force-push, excused by force-with-lease" — the
distinction is kept in the return value and in the tests, not in the UI; and the two internal reasons
(`scan-error`, `not-inspected`) render through the same "matched: {reason}" sentence as a pattern name,
which reads oddly for them. Both are one-line changes if you want them surfaced differently.

It is not command-gated like `permissionAutomationMode`: those keys widen what runs without a human and
need a confirmed transition, this only narrows it, so it goes through the ordinary `settings:update`.
The setting description names the limits — commands inside scripts, unusual spellings, argument arrays,
text past the budget, and that `--dangerously-skip-permissions` sessions never reach Clawd at all —
and unmatched requests are described as unmatched, never as reviewed.

## Tests

`test/permission-destructive-reminder.test.js` (new, 83) and 5 lanes added to
`test/server-route-permission.test.js` (111 in that file).

| Acceptance item | Proven by |
|---|---|
| off: existing behavior unchanged | the 22 existing policy tests unmodified, plus two runtime lanes (absent getter, explicit false) |
| on: base automation would allow, the reminder holds, nothing is sent | runtime lane "on: the request is held, and nothing was sent" — entry still pending, `sendPermissionResponse` never called, `res.writableEnded` false |
| human Allow resolves only the intended request | runtime lane with two held siblings; the unanswered one is still pending and its response untouched |
| session-grant sweep | `canAutoResolvePendingPermission(entry, {sessionOnly:true, mode:"auto-tools"})` is false for the held request and true for an ordinary sibling |
| session-trust offer | the source-shape lane quoted above |
| past the display-preview cap, inside the budget | two lanes: the first asserts the truncated copy really has lost the command (so the second means something), the second asserts the hold |
| remote-only, and the reason on the remote card | a fake approval client through `maybeStartRemoteApproval`, asserting the reason is a field and not only inside the detail blob. 🟥 **This lane's name outran its fixture** — it left bubbles ON, so it only ever exercised the held case. That is how the remote-only gap shipped |
| **both tiers on the remote card** (added `509e3ec0`) | four arms pinning the *rendered string* verbatim: remote-only + automation off, automation on but not auto-allowable (`zcode`, `autoTools:false`), the held case, and the setting off. Two mutants fail it — reverting the fix, and swapping tier 2 to tier 1's wording |
| never a deny, never an upgrade, questions and plans untouched | four policy lanes |
| the fixture lists | 25 held / 9 unmatched / 14 excused, each excused fixture asserting its exception name |
| every shell tool automation can allow is scanned | derived from the policy's eligible-tool set, with three reviewed exemptions |
| every accepted request carries a view | two runtime lanes — 7 adapters on the ordinary path, and claude-code / codebuddy / zcode / dsh on the remote-only path with a branch-taken assertion |
| a question card is never given a reason | an elicitation entry carrying the safety-net stamp still renders `reminderTag: null` |
| no-decision is untouched | a held entry resolved as `no-decision` writes no body |
| session trust behavior (not just its definition) | `buildPermissionBubblePayload(held).canOfferSessionTrust === false`, ordinary sibling `true` |

Each guard was also checked by breaking it on purpose and confirming that exactly the intended lanes
go red: removing the policy hold, pointing the scan at the truncated copy, making a scan error quiet,
disabling the exceptions, dropping the remote-only spreads, removing the intent scope, narrowing the
shell tool names back, removing argv support, removing the `--help` exception, not peeling the
trailing glob, and letting the reason show when automation is off.

**Rendered, not only asserted.** Because of how #613 went — merged, then a follow-up because the badge
shipped hardcoded light-mode colors — I ran the real renderer in Electron rather than trusting the
tests on the display side. Six payloads x light/dark, measured in the actual window:

```
light  contrast 5.44:1   dark  contrast 8.44:1    no clipping, no card overflow, badge 342/408, 0 console errors
       "Held for your review — matched: force-push" / "matched: scan-error" / "matched: history-rewrite"
       ko "확인을 기다리며 멈췄습니다 — 일치: force-push"   ja "確認のため保留しました — 一致: infra-destroy"
```

Both pass AA, because the badge reuses the `--irreversible-*` custom properties your #625 introduced
rather than adding colors of its own. Two things that only showed up there:

- The settings row's description was **669 characters — the longest on the General tab by almost 3x**
  against a 240 maximum. The limits moved to the `descExtraKey` slot the agent rows already use, so it
  is now 207 + 216. A test pins the ceiling.
- The toggle renders `aria-checked="false"` in both themes, i.e. the default is off on screen and not
  only in the schema. That render predates the switch refactor that has since landed; the row is
  built with the shared `helpers.buildSwitchRow(...)` like the other twenty rows on the tab, so it
  inherits whatever that unification produces rather than carrying markup of its own.

Also worth knowing if you ever render that window in a test: a second `onPermissionShow` on a live
bubble page takes the `currentData` whitelist path at `bubble-renderer.js:1196` and does not re-render
the badge. My first measurement pass reported every case as the first one because of it.

**Suite status, stated precisely — and re-measured on today's `main`.** The branch is rebased onto
`782edefc` (the tip as of this writing), which was a clean rebase. On that tree:

```
this branch      10254 tests   13 fail
upstream/main    10165 tests   13 fail     <- control, same machine, same run session
```

The two failing sets are **identical, name for name** — I diffed them rather than comparing counts,
because equal counts would not have told me they were the same failures. They live in
`test/focus-windows.test.js` and `test/session-renderer-behavior.test.js` (Dashboard / HUD /
Direct-Send cases); I did not touch either file and have not investigated them. Happy to open a
separate issue if that is useful. `npm run audit:assets` and `npm run verify:winget-arch` are clean
(the asset audit's one warning is the pre-existing tracked-tree budget).

## The real-machine run

Per AGENTS.md this route needs a run against real Claude Code rather than a hand-rolled request, so
here is what was actually exercised, and what was not.

```
agent    : Claude Code 2.1.272
platform : macOS 26.6.2, node v24.15.0
channel  : local card
mode     : Permission request handling = Auto-approve, reminder ON
```

| | Result |
|---|---|
| A held `git push --force` under Auto-approve | held, **2/2** — once denied, once allowed, and in both cases the request stopped at the card first |
| The same session's other commands | **0 false positives**. `git push --force-with-lease` and `rm -rf node_modules` were excused by the named policy exceptions above, and `grep force` never matched at all |
| A tool with no command line | `AskUserQuestion` came through as `not-inspected`, i.e. the safety net, not an unmatched pass |

## Field run — the two arms, 2026-09-16

Both arms you asked for are now **run on a real machine**, and the second one found a defect in this
PR. Recording the result rather than only the fix, because the defect is the interesting part.

```
agent    : Claude Code (Opus 5)
platform : macOS 15 (Darwin 25.6.0), node v24.15.0
channel  : telegram + remote-only (bubbles disabled)
build    : this branch, run from source (npm start) - the released 1.0.0 has no such toggle
```

**Arm 1 — Telegram, bubbles on: PASS.** `git push --force origin HEAD` under permission automation.

```
destructive reminder: holding for a human mode=unattended reason=force-push tool=Bash agent=claude-code
telegram approval debug: native approval card sent
response: {"decision":{"behavior":"allow"}}          # 26s later, tapped on the phone
```

The card carried the reason as its own labelled field, on both surfaces:

```
local (bubble)          ⚠ 확인을 기다리며 멈췄습니다 — 일치: force-push   (amber strip, above the command box)
Telegram                알림: 확인을 기다리며 멈췄습니다 — 일치: force-push
```

Not over-blocking: the three Bash requests immediately before it were auto-allowed with no hold in
the same session and the same mode. Only the force-push stopped.

**Arm 2 — remote-only, bubbles off: the `알림` field was absent entirely. 2/2 reproduced.**

```
permission bubbles disabled, routed to Telegram-only approval: tool=Bash
telegram approval debug: native approval card sent
                                                    # no "destructive reminder: holding" line
```

The card rendered Agent / Tool / Folder / Summary and stopped. `reminderIsWhyThisIsPending()` is
false whenever the request was reaching a human anyway — which, with bubbles off, is always — so the
single-tier payload had nothing to render. The local card degrades to the weaker irreversible badge
in that same state; a remote-only operator has no local badge, so the one configuration the code
comment named by hand was the one told nothing.

Fixed in `509e3ec0` by porting `bubble-renderer.js`'s two-tier badge into the remote payload, with a
separate i18n value for tier 2 (7 locales) so it does not claim Clawd stopped a request it did not:

```
held by the reminder        Held for your review — matched: force-push
matched, pending anyway     Destructive action — may not be recoverable (matched: force-push)
```

**Two instrument failures before the arms could run at all**, named because each would have produced
a confident wrong answer: the session's own auto-approval mode absorbed the permission request before
any hook fired (`permission-debug.log` had zero entries — *not* "the reminder missed"), and the first
force-push was sent as `timeout 150 git push --force …`, a prefix the matcher does not scan. `cd … &&`,
`sudo` and `env FOO=1` all still match; `timeout` and `--force-with-lease` do not (the latter is a
declared exception, and the setting's own note says an unusual spelling can be missed).

**Cross-family adversarial review** (Gemini 3.x + GPT-5.6, both non-Anthropic) raised ten findings on
the fix. Four were adopted. Four were **refuted at the source** and are listed rather than dropped —
a null dereference, an empty-tag asymmetry, a local/remote pattern-set divergence, and a structural
`reminderTag` payload field. `reminderHolds()` already rejects a null stamp, a malformed one and an
empty tag, and both scanners share one pattern list. Both families converged on the first three
independently, which is worth stating plainly: agreement between decorrelated reviewers is not
evidence, and every one of those three rested on not opening `reminderHolds()`.

Full suite run locally to completion: 10,255 tests, 13 failures — the same 13 this branch already
surfaces in `session-renderer-behavior` and Windows terminal focus, unchanged with and without this
commit (checked both ways).

**One thing the run made obvious that I had not appreciated when I opened #994**, and it may matter
for the "consistent coverage across integrations" concern you raised: when the *agent* is configured
to approve on its own, the request never reaches Clawd at all — no hook fires, no card, nothing for
this setting to hold. A trusted folder in Claude Code, an `allow` entry in its settings, or
`--dangerously-skip-permissions` all have that shape. So this reminder covers "automation inside
Clawd", not "automation anywhere", and the setting description says so rather than leaving a user to
discover it. I mention it because a reader could reasonably expect the stronger thing from the
feature name, and because it is the same boundary that decides which integrations this can ever
cover.

## Files

`src/permission-reminder.js` (new) · `src/permission-automation-policy.js` ·
`src/permission.js` · `src/server-route-permission.js` · `src/bubble-format.js` ·
`src/bubble-renderer.js` · `src/prefs.js` · `src/settings-actions.js` ·
`src/settings-tab-general.js` · `src/main.js` · `src/i18n.js` · `src/settings-i18n.js` ·
2 test files.

The setting name, the pattern list and every row in the tables above are yours to shape.


rullerzhou-afk (changes requested):
Thank you for the exceptionally thorough write-up, the explicit policy choices, and the real-machine evidence. The product direction is sound: a default-off reminder that can only downgrade an automatic allow to human review is a useful safety layer.

I am requesting changes because the current matcher/exception composition fails open for ordinary compound commands. This is inside the exact boundary the setting promises to protect, and it can silently auto-approve a destructive operation.

## 1. Blocking safety defect: an exception on the first match skips every later destructive match

`detectIrreversibleStrict()` returns the first matching pattern/segment. `evaluatePermissionReminder()` then evaluates `documentedException()` only for that match and immediately returns `{ hold: false, ... }` when the exception applies.

That makes the exception apply to the entire request rather than only to the matched operation. Any unexcused destructive command after the first excused match is never examined.

Concrete examples with the setting enabled and `auto-tools`/a session grant:

```text
npm publish --dry-run && rm -rf /
rm -rf node_modules; git push --force origin main
git push --force-with-lease && git reset --hard HEAD^
terraform destroy --help ; npm publish
npm publish --help && gh repo delete me/repo
rm -rf dist/* && rm -rf /etc
kubectl delete --dry-run=client po x && kubectl delete ns prod
```

For example, the first command currently produces the equivalent of:

```json
{"hold":false,"tag":"publish","exception":"dry-run"}
```

and the real permission path still reports that it can auto-resolve the request. `rm -rf /` by itself is held, so this is not an acknowledged matcher miss: the dangerous command is recognizable, but it is skipped because an earlier operation was excused.

There is a second form of the same bug within one segment. For example:

```text
git push --force-with-lease --delete origin main
```

The force-push match is excused by `--force-with-lease`; the separate remote-delete pattern is never evaluated.

The required invariant should be:

- inspect every shell segment and every applicable destructive pattern;
- if any match is not explicitly excused, hold the whole request;
- return `hold: false` only when at least one destructive match exists and every destructive match is excused;
- return `null` only when no destructive match exists.

Please add regression tests that exercise both cross-segment and same-segment composition. At minimum:

```text
npm publish --dry-run && rm -rf /
rm -rf node_modules; git push --force origin main
git push --force-with-lease && git reset --hard HEAD^
git push --force-with-lease --delete origin main
```

The existing fixtures are comprehensive for individual decisions, but every exception fixture contains only one effective destructive decision, so they cannot detect this composition failure.

## 2. Blocking safety defect: `--dry-run` accepts values that mean “execute normally”

The current expression accepts any non-whitespace value after `--dry-run=`:

```js
/(^|\s)--dry-?run(=\S*)?(\s|$)/
```

As a result, this is excused and can still be auto-approved:

```text
kubectl delete namespace prod --dry-run=none
```

For `kubectl`, `none` explicitly disables dry-run. The same matcher also excuses `--dry-run=false`, `--dry-run=0`, and arbitrary values without proving that they are non-executing.

Because this exception is shared across commands, it needs command-aware or strictly allowlisted semantics. Unknown values must fail closed. Please cover at least:

- forms that are actually non-executing for the relevant command (`--dry-run`, and Kubernetes `client`/`server` where applicable);
- `none`, `false`, `0`, and unknown values, which must not be excused;
- option-boundary behavior such as `npm publish -- --dry-run`, where a token after `--` must not be assumed to make the publish itself a dry run.

## 3. Incorrect Tier 1 explanation when another gate already requires human review

`reminderIsWhyThisIsPending()` partially re-evaluates the automation policy, but it does not use all of the entry-level gates in `canAutoResolvePendingPermission()`.

A reachable example is a remote-only entry with a session grant where either:

- the Codex permission intercept is disabled; or
- the subagent automation gate is disabled.

Those requests already have to reach a human, yet the remote card renders the Tier 1 “Held for your review” wording. It should render Tier 2 because the reminder was not the reason the request remained pending.

Please derive this explanation from the same canonical predicate used by the sweep/session-grant decision, with the reminder hold temporarily neutralized, rather than maintaining a partial duplicate. A regression test should pin at least the Codex-intercept-disabled + session-grant case. The current tests named “remote-only” should also create a real `remoteOnly: true` entry/session record; the existing lanes do not fully model that entry shape.

## 4. The advertised scan budget does not bound the argv conversion

For argv-shaped input, `parts.join(" ")` happens before `boundCommandString(..., SCAN_MAX)`. `ARGV_MAX` limits the number of elements, not their total byte/character count, so the reminder allocates a full joined copy before retaining 4096 characters.

The current HTTP body limit bounds the practical impact to roughly another request-sized allocation, so I do not consider this a security blocker today. However, the module comment and PR claim that time and memory are fixed regardless of accepted request size are not correct, and the safety currently depends on an unrelated ingress limit.

Please either build the string incrementally and stop at the scan budget, or correct the documentation and make the upstream dependency explicit. A large-element argv test should observe the pre-scan allocation bound, not only the final returned string length.

## 5. Setting-off behavior and pending-request policy need to be stated accurately

The PR says that when the setting is off, “nothing reads it.” In the current routing flow, all accepted permission paths still run `preparePermissionReminder()` and store the stamp; the setting is consulted later when the policy/payload consumes it. The observable decision/UI behavior remains unchanged, but the implementation and description disagree. I am fine with retaining eager scanning—the present cost is small—but the comments and PR text should describe it accurately.

There is also a time-of-check/time-of-use policy choice: the match stamp is computed when the request arrives, while the setting is read when resolution/sweeping occurs. Turning the setting off can therefore make an already-pending held request eligible for a later session-grant sweep, while its existing card may still display “Held for your review.” Please decide and test the intended behavior for pending entries. Either snapshot the enabled/disabled decision with the entry, or explicitly update/re-render and document the live-setting semantics.

## 6. Coverage and platform claims

Adding `powershell` to `SHELL_TOOLS` makes PowerShell-shaped requests eligible for scanning, but there are no PowerShell/cmd destructive-command patterns. For example, `Remove-Item -Recurse -Force ...` and `cmd /c rmdir /s /q ...` remain unmatched. This can remain a documented best-effort limitation, but adding the tool name should not be presented as meaningful Windows destructive-command coverage. Please add known-miss fixtures so that boundary is explicit.

Finally, the repository checks currently do not execute the two test files that carry this feature (`permission-destructive-reminder` and the relevant server-route coverage); the workflow unit/audit jobs use fixed file lists. These regressions are too safety-sensitive to depend only on local execution. Please include the relevant tests in CI.

## Verdict

The architecture and authority boundary are otherwise well chosen: the feature is default-off, only converts `AUTO_ALLOW` to `DEFER`, does not synthesize denials, and keeps question/plan-review behavior separate. The held-card/session-trust choice is also coherent.

The first two findings are merge blockers because the feature can say “this destructive operation will wait for you” while silently auto-approving it. Once all matches are evaluated, dry-run exceptions are made semantically safe, and composition tests are added, I will be happy to review the revision.

chrono-meta:
Thanks — the composition point was right, and following it found more than the two I first fixed.

Your wording was that every exception fixture contains only one effective destructive decision, so
they cannot detect this composition failure. That was the whole defect: the matcher stopped at the
first pattern hit per segment, and `optionTokens` read the granting options from the wrong place, so
an exception earned by one operation covered a second one that was never examined.

Two changes for that:

- `detectIrreversibleMatches` now evaluates every segment × pattern rather than returning on the
  first hit. `detectIrreversibleStrict` delegates to it with a limit of 1, so the existing callers
  keep their shape.
- `optionTokens` returns `{ granting, all }`. Documented exceptions (`--dry-run`, help-only,
  `--force-with-lease`) consult `granting`, which is head-filtered; `hasPlainForce` consults `all`.
  The asymmetry is deliberate: a quoted token must not be able to *grant* an exception, but it must
  still be able to *cancel* one.

While re-reading the segmenter against your finding I found four more spellings of the same class.
All four allowed a delete that the plain form holds:

| command | before | after |
|---|---|---|
| `rm -rf ./d` | hold | hold |
| `( rm -rf ./d )` | hold | hold |
| `{ rm -rf ./d; }` | **allow** | hold |
| `:; { rm -rf ./d; }` | **allow** | hold |
| `echo "$(rm -rf ./d)"` | **allow** | hold |
| ``echo `rm -rf ./d` `` | **allow** | hold |

`(` was normalised and `{` was not, so the same delete held or allowed purely by which grouping
keyword was typed. The shell requires a blank after `{` for the group keyword, which is the
discriminator that leaves `{a,b}` and `${VAR}` alone.

Substitution bodies were unreachable for the quote-aware split by construction — the double quote
swallows the whole string, so the only segment is `echo …`. That one is a second, additive pass: it
can only append segments, so if it is wrong it produces a false hold, never a missed delete. Single
quotes suppress it, `$((` is arithmetic, depth is capped at 3, and the existing 50-segment cap
still applies.

Each fix has a case with controls, and reverting the source reddens exactly those cases and nothing
else (152 pass / 2 fail), with a byte-identical restore. Suite is 154 pass / 0 fail
(`node test/run-tests.js` also has 10 pre-existing failures in `session-renderer-behavior.test.js`
that are unrelated — they reproduce identically on the base commit, so I left them alone).

Still open, and I would rather name them than let them read as covered:

- **Heredoc bodies are scanned as commands.** That is a false hold, not a missed delete. I did write
  the skip and reverted it: it swallowed a `;` on the opener's own line, read `<<E'OF'` as
  delimiter `E`, and took `$((1<<2))` for a heredoc opener. Each of those hid a command the shell
  runs, so the cheap failure was better than the three expensive ones. Doing it properly needs real
  parsing rather than a scan.
- **Variable indirection** (`RM=rm; $RM -rf x`) and shell aliases are out of reach for a scanner.
- `$()`/backtick spans inside a `#` comment are treated as command positions — again a false hold.
- The 4KB and 50-segment inspection caps are unchanged; input past them is not examined.

chrono-meta:
Pushed `b7158ec0` — one gap I found while re-reading the reminder path after your hardening commits.

`sh -c '<command>'` reached `evaluatePermissionReminder` as a single opaque segment, so a request the plain spelling holds was allowed once it was wrapped. Measured on this branch before the change, with the plain spellings as controls so the probe is not silently dead:

```
git push origin main --force              -> {hold: true, tag: "force-push"}
rm -rf /tmp/x                             -> {hold: true, tag: "file-delete"}
git status                                -> null
sh -c 'git push origin main --force'      -> null
bash -c "rm -rf /tmp/x"                   -> null
/bin/sh -c 'git push origin main --force' -> null
sh -c "sh -c 'rm -rf /tmp/x'"             -> null
```

The reason it slipped past `WRAPPER` is that an interpreter is a different shape. `sudo`/`env`/`nohup` prefix a command without changing what runs, so stripping the leading word leaves the command behind. `sh -c` carries the command inside a quoted argument instead — strip the prefix and you are left with a quoted word, which no anchored pattern matches.

The body goes back through `segmentCommands`, so it reuses the depth cap and `SEGMENT_MAX` budget the substitution path already established rather than introducing a second recursion with its own bound. The interpreter line stays in the segment list, since it is a real command a later pattern may want.

`python3 -c` and friends are deliberately outside the alternation: only a shell runs its `-c` argument as shell syntax, and widening it would start matching program source as commands. That boundary is pinned by lanes rather than left to a comment.

Lanes: six in `HOLD` for the spellings that show up in practice (quoted both ways, absolute path, bundled `-ec`, a flag before `-c`, one level of nesting) and four in `UNMATCHED` for the boundary — `python3 -c`, a script file, a harmless carried command, `ssh`. Removing the change turns exactly those six red and leaves the four green; I checked that rather than assuming it. Full file: 194 pass, 0 fail.

Two notes that are not this change:

- `test (macos-latest)` is red on `c001599b` while ubuntu and windows pass. I could not read why — the served job log is truncated, and the successful ubuntu job's log truncates at the same point, so the absence of an assertion failure in it is not evidence of anything. Re-running that job is the only thing that would settle it.
- Things that still carry a command as data are out of reach the same way: `eval`, `xargs sh -c`, `ssh host 'cmd'`, herestrings, `source`. I did not chase them by extending the alternation, since that list has no natural end.

rullerzhou-afk (approved):

rullerzhou-afk:
Thank you for all the work on this, @chrono-meta — I really appreciate the care you put into it. Sorry I missed saying this when I merged it this morning!

PR #1025, opened 2026-09-17
fix(renderer): clear data-clawd-drag-direction when the drag ends

chrono-meta:
`endDragReaction()` clears the JavaScript state but leaves the DOM attribute behind:

```js
function endDragReaction() {
  const wasDragReacting = isDragReacting;
  isDragReacting = false;
  currentDragSvg = null;
  currentDragDirection = null;   // JS only
  if (!wasDragReacting) return;
  window.electronAPI.resumeFromReaction();
}
```

`data-clawd-drag-direction` is written to the SVG document root by `applyDirectionalDragToObject()`, and that function has no clearing path — `normalizeDragDirection(null)` returns `null` and the function returns early at line 1565, so nothing removes the attribute. `grep -n 'removeAttribute("data-clawd-drag-direction")' src/renderer.js` returns nothing.

That would be harmless if the document were replaced on release, but it is deliberately not. The universal Codex Pet wrapper keeps **one** document across drag, release and state changes — `test/renderer-low-power.test.js` says exactly that in its own test name:

> `keeps drag, release, and mid-drag state changes in one Codex Pet document`

So after a drag ends, the stale direction stays on the live document and keeps feeding the two consumers that read it:

- the direction-dependent rules in `src/codex-pet-adapter.js`
- `shouldFlipAssetDirection()` in `src/pet-accessory-mirror.js`, which mirrors again on top of it

The two can compose, which is how this surfaces as a sprite facing the wrong way after a drag rather than as an obvious error.

### The existing test walks right past it

That same test asserts the visual swaps correctly after release, but never checks the direction attribute:

```js
harness.electronHandlers.onEndDragReaction();
harness.electronHandlers.onStateChange("idle", "codex-pet-idle-loop.svg");
...
assert.strictEqual(attached.root.getAttribute("data-clawd-codex-pet-visual"), "idle-loop");
// data-clawd-drag-direction is still "right" here, and nothing asserts on it
```

### Change

One call in `endDragReaction()`, plus a helper that is the mirror of the existing setter:

```js
function clearDirectionalDragFromObject(objectEl) {
  if (!objectEl || objectEl.tagName !== "OBJECT") return false;
  try {
    const root = objectEl.contentDocument && objectEl.contentDocument.documentElement;
    if (!root || !root.hasAttribute("data-clawd-drag-direction")) return false;
    root.removeAttribute("data-clawd-drag-direction");
    return true;
  } catch {
    return false;
  }
}
```

`data-clawd-drag-directional` is **not** touched. That one is the document's capability marker rather than drag state, so it has to survive for the next drag — there is a test for that below.

`cancelReaction()` is deliberately left alone. `does not apply a stale drag direction when a pending object loads after cancel` pins the behaviour that the cancel path must not write to a pending document, and clearing there breaks it. Release and cancel are different paths here.

### Tests

| Test | What it pins |
|---|---|
| `clears data-clawd-drag-direction when the drag ends` | the attribute is gone after release, and stays gone across the following state change |
| `leaves the directional marker in place so the next drag can reuse the bridge` | `data-clawd-drag-directional` survives, and a second drag still sets a direction |

`FakeElement` in the test harness gained `hasAttribute` / `removeAttribute`. Both are new methods, so no existing assertion changes behaviour.

Verified:

- `node --test test/renderer-low-power.test.js` fails on the new test before the fix, passes after
- removing only the `clearDirectionalDragFromObject(clawdEl)` line turns exactly that one test red again, and nothing else
- `npm test` — full suite result below

### What I could not verify

`AGENTS.md` puts real drag behaviour in manual-verification territory, and this machine is macOS while the project is Windows-first, so I did not exercise a physical drag. The change is pinned at the unit level against the harness the repo already uses for this area; the visual end of it is unverified by me.

I also left `state-visual-resolver.js` alone. `bindings.roam` is still `{ files: [], fallbackTo: "idle" }`, so a theme without a roam asset walks in its idle pose — that looked like a separate question, and reusing the drag wrapper for it is exactly what this attribute bug would have corrupted.

### Full suite, against a baseline

| | `upstream/main` | this branch |
|---|---|---|
| pass | 10106 | 10108 |
| fail | 13 | 13 |
| skipped | 46 | 46 |

The 13 failures are pre-existing and identical on both runs — 10 in `test/session-renderer-behavior.test.js`, 3 in `test/focus-windows.test.js`, none in the file this PR touches. The +2 are the two tests added here.


rullerzhou-afk:
Thanks for the thorough write-up and tests. We re-reviewed this against current `main` and the original #620 lifecycle contract, including checking the generated wrapper's computed styles in Chromium.

We're closing this PR because the proposed root cause does not hold in the current implementation, while the change creates a reachable wrong-direction state during release:

- `data-clawd-drag-direction` has one production consumer: the generated CSS selector scoped by `[data-clawd-codex-pet-visual=drag-directional]`. `pet-accessory-mirror.js` / `shouldFlipAssetDirection()` does not read this attribute, so the two consumers described in the PR do not compose.
- Once the visual changes to idle/working, the leftover direction attribute is dormant. The next horizontal drag explicitly writes the new direction before using the directional visual, so the old value does not leak into the next drag.
- On a leftward release, this patch removes `direction=left` while the document is still `visual=drag-directional`. Missing/invalid direction intentionally falls back to the right-running row. The real state visual arrives only after the renderer -> main -> renderer `resume-from-reaction` / `state-change` IPC round trip, so the patch opens a left -> right -> idle/working transition window. It can last longer when the mini transition or another reaction defers the state resend.

The added test validates attribute removal, but it immediately invokes `onStateChange()` afterward and therefore does not cover the rendered state between drag end and the state-change IPC. The cleanup helper also lacks the existing `data-clawd-drag-directional="v1"` ownership gate, runs before the `wasDragReacting` guard, omits `pendingNext`, and conflicts with current `main`'s `|| isReacting` cursor-polling ownership fix.

Because the retained attribute has no demonstrated user-visible effect and removing it at this point can reintroduce the direction flicker that #620 avoided, we do not want to pursue this cleanup as proposed. If there is a reproducible post-release wrong-facing case beyond this dormant attribute, please open an issue with the exact theme/state sequence or a recording and we can investigate that path directly.

Thanks again for taking the time to investigate this area.

PR #1031, opened 2026-09-19
fix(claude): probe transcript by raw session id, not canonical key (#908)

### LINKED ISSUE #908: Track Claude transcript sessionId canonical/raw mismatch
## Problem

Claude transcript completion fallback currently compares transcript entry `sessionId` values against Clawd's canonical session key.

Production state ingestion canonicalizes Claude's raw `session_id` into a local session key, while real Claude transcript entries commonly carry the raw UUID-shaped `sessionId`. The transcript extractor only accepts entries with no `sessionId` or entries whose `entry.sessionId === sessionId`, so a real raw `sessionId` transcript can miss the probe when the probe is called with the canonical key.

## Why this matters

PR #904 intentionally did not fix this mismatch. Its scope is assistant-output lifecycle cleanup and deletion cleanup for delayed completion callbacks. Fixing this mismatch would reactivate a currently quiet production path and should be reviewed separately with real Claude Code evidence.

The #904 test suite now includes a current-state guard for this boundary: a transcript fixture with raw `sessionId` plus a canonical map key must not be treated as a production end-to-end success. When this follow-up is fixed, that current-state test should turn red and be updated with the new expected behavior.

## Evidence

- `hooks/clawd-hook.js` transcript filtering accepts entries with no `sessionId` or `entry.sessionId === sessionId`.
- `src/state.js` stores local Claude sessions under canonical keys produced from the raw session id.
- `test/state.test.js` includes a raw `sessionId` mismatch guard added by #904.

## Suggested acceptance criteria

- Decide the intended contract for transcript probe lookup: canonical key, raw session id, or both.
- Update the transcript probe/extractor wiring so real Claude transcript entries with raw `sessionId` are handled intentionally.
- Keep the no-`sessionId` fixture path covered, since it exercises the mechanism-level fallback used in #904's deletion-cleanup regression tests.
- Add or update tests for both real raw-`sessionId` transcript shape and no-`sessionId` fixture shape.
- Validate with real Claude Code before describing the path as production-verified.

hanzhe-one:
## Problem

The delayed Claude transcript completion probe (`scheduleClaudeTranscriptCompletionProbe` in `src/state.js`) re-reads the transcript tail and filters entries by session id, but it passed Clawd's **canonical** session key. Real Claude transcript entries carry the **raw** `session_id` (UUID). `entry.sessionId === canonicalKey` never matched, so a genuine raw-session transcript could not promote the withheld completion through this fallback path. This is the follow-up #904 deliberately left out of scope.

## Contract decision

The transcript probe now looks up by **raw session id**. The canonical key is a Clawd-internal identifier and never appears in Claude transcripts, so the raw id is the only value that can match real entries.

## Change

`src/state.js` — pass `session.rawSessionId` to the extractor, falling back to the canonical key only when no raw id was captured:

```js
const assistantOutput = extractLastClaudeAssistantTextFromEntries(
  readClaudeTranscriptTailEntries(safePath),
  session.rawSessionId || sessionId
);
```

Entries that carry no `sessionId` are still accepted by the extractor (`!entry.sessionId`), so the mechanism-level fallback used by #904's deletion-cleanup regression tests is unchanged.

## Tests

The #904 current-state guard in `test/state.test.js` (`"Claude transcript fallback documents raw transcript sessionId mismatch"`) previously documented the broken behavior — session stuck `working`, no assistant output, no completion. As #908 predicted, it turns red under the fix, so it is updated to the new expected behavior: the probe promotes to `done` with the raw transcript's final assistant text. Added an assertion pinning that the canonical key differs from the raw id, so the guard cannot silently pass if the two ever coincide.

Verification:

- `node --test test/state.test.js test/clawd-hook.test.js` → 483 pass, 0 fail.
- Confirmed the fix is load-bearing: reverting only the `src/state.js` change turns the #908 test red again.
- The no-`sessionId` fixture path (`"...falls back to transcript completion when Stop is missed"`) still passes, so the #904 mechanism-level fallback is preserved.

> Note: I could not run the full `npm test` locally because native deps (`koffi`) are not built in my environment; those are the only failures and are unrelated to this change. CI will run the full suite with deps installed.

## Acceptance criteria (from #908)

- [x] Decide the transcript-probe lookup contract → raw session id.
- [x] Wire the probe so real raw-`sessionId` transcripts are handled.
- [x] Keep the no-`sessionId` fixture path covered.
- [x] Tests for both raw-`sessionId` and no-`sessionId` shapes.
- [ ] Validate with real Claude Code — I don't have a repro capture; flagging for a maintainer with real transcript evidence before calling the path production-verified.

Closes #908


rullerzhou-afk (changes requested):
Thanks — the raw-session lookup itself is correct and independently load-bearing. I confirmed that real transcript entries use the raw UUID, and the targeted and full test suites pass on this head.

I am requesting changes for one lifecycle regression that becomes active once this previously dormant probe starts matching production transcripts.

After an AskUserQuestion `PostToolUse`, the delayed probe reuses `extractLastAssistantTextFromEntries()` as a completion predicate. That extractor accepts text from an assistant entry even when the same entry also contains a `tool_use` or `server_tool_use` block. If Claude writes a text-plus-tool preamble and the 2-second probe fires before the corresponding `PreToolUse` hook reaches state.js, the probe synthesizes a completion while the turn is still running.

I reproduced this on the exact PR behavior with a raw-session transcript entry containing text plus an Edit tool use. The probe changed the session to idle/done, stored the preamble as the final output, and entered attention. A later PreToolUse can restore working state, but the completion sound, notification/snapshot, and recap turn-complete event have already happened and cannot be retracted.

Please add a probe-specific completion guard — for example, reject candidate assistant entries that still contain `tool_use` or `server_tool_use` blocks — and add a state-level regression test for this sequence. An equivalent affirmative completion/quietness fence is also fine, but the current extractor alone is not a safe mid-turn completion signal.

Two smaller follow-ups:

- The new comment says the code falls back to the canonical key when no raw id was captured, but stored sessions always have a truthy rawSessionId; missing wire identity is represented as `default`. Please align the comment with the actual contract rather than documenting an unreachable fallback.
- Both the PR and commit say `Closes #908`, while the real Claude Code end-to-end acceptance item in #908 is still unchecked. Please keep #908 open or remove the auto-close keyword unless that real-environment evidence is added.

Independent validation on head 7b188927: targeted tests 483/483 pass; full suite 10,730 pass, 0 fail, 46 skip; clean integration with current origin/main and the targeted tests still pass. The exact real Claude timing frequency remains NOT TESTED, so the reproduction proves reachability, not incidence.

hanzhe-one:
Thanks for the thorough review — all three points addressed in `8f96bb05`.

**Mid-turn completion regression (blocker).** Added an opt-in `rejectToolUse` mode to `extractLastAssistantTextFromEntries`. When on, the newest matching assistant entry becomes a completion predicate: if it still carries a `tool_use`/`server_tool_use` block the extractor fails closed (returns null) rather than falling back to an older, superseded text entry, so the probe reschedules instead of synthesizing a completion. Only the mid-turn probe passes the flag; the Stop-time extractor (which runs after a turn has genuinely ended) is untouched. New state-level regression `"...does not complete on a text-plus-tool-use preamble (#908 review)"` reproduces your exact sequence (narration + `Edit` tool_use) and asserts the session stays `working` with no completion sound. I confirmed it's load-bearing — reverting the guard turns it red.

**Unreachable fallback comment.** Corrected. The comment no longer claims a canonical-key fallback; it now states that stored sessions always carry a truthy `rawSessionId` (it falls back to the session key at construction, and a missing wire identity is `default`), so the value passed is the real underlying id.

**Auto-close keyword.** Changed `Closes #908` → `Refs #908` in both the PR and commit. The real-Claude-Code E2E acceptance item stays open — the new test proves the guarded path is reachable, not its production incidence, and I don't have a live capture to close that item.

Local verification: `node --test test/state.test.js test/clawd-hook.test.js` → 484 pass, 0 fail. I can't run the full suite here (`koffi` native dep isn't built in my env), so CI / your run remains the source of truth for the full 10.7k.

rullerzhou-afk (approved):
Re-reviewed current head `8f96bb05`. The probe-specific tool-use guard fixes the premature-completion case, the raw-session-id lookup is correct, and the comment/issue reference now match the actual scope.

Independent validation:
- focused tests: 484 passed
- full suite: 10,973 passed, 0 failed, 46 skipped
- real macOS smoke with Electron 41.10.4 and Claude Code 2.1.269: completion recovered with no Stop hook, while a live AskUserQuestion → Bash tool sequence did not complete early

Approved.

